PR #17, opened 2025-11-11
Pull for Issue #16

### LINKED ISSUE #16: Purchases from Distributors (Devin)
- Update initialization in Society to map each product to a vector of distributors.
- Update Person to have a vector of distributors in ranked order.  The size of this vector is the same as the size of the society's vector of distributors.
- Implement the two purchase functions by finding a distributor that sells the desired product and then calls that distributor's _sell_ function.

lennyyyyyyyy:
Resolve #16
Besides this: 
 - Define set_society in Person and update Society constructor to call this method on all its people
 - Define has_product in Distributor

### DIFF
--- a/supply-demand-signals-2/header/Constants.h
+++ b/supply-demand-signals-2/header/Constants.h
@@ -8,6 +8,7 @@ const int MONTH = DAY * 30;
 const int YEAR = DAY * 365;
 
 const double PRODUCTIVITY_THRESHOLD = 60.0; // threshold for productivity before lowering working hours
+const int INITIAL_WORK_HOURS_DAILY = 8;
 
 const double RANDOM_RETIREMENT_CHANCE = 8.3e-7;
 const int GUARANTEED_RETIREMENT_AGE = 50 * YEAR; 
--- a/supply-demand-signals-2/header/Constants.h
+++ b/supply-demand-signals-2/header/Constants.h
@@ -7,7 +7,8 @@ const int WEEK = DAY * 7;
 const int MONTH = DAY * 30;
 const int YEAR = DAY * 365;
 
-const double PRODUCTIVITY_THRESHOLD = 60.0; // threshold for productivity before lowering working hours
+// Threshold for productivity before lowering working hours
+const double PRODUCTIVITY_THRESHOLD = 60.0;
 const int INITIAL_WORK_HOURS_DAILY = 8;
 
 const double RANDOM_RETIREMENT_CHANCE = 8.3e-7;
--- a/supply-demand-signals-2/header/Distributor.h
+++ b/supply-demand-signals-2/header/Distributor.h
@@ -1,5 +1,4 @@
 #pragma once
-
 #include <string>
 #include <tuple>
 #include <unordered_map>
@@ -10,14 +9,16 @@
 #include "Producer.h"
 
 class Distributor : public Firm {
-  private:
-    std::vector<Producer*> suppliers;
-    std::unordered_map<Product*, int> inventory;
-    
   public:
     Distributor();
     Distributor(std::vector<Machine*> machines, std::vector<Person*> workforce, std::vector<Plan*> plans);
     double get_output_ratio(Product& product);
     double planned_satisfaction_per_person(Product& product, Person& person);
+	bool has_product(Product* product);
     void sell_goods(Product& product, int quantity, Person * person);
+
+  private:
+    std::vector<Producer*> suppliers;
+    std::unordered_map<Product*, int> inventory;
+    
 };
--- a/supply-demand-signals-2/header/Person.h
+++ b/supply-demand-signals-2/header/Person.h
@@ -16,7 +16,7 @@ class Person {
     enum HealthStatus { HEALTHY, UNHEALTHY, RECOVERING };
     enum Expertise { EXPERTISE_1, EXPERTISE_2, EXPERTISE_3 };
 
-    Person(const std::unordered_map<std::string, int>& expertise, int age, HealthStatus health_status, const std::unordered_map<Product*, double>& purchase_frequencies);
+    Person(const std::unordered_map<std::string, int>& expertise, int age, HealthStatus health_status);
   
     std::unordered_map<Product*, double>& get_purchase_frequencies();
     HealthStatus get_health_status();
@@ -24,18 +24,18 @@ class Person {
     float avg_productivity_over_time_step(std::string product_name);
     void get_paid(double income);
     void charge(double cost);
-    void purchase_good(Product* p, int quantity);
+    void purchase_good(Product * p, int quantity);
     void purchase_goods();
     bool will_retire();
-	void set_society(Society* society);
+	void set_society(Society * society);
   
   private:
     std::unordered_map<std::string, int> expertise;
     int age;
     HealthStatus health_status;
     std::unordered_map<Product*, double> purchase_frequencies;
-    Firm* employer = nullptr;
-	Society* society = nullptr;
+    Firm * firm = nullptr;
+	Society * society = nullptr;
     double account;
 	std::vector<Distributor*> ranked_distributors;
 };
--- a/supply-demand-signals-2/header/Person.h
+++ b/supply-demand-signals-2/header/Person.h
@@ -5,10 +5,11 @@
 #include <unordered_map>
 #include <vector>
 
-#include "Firm.h"
 #include "Product.h"
 
+class Distributor;
 class Firm;
+class Society;
 
 class Person {
   public:
@@ -23,15 +24,18 @@ class Person {
     float avg_productivity_over_time_step(std::string product_name);
     void get_paid(double income);
     void charge(double cost);
-    void purchase_good(Product& p, int quantity);
+    void purchase_good(Product* p, int quantity);
     void purchase_goods();
     bool will_retire();
+	void set_society(Society* society);
   
   private:
     std::unordered_map<std::string, int> expertise;
     int age;
     HealthStatus health_status;
     std::unordered_map<Product*, double> purchase_frequencies;
-    Firm* employer;
+    Firm* employer = nullptr;
+	Society* society = nullptr;
     double account;
+	std::vector<Distributor*> ranked_distributors;
 };
--- a/supply-demand-signals-2/header/Society.h
+++ b/supply-demand-signals-2/header/Society.h
@@ -2,6 +2,7 @@
 
 #include <unordered_map>
 
+#include "Constants.h"
 #include "Distributor.h"
 #include "Firm.h"
 #include "Machine.h"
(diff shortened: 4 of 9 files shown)


supply-demand-signals-2/header/Constants.h
```diff
@@ -8,6 +8,7 @@ const int MONTH = DAY * 30;
 const int YEAR = DAY * 365;
 
 const double PRODUCTIVITY_THRESHOLD = 60.0; // threshold for productivity before lowering working hours
```
Amittai-Aviram: Could you please move the comment to right above the statement (and please capitalize the first character)?  I'm not sure whether I understand how poductivity is going to work or why it starts at 60.0.  Maybe you can explain it on Sunday?
Also — not part of this PR — will we ever use MONTH?  It's wrong about half the time, so maybe we should avoid it.

supply-demand-signals-2/header/Constants.h
```diff
@@ -8,6 +8,7 @@ const int MONTH = DAY * 30;
...
+const int INITIAL_WORK_HOURS_DAILY = 8;
 
 const double RANDOM_RETIREMENT_CHANCE = 8.3e-7;
 const int GUARANTEED_RETIREMENT_AGE = 50 * YEAR;
```
Amittai-Aviram, line 15: Is this from birth or from entry into the workforce?

supply-demand-signals-2/header/Constants.h
```diff
@@ -8,6 +8,7 @@ const int MONTH = DAY * 30;
 const int YEAR = DAY * 365;
 
 const double PRODUCTIVITY_THRESHOLD = 60.0; // threshold for productivity before lowering working hours
+const int INITIAL_WORK_HOURS_DAILY = 8;
 
 const double RANDOM_RETIREMENT_CHANCE = 8.3e-7;
```
Amittai-Aviram, line 14: Is this just an arbitrary number?  Do you have a rationale for it?

supply-demand-signals-2/header/Distributor.h
```diff
@@ -10,14 +9,16 @@
...
-  private:
-    std::vector<Producer*> suppliers;
-    std::unordered_map<Product*, int> inventory;
-    
   public:
     Distributor();
     Distributor(std::vector<Machine*> machines, std::vector<Person*> workforce, std::vector<Plan*> plans);
     double get_output_ratio(Product& product);
     double planned_satisfaction_per_person(Product& product, Person& person);
+	bool has_product(Product* product);
```
Amittai-Aviram, line 17: Nice!

supply-demand-signals-2/header/Distributor.h
```diff
@@ -10,14 +9,16 @@
...
     void sell_goods(Product& product, int quantity, Person * person);
+
+  private:
+    std::vector<Producer*> suppliers;
+    std::unordered_map<Product*, int> inventory;
```
Amittai-Aviram, line 22: Nice! 👍

supply-demand-signals-2/header/Person.h
```diff
@@ -23,15 +24,18 @@ class Person {
     float avg_productivity_over_time_step(std::string product_name);
     void get_paid(double income);
     void charge(double cost);
-    void purchase_good(Product& p, int quantity);
+    void purchase_good(Product* p, int quantity);
```
Amittai-Aviram: Yeah, that will probably work better, at least for now.

supply-demand-signals-2/header/Person.h
```diff
@@ -23,15 +24,18 @@ class Person {
...
     bool will_retire();
+	void set_society(Society* society);
   
   private:
     std::unordered_map<std::string, int> expertise;
     int age;
     HealthStatus health_status;
     std::unordered_map<Product*, double> purchase_frequencies;
-    Firm* employer;
+    Firm* employer = nullptr;
```
Amittai-Aviram: I guess we can assume, for now, that a person only works for one firm at a time, right?  Does the person need access to his/her firm?  The firm needs to credit the worker's account somehow, but the firm has a vector of workers, right?  So I'm not sure how we'll use the pointer to the firm.  Also, our LTE-friendly critics might bristle at the term "employer."  Maybe just "firm" for the variable name?  It's a worker co-op.

supply-demand-signals-2/header/Person.h
```diff
@@ -23,15 +24,18 @@ class Person {
...
+	Society* society = nullptr;
```
Amittai-Aviram: OK, this is picky, but, for our global style, could we please put a space before and after the asterisk when it is part of the type designation in a declaration? — Here,
Society * society = nullptr;
Thank you!

supply-demand-signals-2/header/Society.h
```diff
@@ -19,13 +20,12 @@ class Society{
 	std::vector<Product*> products;
 	std::vector<Producer*> producers;
 	std::vector<Distributor*> distributors;
-	std::unordered_map<Product*, Distributor*> product_to_distributor;
+	std::unordered_map<Product*, std::vector<Distributor*>> product_to_distributors;
     std::unordered_map<Firm*, double> prices;
     std::unordered_map<std::string, int> avg_needs();
-    const int initial_work_hours_daily = 8;
-    int current_work_hours_daily = initial_work_hours_daily;
+    int current_work_hours_daily = INITIAL_WORK_HOURS_DAILY;
```
Amittai-Aviram, line 26: 👍

supply-demand-signals-2/header/Society.h
```diff
@@ -19,13 +20,12 @@ class Society{
...
 
-    Society(std::vector<Person*> people, std::vector<Product*> products, std::vector<Producer*> producers, std::vector<Distributor*> distributors, std::unordered_map<Product*, Distributor*> product_to_distributor, std::unordered_map<Firm*, double> prices);
+    Society(std::vector<Person*> people, std::vector<Product*> products, std::vector<Producer*> producers, std::vector<Distributor*> distributors, std::unordered_map<Product*, std::vector<Distributor*>> product_to_distributors, std::unordered_map<Firm*, double> prices);
```
Amittai-Aviram, line 28: Good.  (Distributor * — etc.). I'm going to ask Sal to revise the Society so that it constructs all this stuff itself instead of taking it in from **main**.

supply-demand-signals-2/src/Distributor.cpp
```diff
@@ -15,6 +15,10 @@ double Distributor::planned_satisfaction_per_person(Product& product, Person& pe
     return get_output_ratio(product) * person.get_purchase_frequencies()[&product];
 }
 
+bool Distributor::has_product(Product* product) {
+	return inventory[product];
```
Amittai-Aviram, line 19: 😊

supply-demand-signals-2/src/Person.cpp
```diff
@@ -56,23 +56,26 @@ float Person::get_current_productivity() {
...
+			distributor->sell_goods(*p, quantity, this);
+			return;
+		}
+	}
 }
 
 void Person::purchase_goods() {
 	static std::uniform_real_distribution<> dist(0, 1);
 
 	for (std::pair<Product*, double> p : purchase_frequencies) {
```
Amittai-Aviram, line 70: Can this be a reference so as to avoid copying the pair?

supply-demand-signals-2/src/Person.cpp
```diff
@@ -56,23 +56,26 @@ float Person::get_current_productivity() {
...
 		if (dist(Sim::gen) < p.second) {
-			purchase_good(*(p.first), 1);
+			purchase_good(p.first, 1);
```
Amittai-Aviram, line 72: OK, I know we discussed this, and I guess it does work, right?  On average, over time, the person's frequency of purchase of this product is the set frequency, because we are getting a uniformly distributed random number and only purchase if it's less than the frequency.  Can you maybe put together a 2-to-5-minute explanation of this logic to present to the group on Sunday?

supply-demand-signals-2/src/Person.cpp
```diff
@@ -86,4 +89,10 @@ bool Person::will_retire() {
 
 	return dist(Sim::gen) < RANDOM_RETIREMENT_CHANCE;
 }
+
+void Person::set_society(Society* society) {
+	this->society = society;
+	ranked_distributors = society->distributors;
+	std::shuffle(ranked_distributors.begin(), ranked_distributors.end(), Sim::gen);
```
Amittai-Aviram, line 90: Cool!

supply-demand-signals-2/src/Society.cpp
```diff
@@ -4,14 +4,17 @@
...
+Society::Society(std::vector<Person*> people, std::vector<Product*> products, std::vector<Producer*> producers, std::vector<Distributor*> distributors, std::unordered_map<Product*, std::vector<Distributor*>> product_to_distributors, std::unordered_map<Firm*, double> prices)
+	: people(people), products(products), unemployed_people(people), producers(producers), distributors(distributors), product_to_distributors(product_to_distributors), prices(prices) {
 	for (Producer* producer : producers) {
 		firms.push_back(producer);
 	}
 	for (Distributor* distributor : distributors) {
 		firms.push_back(distributor);
 	}
+	for (Person* person : people) {
+		person->set_society(this);
```
Amittai-Aviram, line 16: For now, very good!  After Sal redoes the Society's constructor (if that's what happens), then we can go back and pass the Society * to the Person constructor instead, and everything will look better.

Amittai-Aviram (commented):
Excellent work!  😊. I left you a few questions and suggestions.  So please go through those, and feel free to respond to anything directly with a reply.  I should be able to approve before Sunday.

supply-demand-signals-2/header/Constants.h
lennyyyyyyyy: I remember putting this here because Sal had defined this constant before I made constants.h. Could ask Sal on Sunday.

supply-demand-signals-2/header/Constants.h
lennyyyyyyyy: Right now those two are equivalent for us, I believe. When we make that distinction it should be from entry into the workforce.

supply-demand-signals-2/header/Constants.h
lennyyyyyyyy: I also think this is another one of Sal's constants that I grouped with the rest.

supply-demand-signals-2/header/Person.h
lennyyyyyyyy: We're currently not using it right now, but I think it might be nice to have later. I'll rename it to firm, though.

supply-demand-signals-2/src/Person.cpp
lennyyyyyyyy: Got it.

Amittai-Aviram (changes requested):
I think we'll be good to merge once you've changed that one iteration variable in the for-loop in Person.cpp, if that's possible.   Thanks!

supply-demand-signals-2/src/Person.cpp
lennyyyyyyyy: Actually, I forgot that now that a person has access to a global list of products, purchase_frequencies doesn't have to be passed in the constructor. I'll get rid of this and initialize it in set_society instead without any copying.

Amittai-Aviram (approved):
LGTM!

Issue #27, opened 2025-11-17
Model Worker Allocation (Devin)


Amittai-Aviram:
# Product
- Define an Ability enumeration type.  Remove the Expertise enumeration type from Person.
- Add a field `required_abilities` that is a vector of Ability enumerators, listed in ranked order from high to low priority.
- Have the Product constructor construct this ranking by making a random-sized random subset of abilities.  (These can be uniformly distributed.)

# Person
- Change the Expertise map to a map Ability -> double, `abilities`.
- Have the constructor choose a random subset of Ability enumerators and set their values at a random rational number with a normal distribution and mean of 1.  The variance can be fairly high.

# Society
- Add a queue of people seeking work.
- Define accessors so that a person can add him/herself to the list of those seeking work and so that a firm can look through the list.  A person is added to the end of the queue; a firm starts looking from the front of the queue.
- At the beginning of time, all workers are on this list.

# Plan (in Firm)
- Add `std::vector<Person *> workers`.
- Add `int allotted time`.  This functions as deadline relative to the start time.

# Firm
- Define a function `int assign_workers(std::vector<Ability>, int num_workers, Plan& plan)` to look through the firm's current workers for a good match with the requirements.
  - The function takes a list of required abilities and a number of workers as inputs, so that it will work for both producers and distributors. It returns an integer representing how many hours (time steps) the firm estimates that it needs to fulfill the order that had given rise to the plan.
  - The function first looks in the firm's current workers for a match.  Assume that the input vector of abilities is in ranked order from high to low priority.
  - For each suitable worker, it copies the address (pointer) from the firm's general pool of workers to the plan.
  - If it cannot find enough workers in the firm, it looks in the society for workers who are not already in firms, following the same procedure of copying addresses.
  - If it can find enough workers all together, it removes those workers from their current places and returns the expected completion time, which should be the same as the requested completion time.
  - Otherwise, it needs to decide between two options: dividing the work up among the available workers, training workers without the required abilities.
    - Assume (for now) that it takes a constant week (defined as such in `constants.h`) to train a worker to acquire a new ability.  (We need some support infrastructure to list workers to be trained for a given ability.)
    - If it takes less time to train than to divide up the work, add the hours needed for training to the total estimate, record which workers need which training, and return the estimate.
    - Otherwise, return the estimate resulting from simply working longer with fewer workers.

# Producer
[...]

PR #31, opened 2025-11-23
Person shopping events and abilities

### LINKED ISSUE #25: Person shopping events (Devin)
# Give all entities access to the current timestep in Sim:
 - Change current_time_step to be static
 - Create static int get_current_time_step

# In constants:
 - Create const int SHOPPING_PERIOD

# In Person:
 - Create int shopping_offset 
    - analogous to what time/day of the week does the person go shopping
 - Change void purchase_goods() to void shop()
    - figure out what the average should be by multiplying product frequency with SHOPPING_PERIOD
    - then normally vary that by some standard deviation
 - Create void on_time_step()
    - Eventually there should be one of these in every entity (Society, Firm, Producer, Distributor)
    - For now, just age the person, shop if person will shop, retire if person will retire

lennyyyyyyyy:
Resolve #25 
Complete the Product, Person, and Society parts of #27


supply-demand-signals-2/header/Constants.h
```diff
@@ -18,10 +18,15 @@ const double AVERAGE_RETIREMENT_AGE = (1 - pow(1 - RANDOM_RETIREMENT_CHANCE, GUA
...
+const int PERSON_SHOPPING_PERIOD = WEEK;
+const double PERSON_SHOPPING_OFFSET_STDDEV = DAY / 2.0;
+const double PERSON_ABILITY_STDDEV = 0.2;
 
 const int STARTING_PEOPLE = 1000;
 const int STARTING_PRODUCTS = 100;
 const int STARTING_PRODUCERS = STARTING_PRODUCTS;
 const int STARTING_DISTRIBUTORS = STARTING_PRODUCTS;
+
+enum Ability { ABILITY_1, ABILITY_2, ABILITY_3 };
```
Amittai-Aviram: Could we please move this to the Person class definition?  (See the Style Guide: "Enumeration types, wherever possible, should be associated with specific classes and should be defined within their respective class definitions. Typically, an enumeration type belongs in the public section.")

supply-demand-signals-2/header/Person.h
```diff
@@ -14,9 +14,8 @@ class Society;
 class Person {
   public:
     enum HealthStatus { HEALTHY, UNHEALTHY, RECOVERING };
-    enum Expertise { EXPERTISE_1, EXPERTISE_2, EXPERTISE_3 };
```
Amittai-Aviram, line 17: Hmmm.  Why did you move this (or why did we move this) from here into `constants.h`?  Oh!  We also changed the name, right?  Anyway, please keep the new name but move it back here.

supply-demand-signals-2/header/Person.h
```diff
@@ -14,9 +14,8 @@ class Society;
...
 
-    Person(const std::unordered_map<std::string, int>& expertise, int age, HealthStatus health_status);
+    Person(int age, HealthStatus health_status, std::unordered_map<Ability, double> starting_abilities = {});
```
Amittai-Aviram: Why doesn't the constructor just always assign all of the abilities as (normally distributed) random values (with mean = 1)?  What is the scenario in which we would need to pass starting abilities into the constructor?
Also, maybe the age of a worker should by assumption be some constant value (17 or 18 years?)?  And the health status should also be set by the constructor.  In short, we probably should only use a default constructor.

supply-demand-signals-2/header/Person.h
```diff
@@ -25,17 +24,21 @@ class Person {
     void get_paid(double income);
     void charge(double cost);
     void purchase_good(Product * p, int quantity);
-    void purchase_goods();
+	bool will_shop();
+	void shop();
     bool will_retire();
+	void retire();
 	void set_society(Society * society);
+	void on_time_step();
```
Amittai-Aviram: Please put this near the top, right under the constructor.

supply-demand-signals-2/header/Person.h
```diff
@@ -25,17 +24,21 @@ class Person {
     void get_paid(double income);
```
Amittai-Aviram: Please rename this to `register_hours_worked`, which is the name that IDA uses, as it happens.

Amittai-Aviram (changes requested):
Great work!  I have not quite finished reviewing, but here are a few comments that include a couple of requests for changes.  Thank you!

supply-demand-signals-2/header/Constants.h
lennyyyyyyyy: I had a good amount of trouble with circular dependencies when putting it in Person (because then Product would have to include the Person header to get the Ability definition, but Person already includes the Product header). I can get around this as long as I forward declare the enum in Product with:

 `enum Ability: unsigned int;`

What do you think?

Amittai-Aviram: D'oh!  Sorry!  I should have thought of that!  Let's put it into Society!  Then both Person and Product can include Society.h .  (Would this work?)  Conceptually, it sort of makes sense that the society defines the abilities that it considers relevant to production.

supply-demand-signals-2/header/Person.h
Amittai-Aviram: Yeah, cool.  Thanks.

Amittai-Aviram (approved):
LGTM!  Please merge. 🥇

PR #36, opened 2025-11-27
Define abstract ancestor base class Agent

### LINKED ISSUE #35: Define abstract Agent class as universal base class
Define an abstract class `Agent`, with pure virtual function `void on_time_step()`, to serve as the base class (or "interface") for all agent classes in the simulation model.

Amittai-Aviram:
- Defined an abstract `Agent` class with the one pure virtual function `on_time_step`.
- Had all agent classes that would need to be updated (agents proper) inherit from `Agent` and override `on_time_step`.
- Created a new top-level makefile that allows building of one module at a time.
- Fixed a number of build issues and warnings, as well as a few style issues.
- Removed the temporary `main` function from `Sim` and moved the logic to `main.cpp`.

Resolves #35

### DIFF
--- a/supply-demand-signals-2/header/Agent.h
+++ b/supply-demand-signals-2/header/Agent.h
@@ -1,0 +1,5 @@
+#pragma once
+
+class Agent {
+    virtual void on_time_step() = 0;
+};
--- a/supply-demand-signals-2/header/Distributor.h
+++ b/supply-demand-signals-2/header/Distributor.h
@@ -1,4 +1,5 @@
 #pragma once
+
 #include <string>
 #include <tuple>
 #include <unordered_map>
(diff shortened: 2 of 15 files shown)


lennyyyyyyyy (changes requested):
Everything looks good, but I can't find the Agent class definition. I'm assuming it lives in its own header file, so perhaps you used `git commit -am`? This command doesn't stage new files.

Amittai-Aviram:
> Everything looks good, but I can't find the Agent class definition. I'm assuming it lives in its own header file, so perhaps you used `git commit -am`? This command doesn't stage new files.

Ugh!  Thanks for noticing that.  I had forgotten to `git add` the header file.  Yes, it's `header/Agent.h`.  Thanks!

lennyyyyyyyy (approved):
Looks good!

Issue #30, opened 2025-11-21
Revise Producer (Sal)


Amittai-Aviram:
- Producer starts with no orders and no workers.
- The Firm constructor takes a pointer to the Society only.
- The producer has an API to ask it to produce a new product not in its catalogue.
- If the distributor finds no producer for the desired product, it chooses a new producer at random (uniform distribution) and calls the "new product" API on that producer.
- The `int new_product(Order *)` API is something of a wrapper for `receive_order`, as described in Issue #27.  Please coordinate with Devin on this.
   - Both `receive_order` and (therefore) `new_product` should return a time estimate for completion of the order that may have a special value (`-1`, defined as a symbolic constant in `constants.h`) meaning "reject order."  (I think Devin will set this up.)
- When the distributor accepts the bid by calling `receive_order` again with the same order (see Issue #27), the producer executes the plan:
      - Purchases inputs from distributors.  (It finds distributors the same way as a consumer does.)  Labor time for purchases of inputs comes from the "raw materials" or "machinery" accounts of the plan.  (Focus on raw materials.)
      - Employs the workers already allocated to the plan when it was a draft plan.  (Again, see Issue #27).
      - On each time step, deducts `num_workers` from its living labor account and transfers one hour to each worker's personal account.
      - At the end of the plan, we assume (for now) that the quantity of ordered products has been produced.
      - (This part will be subject to different work paces and spontaneous innovations later on.)
      - Transfers the products to the distributor.  The distributor increases the inventory and the producer's PRD goes to 0.
      - Reduces the remaining lifespan of each machine listed in the plan by (hours-of-living-labor / num-workers).  (In sum, this should be the same as the machinery entry in the plan.  We will get to that.)

PR #38, opened 2025-12-03
Order handling, Plan execution, worker allocation

lennyyyyyyyy:
Implemented the majority of #27 and #30 including: 
 - In Firm, drafting a plan when recieving an order, allocating workers considering options with and without training
 - In Producer, draft_order(), drop_order(), pursue_order(), start_plan(), execute_plan(), end_plan()
 - Cleaned up a lot of header directives that were causing issues

Things unfinished / next steps:
 - Machines are currently not considered (but I feel like I need to talk to someone more before I implement them). I think this pull request is good though, so the API is set and branches don't diverge too much
 - Distributor needs to be updated to use the new Producer API
 - PriceController needs to return doubles instead of void 
 - Pretty bad efficiency removing workers from their current pools when pursuing an order
 - Producers need a way to buy raw materials and machines from Distributors in the same way People do


supply-demand-signals-2/header/Constants.h
```diff
@@ -21,6 +21,10 @@ const double AVERAGE_RETIREMENT_AGE = (1 - pow(1 - RANDOM_RETIREMENT_CHANCE, GUA
 const double PRODUCT_FREQUENCY_MIN = 1.0 / (2 * WEEK);
 const double PRODUCT_FREQUENCY_MAX = 1;
 
+const int FIRM_TRAINING_TIME = INITIAL_WORK_HOURS_DAILY * WEEK / DAY;
```
Amittai-Aviram, line 24: This is correct, but the formulation looks confusing because you have to divide by DAY.  Let's replace the relevant definitions above as follows:
```
const int DAYS_PER_WEEK = 7;
const int WEEK = DAY * DAYS_PER_WEEK;
// ...
const int INITIAL_WORK_HOURS_DAILY = 8;
const int INITIAL_WORK_DAYS_PER_WEEK = 5;
// ...
```
This way, you can have 
```
const int FIRM_TRAINING_TIME = INITIAL_WORK_HOURS_DAILY * INITIAL_WORK_DAYS_PER_WEEK;
```

supply-demand-signals-2/header/Firm.h
```diff
@@ -6,36 +6,53 @@
...
 class Person;
 
 struct Plan {
-    Product * product;
-    std::vector<Machine> machines;
+	// independent/input fields
+	Order * order;
     Firm * firm;
+	std::vector<Person*> workers;
+	int training_time;
```
Amittai-Aviram, line 22: Are this and the one below really input fields? The firm should set them to 0 by default and then set them to a non-zero value only if it cannot find enough qualified workers.

supply-demand-signals-2/header/Firm.h
```diff
@@ -6,36 +6,53 @@
...
+struct Order;
+struct Product;
 class Firm;
+class Machine;
 class Person;
 
 struct Plan {
-    Product * product;
-    std::vector<Machine> machines;
+	// independent/input fields
```
Amittai-Aviram, line 18: Maybe we should set these through the constructor — or at least _a_ constructor — and get rid of the comment?

supply-demand-signals-2/header/Firm.h
```diff
@@ -6,36 +6,53 @@
...
+	Order * order;
     Firm * firm;
+	std::vector<Person*> workers;
+	int training_time;
+	int training_time_remaining;
+
+	// dependent/output fields
```
Amittai-Aviram, line 25: See above. I am hoping that we can make this clear enough so that the comment is not necessary. The field names are very good.

supply-demand-signals-2/header/Firm.h
```diff
@@ -6,36 +6,53 @@
...
 	
-    Firm();
-    Firm(std::vector<Machine*> machines, std::vector<Person*> workforce, std::vector<Plan*> plans);
+	Firm();
     
     double get_avg_productivity();
+
+  protected:
+    std::unordered_map<Product *, int> inventory;
+    std::unordered_map<Product*, std::vector<Plan*>> plan_history; // unused and prob need to change later
```
Amittai-Aviram, line 47: Yeah. Let's remember to write an issue about this so that it does not disappear. We do  need a history of actually adopted plans, because it's part of the public "general ledger." We'll also probably use it when we set up a plan approval process (equivalent to IDA's Accountant role), to prevent overproduction.

supply-demand-signals-2/header/Product.h
```diff
@@ -8,13 +8,14 @@
...
     Product(const std::string name, double price, int order);
-    int get_required_labor();
     void add_input(Product * input, int quantity);
     std::string product_name;
     double price_per_unit;
     int order_size;
 	double base_frequency;
-    std::unordered_map<Product *, int> inputs;
+	
+    std::unordered_map<Product *, int> inputs_per_order;
```
Amittai-Aviram, line 17: I would expected inputs per unit. I think that might wind up being more useful for various other computations.

supply-demand-signals-2/src/Distributor.cpp
```diff
@@ -62,7 +64,9 @@ void Distributor::check_and_reorder() {
                 int order_quantity = product->order_size;
                 std::cout << "Reordering " << order_quantity << " units of " 
                           << product->product_name << std::endl;
-                Order * order = producer->accept_order(product, order_quantity, this);
+				//NOTE: SOLELY FOR BUILDING
```
Amittai-Aviram, line 67: I do  not quite understand how this is supposed to work. Let's make sure to write up an issue to fix it.

supply-demand-signals-2/src/Distributor.cpp
```diff
@@ -98,9 +102,9 @@ int Distributor::get_inventory(Product * product) {
 }
 
 bool Distributor::is_overproduced(Product* product) {
```
Amittai-Aviram, line 104: We need to ask Sal what this function is about.  It looks like a relic from some logic that I do not think that we are supporting.

supply-demand-signals-2/src/Firm.cpp
```diff
@@ -1,10 +1,123 @@
...
+	double suitability = 0.0;
+	for (Ability ability : required_abilities) {
+		suitability += person->get_abilities()[ability];
+	}
+	suitability /= required_abilities.size();
+	return suitability;
+}
+
+int Firm::predict_workers_needed(Order * order) {
+	return ceil(order->quantity * order->product->living_labor_per_order * DAY / Society::instance->current_work_hours_daily / order->requested_turnaround_time);
```
Amittai-Aviram, line 25: Good!  But let's see whether a few tweaks can simplify this arithmetic. See my comments on the constants. Maybe we can make corresponding changes to Society.

supply-demand-signals-2/src/Firm.cpp
```diff
@@ -1,10 +1,123 @@
...
+
+int Firm::predict_turnaround_time(Order * order, double total_suitability) {
+	return std::ceil(order->quantity * order->product->living_labor_per_order * DAY / total_suitability / Society::instance->current_work_hours_daily);
+}
+
+int Firm::predict_labor_hours(Order * order, double total_suitability) {
+	return std::ceil(order->quantity * order->product->living_labor_per_order / total_suitability);
+}
+		
+void Firm::assign_plan_dependent_fields(Plan * draft_plan, std::vector<Ability>& required_abilities) {
```
Amittai-Aviram, line 62: Style: please break up overlong lines of code.

supply-demand-signals-2/src/Firm.cpp
```diff
@@ -1,10 +1,123 @@
...
+	double total_suitability = 0.0;
+	double max_suitability = 0.0;
+	for (Person * worker : draft_plan->workers) {
+		double worker_suitability = suitability(worker, required_abilities);
+		total_suitability += suitability(worker, required_abilities);
+		max_suitability = std::max(max_suitability, worker_suitability);
+	}
+	if (draft_plan->training_time) {
+		draft_plan->predicted_turnaround_time = draft_plan->training_time + predict_turnaround_time(draft_plan->order, max_suitability * draft_plan->workers.size());
```
Amittai-Aviram, line 71: Why do we multiply `max_suitability` by the number of workers? (Sorry if we already talked about this — I do  not remember covering it.)

supply-demand-signals-2/src/Firm.cpp
```diff
@@ -1,10 +1,123 @@
...
+	draft_plan->prd = -(draft_plan->total_hours);
+}
+
+void Firm::draft_optimal_plan(Plan * draft_plan, std::vector<Ability>& required_abilities) {
+	// try without training first
+	Plan * draft_plan_without_training = new Plan(*draft_plan);
+	assign_workers_by_suitability_threshold(draft_plan_without_training, required_abilities, 0.0);
+	assign_plan_dependent_fields(draft_plan_without_training, required_abilities);
+	if (draft_plan_without_training->predicted_turnaround_time <= draft_plan->order->requested_turnaround_time) {
+		*draft_plan = *draft_plan_without_training; return;
```
Amittai-Aviram, line 93: Style: put every statement (here, `return;`) on its own line.

supply-demand-signals-2/src/Firm.cpp
```diff
@@ -1,10 +1,123 @@
...
+	
+	// did not meet the deadline without training
+	// pick the better option between training or no training
+	Plan * draft_plan_with_training = new Plan(*draft_plan_without_training);
+	draft_plan_with_training->training_time = draft_plan_with_training->training_time_remaining = FIRM_TRAINING_TIME;
+	assign_plan_dependent_fields(draft_plan_with_training, required_abilities);
+	if (draft_plan_without_training->predicted_turnaround_time < draft_plan_with_training->predicted_turnaround_time) {
+		*draft_plan = *draft_plan_without_training;
+	} else {
+		*draft_plan = *draft_plan_with_training;
```
Amittai-Aviram, line 104: Should you `delete` one of the draft plans? It looks as if you have constructed two on the heap, but are keeping only one. Also — OK, this might deserve a separate issue — if we are allocating the plan on the heap, we should probably just keep it on the heap and store its address in the firm.

supply-demand-signals-2/src/PriceController.cpp
```diff
@@ -10,10 +12,10 @@ void PriceController::update_prices() {
 
 }
 
-void PriceController::get_price(Product * product) {
+double PriceController::get_price(Product * product) {
     if (product == nullptr) {
         std::cerr << "Error: Cannot get price for null product" << std::endl;
-        return;
+        return 0.0;
```
Amittai-Aviram, line 18: Less ambiguous if you return -1.0.

supply-demand-signals-2/src/PriceController.cpp
```diff
@@ -87,10 +90,10 @@ void PriceController::get_plan_history() {
     
     for (size_t i = 0; i < plan_history.size(); ++i) {
         Plan * plan = plan_history[i];
-        if (plan != nullptr && plan->product != nullptr) {
+        if (plan != nullptr && plan->order->product != nullptr) {
```
Amittai-Aviram, line 93: Style: `if (plan && plan->order->product)`

Amittai-Aviram (changes requested):
Bravo! Looks great so far!  I am asking for a few small changes. Also, we need to make sure to write up new issues for things remaining to be done. In that regard, I do not see the logic involved in the distributor comparing "bids" and then coming back and choosing the best one. Is that work to be done later?

Issue #34, opened 2025-11-26
Compute estimated machinery costs for a draft plan (Sal)


Amittai-Aviram:
When a producer drafts a plan (see Issue #27), we shall assume the following (for  now):
- All machines are used continually all the time (i.e., throughout the work day, all the time that workers are working).
- Machines are used _equally_.

The machinery component of costs should therefore be computed as follows:
$m = (l/w) * |m|$
where $m$ is the cost in labor time for machinery for the plan, $l$ is the living labor time in the plan, $w$ is the number of workers available to execute the plan, and $|m|$ is the number of machines.

It is then safe, at the end of a plan's execution, to subtract $m / |m|$ from each relevant machine's remaining lifespan.

PR #39, opened 2025-12-07
Added choose_order and Updated on_time_step (Issues #30 & #34)

Sal47506:
Features Implemented:

- choose_order: Distributor calls draft_plan_order for all of its suppliers, then picks an order based on fastest time completion and that leads to the other draft_plans being deleted for the other suppliers. The draft plan it chooses ends up being finalized through pursue_order

- on_time_step: prd gets updated each time step by adding m and the difference between labor_hours and labor_hours_remaining, machine hours get updated as mentioned in #34 and labor_hours_remaining gets updated for each time step

Issues addressed: #30 & #34


supply-demand-signals-2/src/Distributor.cpp
```diff
@@ -98,9 +154,9 @@ int Distributor::get_inventory(Product * product) {
 }
 
 bool Distributor::is_overproduced(Product* product) {
```
Amittai-Aviram, line 156: What exactly is this for?

Amittai-Aviram (approved):
I'm concerned—or just confused—about overlap with Devin's current PR, but I am going to approve this anyway so as not to hold it up. Then we'll have to deal with merge conflicts if they arise.
Please be sure to use the key word `resolves` in your PR headnote so as to cause the issues to close automatically.  Please also make a list of remaining issues, so that we can write them up as new issues.
I am confused about `is_overproduced`, but perhaps you can explain that to me, so that I need not hold you up in merging this code into `main`.

PR #53, opened 2025-12-10
Remove price from Product constructor

### LINKED ISSUE #51: Remove price from Product constructor (Amittai)
When a product is constructed, it only has a name — and, either at construction or later, it has
- Estimated living labor per unit
- Estimated quantities (in natural units) of raw materials per unit
- A vector of needed machines (not quantified)

The Price Controller computes the overall price. On all time steps after the product is constructed, the price controller updates living labor with average actual cost in living labor and recomputes and updates the total per-unit price.

Amittai-Aviram:
This change cleans up some of the logic related to products. The agent that constructs a product (the society at start-up, and then, presumably, producers responding to demands for new products — to be modeled later) does not already have the total price and does not need it. The constructor does assign an initial quantity of living labor presumed necessary to make _one unit_ of the product, along with per-unit entries of inputs. The Price Controller will then compute the overall price by multiplying and adding as necessary.

The constructor cannot assign input unit quantities because it first needs to set up the initial collection of products. It must then go through the products on a second round have each one determine random selections of inputs and how many units of each input are needed to make the product.

A somewhat similar sequence is necessary for machinery, but that presupposes that the society make a separate collection of machines, which is a separate issue.

This issue blocks further progress on the Price Controller.

Resolves #51 .


Amittai-Aviram:
(Bump.)

PR #59, opened 2025-12-12
Set Product inputs without circular dependencies; clean up Society class

### LINKED ISSUE #55: Improve Society API (Amittai)
RIght now, all members of `Society` are public, making for a poor separation of concerns and suboptimal OO design.  The entities right  now that require access to `Society`'s members are `Person`, `Firm`, and `Producer`.
- Set members appropriately as `private`
- Provide accessors and access techniques as necessary.
- Declare classes as `friend`s only if necessary.

### LINKED ISSUE #58: Set Product inputs without circular dependencies (Amittai)
The current procedure for setting the inputs (products and quantities) for each product, in the `Society` constructor, risks circular dependencies, including, a product requiring itself as an input, or product A requiring product B, where product B requires product A.

Instead, set an initial group of "base products" that have no inputs. Then set inputs incrementally through the vector of products. Each product can only list as inputs the products whose inputs have already been set (including the base products, which have no inputs).

Amittai-Aviram:
- Divided `Society` class definition into public API and private members and removed obsolete code.
- Moved the handling of `Product` initialization in the `Society` constructor to a helper function that first creates the products, then skips inputs for the "base products," and, finally, sets the inputs for the other products incrementally, so that each product can only depend on products whose inputs are already set (or deliberately empty).

Resolves #55 
Resolves #58


Sal47506 (approved):
Approved

PR #60, opened 2025-12-13
Capitalist Economy Driven By Events

Kkenize:
Capitalist Economy now has its market demand fully driven by events. Graphs are drawn to demonstrate this.


Event-Based Simulation/CapitalistEconomy.py
```diff
@@ -45,8 +46,60 @@ def __init__(self, params):
 
         # calculate initial f(t,y) 
         self.dydt = self._get_dydt(self.params)
+        
+        # RANDOM EVENT SYSTEM 
+        self.events_catalog = []          # to be loaded later by simulation
+        self.active_events = []           # list of event objects currently active
+        self.allow_random_events = False  # becomes True after equilibrium
+        self.event_steps_remaining = 0    # counts down 40 steps
```
Amittai-Aviram: The comment is confusing: you set steps_remaining to 0. Will you count _up_ to 40? You would count _down_ is you started at 40 and counted _down_ to 0.

Event-Based Simulation/CapitalistEconomy.py
```diff
@@ -45,8 +46,60 @@ def __init__(self, params):
...
+        self.traj["events"] = []          # track active events each step
```
Amittai-Aviram: The name is  not clear and the comment does not help at all.  How can a list be a "trajectory"? A trajectory is the path of an object being thrown (_ject-_) across space (_tra-_). Names should make immediate sense.

Event-Based Simulation/CapitalistEconomy.py
```diff
@@ -45,8 +46,60 @@ def __init__(self, params):
...
+        self.detected_equilibrium = False # whether equilibrium is reached
```
Amittai-Aviram: Is this redundant with **allow random events** declared above? That variable "becomes true after equilibrium." If "after" means "immediately after," then it seems redundant.

Event-Based Simulation/CapitalistEconomy.py
```diff
@@ -45,8 +46,60 @@ def __init__(self, params):
...
+
 
     def step(self):
+        print(f"[STEP DEBUG] t={self.current_t:.0f} | allow_random_events={self.allow_random_events} | event_steps_remaining={self.event_steps_remaining}")
```
Amittai-Aviram: In a future issue, can you replace simple **print** calls with calls to some function that prints only when you pass a special command-line parameter such as **debug** to the executable?

Event-Based Simulation/CapitalistEconomy.py
```diff
@@ -156,7 +221,7 @@ def _get_dydt(self, params):
         alpha_l = params.alpha_l
 
         # Creates the right hand side of the equation dy/dt = f(t,y)
-        # This is what you will want to make alterations to in order to tweak the dynamics of the system. 
+        # This is what you will want to make alterations to in order to tweak the dynamics of the system.
```
Amittai-Aviram: You have two copies of the same line of comment. More importantly, do you still need this comment? Does it make sense in your context? Does it say exactly what it should say? If not, please revise it or eliminate it. Do not let nonsensical comments accumulate in your code!

Event-Based Simulation/CapitalistEconomy.py
```diff
@@ -165,43 +230,64 @@ def rhs(t: float, y: np.ndarray) -> np.ndarray:
...
+                kappa_eff = kappa * eff["kappa"]
+            else:
+                q_eff = q
+                p_eff = p
+                l_eff = l
+                L_eff = L
+                A_eff = A
+                kappa_eff = kappa
 
             # self explanatory
```
Amittai-Aviram: If it is indeed self explanatory, remove the comment.

Event-Based Simulation/random_events.py
```diff
@@ -144,16 +145,53 @@ def load_event_table():
...
+        economy.active_events.remove(ev)
+
+        # War ended -> trigger Post-War Recovery
+        for candidate in economy.events_catalog:
+            if candidate.name == "Post-War Recovery":
+                if getattr(candidate, "follow_after", None) == ev:
+                    candidate.active = True
+                    candidate.remaining_steps = candidate.duration
+                    candidate.applied_once = False
+                    economy.active_events.append(candidate)
```
Amittai-Aviram, line 197: Do you need to label the events at all?

Event-Based Simulation/simulation.py
```diff
@@ -33,25 +37,37 @@
...
+# Global to record continuous equilibrium time (first time detected)
+EQUILIBRIUM_TIME = None
 
 # examples of using the CapitalistEconomy class. 
 def get_trajectories(params):
-    global EQUILIBRIUM_INDEX
+    global EQUILIBRIUM_TIME
     economy = CapitalistEconomy(params)
+    
+    from random_events import load_event_table
```
Amittai-Aviram: Why isn't thist with the other `import` statements at the top of the file?

Amittai-Aviram (commented):
Please address my comments. Thanks!

Event-Based Simulation/CapitalistEconomy.py
Kkenize: I renamed traj to history since this structure stores time series of simulation variables, not a physical “trajectory.” I also separated event logging into event_snapshots to make it clear it’s a per-step record of active events.

Event-Based Simulation/CapitalistEconomy.py
Kkenize: I changed the comment to # will be set to 40 once equilibrium is reached, counts down 40 steps after that.

Event-Based Simulation/CapitalistEconomy.py
Kkenize: I changed the comments to make things clearer. self.allow_random_events is "a toggle that enables random events during a limited post-equilibrium window" and will be switched off after the count down of 40 time steps reaches 0. self.detected_equilibrium is "a one-time-latch that remains true once equilibrium is reached".

Event-Based Simulation/CapitalistEconomy.py
Kkenize: Of course, debug prints are now gated behind a debug flag instead of always printing.

Event-Based Simulation/CapitalistEconomy.py
Kkenize: I've deleted the comment.

Event-Based Simulation/CapitalistEconomy.py
Kkenize: I deleted the comment.

Event-Based Simulation/random_events.py
Kkenize: By labels are you talking about the event identifiers like name/category? I am currently using names for specific case logic like "War" to "Post-War Recovery" and I am using categories for event organization. I guess if I replace it with event ids it should work as well, but I think using strings can help make my code to be easier to read. Please let me know if I'm not interpreting your question correctly or if you think using event ids is better. I will make further changes if that's the case.

Event-Based Simulation/simulation.py
Kkenize: I was thinking about lazy import. But yeah moving it to the top makes it more consistent with the other files while the behavior of the code remains unchanged. As a result, I've moved it to the top in the updated code.

Event-Based Simulation/CapitalistEconomy.py
Amittai-Aviram: Good!

Event-Based Simulation/CapitalistEconomy.py
Amittai-Aviram: Good.

Event-Based Simulation/random_events.py
Amittai-Aviram: Yes, I think you interpreted my question correctly. It seemed to me that events can be treated abstractly without our having to give them realistic names. But perhaps you can revisit this question later.

Amittai-Aviram (approved):
LGTM
Please remember to update your local copy of main and merge that into your local branch and then push. Then merge this PR into main on GitHub. (See the Wiki page, "Git Workflow," for details.)

PR #62, opened 2025-12-14
Implement fluctuating Person health status, productivity considers health status, which Firm looks at

### LINKED ISSUE #40: Implement Person::on_time_step (Devin)
Please implement `on_time_step` for the `Person` class:
- Age
- Shop if appropriate
- Change health status occasionally
- Anything else you can think of.

### LINKED ISSUE #45: Add health status as a consideration in allocating workers (Devin)

lennyyyyyyyy:
Resolves #40 
Resolves #45 

Also:
 - Adds INITIAL_WORK_DAYS_WEEKLY and only executes plan on working days
 - Fix math with new unit based Product quantity, inputs, and living labor


Agent-Based Simulation/src/Firm.cpp
```diff
@@ -11,16 +11,23 @@ double Firm::get_avg_productivity() {
...
-        Person * person,
-        std::vector<Ability>& required_abilities
-        ) {
-    double suitability = 0.0;
-    for (Ability ability : required_abilities) {
-        suitability += person->get_abilities()[ability];
-    }
-    suitability /= required_abilities.size();
-    return suitability;
+double Firm::suitability(Person * person,
```
Amittai-Aviram: Style: thanks for avoiding long lines. (I need to add that to the style guide.) Please break up the parameter list as follows:
```
double Firm::suitability(
    Person * person,
    std::vector<Ability>& required_abilities
    ) {
    return suitabilty(
        person->get_abilities(),
        required_abilities,
        person->get_current_productivity()
    );
}
```
The open and close parentheses go on lines by themselves and the list has one item per line at one further indentation level.

Amittai-Aviram (approved):
Super-excellent work!
Please just fix the style issue as noted in my comment—please apply that rule to all long parameter lists and long conditional expressions.

Agent-Based Simulation/src/Firm.cpp
lennyyyyyyyy: Done.

PR #68, opened 2025-12-20
Maintain a Firm's inventory by demand signals

### LINKED ISSUE #63: Maintain a Firm's inventory by demand signals (Devin)
## In Firm
 - keep a queue<tuple<Product *, int, int>> storing demand signals, kept so it stores the signals from the last N timesteps
 - This updates a demand unordered_map<Product *, int> which is the average demand per timestep over the window
 - This demand is used in `check_and_reorder` to set a threshold where products will be ordered if it falls below the threshold
 - At the beginning of time, the threshold still starts at the order_size

## In Producer
 - Recieving an order checks to see if there are enough inputs. If there aren't, Producer flat out declines. If there are, things proceed like normal.
 - Regardless of whether or not order proceeds, order _requests_ are accounted in the demand vector

## In Distributor
 - The demand vector is influenced by rate of purchases of consumers (instead of the threshold being contant)

## Other
 - At the beginning of time, Producers and Distributors start with the order_size of its inputs/stock (or some multiple of that) so that there is no chain reaction of declining and waiting

lennyyyyyyyy:
Resolve #63


Agent-Based Simulation/header/Distributor.h
```diff
@@ -25,4 +26,5 @@ class Distributor : public Firm {
     bool is_overproduced(Product * product);
```
Amittai-Aviram, line 26: Are these last two items relics of Sal's project from summer 2025?  If so, can we get rid of them?  I know that this is not part of the issue at hand, but we might as well clean up as we go.

Agent-Based Simulation/src/Distributor.cpp
```diff
@@ -8,7 +8,15 @@
...
 Distributor::Distributor() : Firm() {}
 
+Distributor::Distributor(std::unordered_set<Product *> initial_catalog) :
+    Firm(initial_catalog) {
+    for (Product * p : initial_catalog) {
+        inventory[p] = p->order_size * FIRM_INITIAL_INVENTORY_MULTIPLIER;
+    }
+}
+
 void Distributor::on_time_step() {
```
Amittai-Aviram, line 18: This function has some serious issues right now, so I've written up Issue #69 , but have not assigned it yet. I'm waiting tofind out who is willing to take it. (I can take it if necessary.)

Agent-Based Simulation/src/Distributor.cpp
```diff
@@ -73,4 +83,6 @@ bool Distributor::is_overproduced(Product* product) {
     return false;
 }
 
-
+std::unordered_set<Product *> Distributor::get_products_to_reorder() {
+    return catalog;
```
Amittai-Aviram, line 87: I do not understand this logic. Are you using the catalog, not to hold all products distributed by the distributor, but to hold all products that need to be reordered from producers?

Agent-Based Simulation/src/Firm.cpp
```diff
@@ -30,15 +40,12 @@ void Firm::add_supplier(Producer * producer) {
...
-void Firm::set_reorder_threshold(Product * product, int threshold) {
-    reorder_thresholds[product] = threshold;
-}
-
-void Firm::receive_shipment(Product * product, int quantity) {
-    inventory[product] += quantity;
-    std::cout << "Received " << quantity << " units of " 
-              << product->product_name << ". New inventory: " 
-              << inventory[product] << std::endl;
+void Firm::receive_order(Order * order) {
```
Amittai-Aviram, line 43: I think we are running into a terminology problem, because "receiving an order" can now mean either (a) getting a request from a consumer or a downstream firm for some number of some product, or (b) getting a shipment of the finished products that this firm has ordered. We should probably go back and rename _order_ in the (a) case as _request_ and may even change _order_ to _shipment_ in the second case. The data structure holding the request should still be called an _Order_. I guess this calls for a discussion, so I'll hold off on writing an issue about it.

Agent-Based Simulation/src/Firm.cpp
```diff
@@ -30,15 +40,12 @@ void Firm::add_supplier(Producer * producer) {
...
+    inventory[order->product] += order->quantity;
+    product_to_outbound_orders[order->product].erase(order);
+    std::cout << "Received " << order->quantity << " units of " 
+              << order->product->product_name << ". New inventory: " 
+              << inventory[order->product] << std::endl;
```
Amittai-Aviram, line 48: I think you mean "New inventory level." An inventory is a stockpile of supplies — it's actually the collection of the stockpiles of all of the supplies.

Agent-Based Simulation/src/Firm.cpp
```diff
@@ -73,22 +83,33 @@ Producer * Firm::send_order(Order * order) {
 }
 
 void Firm::check_and_reorder() {
```
Amittai-Aviram, line 85: The logic looks pretty good here, but this function has now grown just a bit too big. I've just written Issue #70 and assigned it to you.

Amittai-Aviram (approved):
LGTM!

Agent-Based Simulation/header/Distributor.h
lennyyyyyyyy: I'm not really sure what get_output_ratio does. For is_overproduced, I think Sal made this in anticipation for feedback signals to send back to the Producers, but I'm not sure its really applicable in our setup (since Producers don't decide themselves how much to produce, a Distributor orders exactly how much it needs).

I don't think they're from Sal's summer project, though.

Agent-Based Simulation/src/Distributor.cpp
lennyyyyyyyy: I agree.

Agent-Based Simulation/src/Distributor.cpp
lennyyyyyyyy: I've abstracted reordering logic to Firm, but to do this, I need information on what to reorder (demand signals aren't enough to tell me this, because otherwise any product with 0 demand, even if it is completely unrelated, would be bumped up to the default threshold of order_size).

For both Producer and Distributor, `catalog` represents what products it offers. However, the products that need to be replenished differ, so I made a virtual method in Firm that both can override.

Distributor's reordering set is just the same as its catalog.
Producer's reordering set is the set of all the _inputs_ of its catalog.

Agent-Based Simulation/src/Firm.cpp
lennyyyyyyyy: Ok. I think the `Order` data structure still holds all the relevant information needed for cases (a) and (b), so it might be good to just distinguish these as "receiving a request order" and "receiving a finished order". We could rename the methods like this, and also maybe add a boolean in the data structure that denotes if an Order is a request or finished.

Lmk what you think.

Agent-Based Simulation/src/Firm.cpp
lennyyyyyyyy: I think this is Sal's, but I'll keep that in mind.

Agent-Based Simulation/src/Firm.cpp
lennyyyyyyyy: Got it.

PR #71, opened 2025-12-23
Compute machinery cost in plan drafting; align with Issues #50, #34 & #69

### LINKED ISSUE #34: Compute estimated machinery costs for a draft plan (Sal)

### LINKED ISSUE #50: Add logic in drafting a plan to set the value of the "machinery" (m) account (Sal)
When an agent orders a product, the firm receiving the order should first allocate workers in the draft plan and then compute the "m" value for the plan as 
cost_of_machine / lifespan = cost_per_hour
m = cost_per_hour * (living_labor / num_workers)

Please also see Issue #34 . Please note that the PR should say that it "Resolves" this issue (50) and also "Resolves" (on a separate line) #34.

### LINKED ISSUE #69: Fix `Distributor::on_time_step` (Sal)
`Distributor::on_time_step` is incorrect and otherwise problematic:
- Use range-based iteration.
- `prd` should not be changed at all until the plan is completed.
  - After a plan is completed, `prd` is incremented only by the value of every product transferred to a consumer. If the entire order is shipped to a distributor or another producer (and assuming that the quantity of product in the plan equals the size of the original order), then `prd` is set to 0 when all the products are transferred to the recipient. `prd` is a record of the debt to future consumers. When you transfer your products to consumers, you "pay off" that debt.
- Do not delete a plan in `on_time_step`! If the plan is completed _and all the product items have been shipped_, i.e., transferred to consuming entities, then it might make sense to delete the plan, but we may decide to push the plan onto a history of previous plans, as we did last summer.  That's for a separate issue.

Sal47506:
### Summary
- Compute machinery cost (m) in Producer::draft_order after workers are assigned: m = (Σ cost_of_machine/lifetime) * (labor_hours / workers.size()).
- Use machine cost/lifetime to derive per-hour machinery cost, matching the “all machines used equally” assumption.
- Leave runtime execution logic unchanged; this only corrects plan drafting to include machinery cost.

### Details
- Added machinery cost calculation post worker assignment to reflect true per-worker machine load.
- Included Machine.h in Producer.cpp to access cost and lifetime.
- No changes to existing execution-time loops or hour deductions.

### Updates to Resolve Issue #69

- Updated `Distributor::on_time_step` to ensure `prd = Plan::total_hours` when `plan->units_purchased == plan->total_quantity`
- Updated to use range_based iteration
- Removed deletion of plan  

Resolves #50 
Resolves #34
Resolves #69


Agent-Based Simulation/src/Distributor.cpp
```diff
@@ -17,26 +17,16 @@ Distributor::Distributor(std::unordered_set<Product *> initial_catalog) :
 
 void Distributor::on_time_step() {
     Firm::on_time_step();
+    double cost_per_hour = 0.0;
     for (auto iter = plans_in_progress.begin(); iter != plans_in_progress.end(); ++iter) {
```
Amittai-Aviram: Can you please change this to range-based iteration? (Clean as you go.)

Agent-Based Simulation/src/Distributor.cpp
```diff
@@ -17,26 +17,16 @@ Distributor::Distributor(std::unordered_set<Product *> initial_catalog) :
...
         
-
-        double m = (static_cast<double>(plan->labor_hours) / plan->workers.size()) * machines.size();
-
-        double hours_per_machine = m / machines.size();
-        for (auto * machine : machines) {
-            machine->hours_used -= hours_per_machine;
-        }
-        
-        plan->prd += m + (plan->labor_hours - plan->labor_hours_remaining);
```
Amittai-Aviram, line 38: These deletions are good. Actually, we should now add back logic to deduct from each machine's remaining usefl life upon every hour of use (here, in `on_time_step`). But we should do that hand-in-hand with re-ordering a machine when its remaining useful life reaches 0, and that belongs in a separate issue.

Agent-Based Simulation/header/Machine.h
```diff
@@ -6,7 +6,8 @@
 #include "Product.h"
 
 struct Machine : public Product {
-    Machine(const std::string& name, int lifetime);
+    Machine(const std::string& name, int lifetime, double cost_of_machine);
     const int lifetime;
     int hours_used = 0;
```
Amittai-Aviram: Please rename `hours_used` to `hours_remaining`, and adjust any logic accordingly. This value should equal `lifetime` at construction and then tick down as the machine is used. When `hours_remaining` reaches 0, the firm must replace the machine.

Agent-Based Simulation/src/Producer.cpp
```diff
@@ -45,11 +46,23 @@ int Producer::draft_order(Order * order) {
 		catalog.insert(order->product);
 	}
 	Plan * draft_plan = new Plan();
-    draft_plan-> m = draft_plan->labor_hours / draft_plan->workers.size();
 	draft_plan->order = order;
 	draft_plan->firm = this;
 	draft_optimal_plan(draft_plan, order->product->required_abilities);
 
+    // compute machinery cost component after workers are assigned (issue #50)
```
Amittai-Aviram: Please remove this comment. (We are avoiding comments as much as possible, in favor of self-explanatory names and short functions.)

Agent-Based Simulation/src/Producer.cpp
```diff
@@ -45,11 +46,23 @@ int Producer::draft_order(Order * order) {
...
+    double cost_per_hour = 0.0;
```
Amittai-Aviram: Please rename to `machinery_cost_per_hour`.

Agent-Based Simulation/src/Producer.cpp
```diff
@@ -45,11 +46,23 @@ int Producer::draft_order(Order * order) {
...
+    for (auto * machine : machines) {
```
Amittai-Aviram: Let's avoid `auto` from now on, unless the type in question is annoyingly complex, e.g., std::pair<std::unordered_map<std::string, int>&, std::set<std::string>> — that sort of thing. It is usually helpful to the engineer to know the type of any datum.

Amittai-Aviram (changes requested):
Almost there!

Agent-Based Simulation/src/Distributor.cpp
ghost: Changed to `for (Plan * current_plan_in_progress : plans_in_progress)`

Agent-Based Simulation/src/Distributor.cpp
ghost: Adding the logic for Issue #69

Agent-Based Simulation/header/Machine.h
ghost: renamed `hours_used` to `hours_remaining`

Agent-Based Simulation/src/Producer.cpp
ghost: renamed to `machinery_cost_per_hour` and removed the comment above

Agent-Based Simulation/src/Producer.cpp
ghost: Removed `auto` and changed to `Machine * machine`

ghost:
### Updates to Resolve Issue #69

- Updated `Distributor::on_time_step` to ensure `prd = Plan::total_hours` when `plan->units_purchased == plan->total_quantity`
- Updated to use range_based iteration
- Removed deletion of plan  

Resolves #69

Agent-Based Simulation/header/Firm.h
```diff
@@ -34,6 +34,7 @@ struct Plan {
     int labor_hours_remaining;
     double raw_materials_remaining;
     double total_hours_remaining;
+	int units_purchased;
```
Amittai-Aviram: This means the number of units that this firm produced and that have been consumed _from_ this firm _by_ some other agent, such as another firm or an individual consumer — correct? If so, please change the name to `outgoing_units_consumed` to make it less ambiguous.

Agent-Based Simulation/header/Machine.h
```diff
@@ -6,7 +6,8 @@
 #include "Product.h"
 
 struct Machine : public Product {
-    Machine(const std::string& name, int lifetime);
+    Machine(const std::string& name, int lifetime, double cost_of_machine);
     const int lifetime;
-    int hours_used = 0;
+    int hours_remaining = 0;
+    double cost_of_machine;
```
Amittai-Aviram, line 12: Sorry — I think this particular change (`cost_of_machine`) has been superseded by my changes now under review. Please see PR #80.  The price per unit of the machine is already recorded as `price_per_unit`, since a machine is a product (`Machine` inherits `Product`). The initial price at the beginning of time is set by `Society` in my PR. After that, whenever a batch of those machines is produced and offloaded to a consumer or distributor, the Price Controller will update its cost based on the real cost of its inputs and the real expenditure of living labor. So it would never be set through the constructor. I think the rest of this should work fine.  The addition of `hours_remaining` is appropriate.

Agent-Based Simulation/src/Distributor.cpp
Amittai-Aviram: LGTM

Agent-Based Simulation/src/Machine.cpp
```diff
@@ -2,6 +2,6 @@
 
 #include "Machine.h"
 
-Machine::Machine(const std::string& name, int lifetime) 
-    : Product{name}, lifetime{lifetime} {}
+Machine::Machine(const std::string& name, int lifetime, double cost_of_machine) 
+    : Product{name}, lifetime{lifetime}, cost_of_machine{cost_of_machine} {}
```
Amittai-Aviram: See above re cost.

Agent-Based Simulation/src/Producer.cpp
```diff
@@ -45,11 +46,22 @@ int Producer::draft_order(Order * order) {
...
 	}
 	Plan * draft_plan = new Plan();
-    draft_plan-> m = draft_plan->labor_hours / draft_plan->workers.size();
 	draft_plan->order = order;
 	draft_plan->firm = this;
 	draft_optimal_plan(draft_plan, order->product->required_abilities);
 
+    double machinery_cost_per_hour = 0.0;
+    for (Machine * machine : machines) {
+        machinery_cost_per_hour += machine->cost_of_machine / machine->lifetime;
```
Amittai-Aviram: This is close, but not quite right. Again, please see PR #80.  The cost for a given machine per hour is the ratio of the number of hours during which it is used (which is living labor for this product divided by number of workers) over the machine's lifetime (that's the portion of the machine that we have used in order to produce one unit of the current product), times the original price of the machine (which is the `price_per_unit` of that machine as a product). I think the rest of your logic is correct.  That is, we take the portion of the machine's lifetime that we are using to make a unit of our new product, which is hours divided by hours to give a pure number (a fraction), and multiply that by the cost (price) of the machine, which is in hours, to give us a result in hours, the cost of using this machine to make one unit of our output product.

Amittai-Aviram (changes requested):
Good work! Please see my comments. You might want to see my PR #80 first before going through my comments here, to give you a better context.

Agent-Based Simulation/header/Machine.h
ghost: removed cost_of_machine

Agent-Based Simulation/header/Firm.h
ghost: changed to outgoing_units_consumed

Agent-Based Simulation/src/Producer.cpp
ghost: Changed to `machinery_cost_per_hour += machine->price_per_unit / machine->lifetime;`

Agent-Based Simulation/src/Machine.cpp
ghost: Changed to `Machine::Machine(const std::string & name, int lifetime) 
    : Product{name}, lifetime{lifetime} {}`

Sal47506 (commented):
Updated machinery cost logic alongside other small stylistic changes

Amittai-Aviram (approved):
LGTM

Great work! 😃

Issue #73, opened 2025-12-26
Rename `receive_order` (Devin)


Amittai-Aviram:
Insofar as "receive order" in `Firm` actually means "receive finished order," please change this name to `receive_shipment` or `receive_finished_order`. I am in favor of the former name, but you may choose either one, so long as you update all the calls.  I could only find this one function that uses the word "receive" in its name, but please check the other functions related to orders and make sure that their names make it clear whether the order is a request or a finished batch of goods, and make sure that the name reflects the status of the order.

Also, please replace the phrase "New inventory" in the trace printout in (what is currently called) `Firm::receive_order` with "New inventory level."

Finally, please add an enumeration type, `OrderStatus`, to `Order`, with enumerators `ORDER_REQUESTED` and `ORDER_FINISHED`, and a field `status` of that type. Set the field appropriately throughout the order's lifetime.

PR #74, opened 2025-12-26
Refactor `Firm::check_and_reorder`, rename `recieve_order`, fix Ability placement

### LINKED ISSUE #70: Break up `Firm::check_and_reorder` (Devin)
`Firm::check_and_reorder` has grown beyond the unwritten size threshold. Let's break it up into two or more helper functions and call those in turn.  The names of the helper functions should help keep the flow of logic clear.

lennyyyyyyyy:
Resolves #70, #73 

I found that I was able to move Ability back to Person without any issues (probably because of the better header organization), so I did that, changing `Ability` into `Person::Ability`.

### DIFF
--- a/Agent-Based Simulation/header/Firm.h
+++ b/Agent-Based Simulation/header/Firm.h
@@ -78,8 +78,12 @@ class Firm : public Agent {
     std::unordered_map<Product*, std::vector<Plan*>> plan_history; // unused and prob need to change later
     std::vector<Plan*> plans_in_progress;
 
-    Producer * find_producer_for_product(Product * product);
     Producer * send_order(Order * order);
+    double get_reorder_threshold(Product * product);
+    int get_pending_inventory(Product * product);
+    void reorder_product_to_threshold(Product * product, 
+                                      double threshold,
+                                      int pending_inventory);
     void check_and_reorder();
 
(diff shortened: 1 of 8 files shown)


Agent-Based Simulation/header/Firm.h
```diff
@@ -78,21 +80,25 @@ class Firm : public Agent {
     std::unordered_map<Product*, std::vector<Plan*>> plan_history; // unused and prob need to change later
     std::vector<Plan*> plans_in_progress;
 
-    Producer * find_producer_for_product(Product * product);
     Producer * send_order(Order * order);
+    double get_reorder_threshold(Product * product);
+    int get_pending_inventory(Product * product);
+    void reorder_product_to_threshold(Product * product, 
+                                      double threshold,
+                                      int pending_inventory);
```
Amittai-Aviram: Please change indentation:
```
void reorder_product_to_threshold(
    Product * product,
    double threshold,
    int pending_inventory
);
```

Amittai-Aviram (approved):
GREAT! I asked for one style change. Please just make the change and then go ahead and merge. (This is the benefit of requring that the author be the one to merge and that the author first have to resolve all comments.)

PR #76, opened 2025-12-30
Compute  initial prices based on Leontief input-output model

### LINKED ISSUE #67: Set initial living labor, input quantities, and prices (Amittai)
There is no need to avoid supply chain cycles, so long as each required input quantity per unit of output product is < 1.0.

### LINKED ISSUE #77: Ensure positive initial prices (Amittai)
To ensure that initial prices are always positive (since they derive from pseudorandom numbers), we should first check whether the maximum eigenvalue of the input-output matrix A is < 1.0. If it is not, we should divide the entries by that eigenvalue and subtract an additional small value epsilon.

Amittai-Aviram:
`Society` now sets all product names, required input quantities, required living labor, and prices as initial conditions of the simulation. After this, prices may change according to average actual living labor.

This change set requires addition of the Eigen linear algebra library to our `header` directory.

Resolves #67 
Resolves #77


Agent-Based Simulation/src/Society.cpp
```diff
@@ -39,11 +47,99 @@ Society::Society() {
...
+        labor_vector(product_to_index[output_product]) =
+            output_product->living_labor_per_unit;
+    }
+}
+
+double get_max_eigenvalue(Eigen::MatrixXd& io_matrix) {
+    Eigen::EigenSolver<Eigen::MatrixXd> eigen_solver(io_matrix, false);
+    Eigen::VectorXcd eigenvalues = eigen_solver.eigenvalues();
+    double max_eigenvalue = 0.0;
+    for (size_t i = 0; i < eigenvalues.size(); ++i) {
```
alexbcreiner0, line 90: Any non-negative matrix is guaranteed to have a strictly real eigenvalue equal to the spectral radius, so this should work just fine.

Amittai-Aviram: Great! Thank you! I am confused, though (sorry!). Did you mean that I need to change anything? If not, would you be willing to approve the PR? Thanks!

alexbcreiner0 (approved):
This looks great, I would only add that there should be an option to declare input and living labor requirements manually (as opposed to randomly). We'll need that to test hypotheses about economies in fixed technological states. 

As long as the living labor vector is strictly positive, the value vector will be too. If we ever drop this assumption, then we will need to add tools to ensure and detect if the matrix is irreducible. Shouldn't be an issue right now though.

Amittai-Aviram:
Yes, that is a good idea. I have created Issue #78 accordingly.

PR #88, opened 2026-01-17
Event based simulation prints csv and trade interface initial structure built

Kkenize:
Event based simulation prints csv and trade interface initial structure built. The version is still not final yet though so reviews can happen a little later.


Amittai-Aviram (commented):
Please replace the name of this PR with a descriptive one. See the other (open and closed) PRs for examples to follow.

Event-Based Simulation/simulation.py
```diff
@@ -5,8 +5,9 @@
 import os
 import matplotlib.pyplot as plt
 import numpy as np
+import csv
 
-# Making folder for graphs 
+# Making folder for graphs and the csv file with the data 
 os.makedirs("graphs", exist_ok=True)
```
Amittai-Aviram: It looks as if the graphs director is part of the Git repository. If so, you should not have any code to create it. It will be there until and unless you manually remove it from the Git repository.

Event-Based Simulation/simulation.py
```diff
@@ -192,6 +193,55 @@ def plot_graphs(traj, t, equilibrium_time):
     plot_prices(time, prices, product_labels, equilibrium_time)
     plot_outputs(time, outputs, product_labels, equilibrium_time)
     plot_averages(time, avg_p, avg_q, equilibrium_time)
```
Amittai-Aviram: What do p and q represent?  Are they just two arbitrary products with independent prices?

Trade Interface/simulation.py
```diff
@@ -0,0 +1,142 @@
...
+                break
+
+    # Safety check: equilibrium must exist
+    if eq_row is None:
+        raise ValueError("No equilibrium row found (is_equilibrium_step == 1).")
+
+    ce_map = {}
+
+    # Build CE product objects for A–Z using equilibrium prices
+    for i in range(26):
```
Amittai-Aviram: Please replace the numeric literal 26 with a global symbolic constant and use the same symbolic constant everywhere instead of the number 26.

Amittai-Aviram (changes requested):
Looks good so far. Please see comments.

Event-Based Simulation/simulation.py
Kkenize: I have deleted folder making.

Event-Based Simulation/simulation.py
Kkenize: I have changed it to avg_price and avg_output.

Trade Interface/simulation.py
Kkenize: I have replaced it with two global variables CE_NUMBER_OF_PRODUCT_TYPES and LTE_NUMBER_OF_PRODUCT_TYPES.

Amittai-Aviram (approved):
LGTM

PR #89, opened 2026-01-17
Add event traces for Producer and Person + Bugfixes + Underproduction feedback

### LINKED ISSUE #86: Print a trace of events and time steps for Producer and Person (Devin)
Write `cout` log statements in important methods of Producer and Person, including:
 - `on_time_step` (for both Producer and Person)
 - `draft_order`
 - `execute_plans`
 - `shop`

### LINKED ISSUE #92: Firm produces nonsensical output from `on_time_step` (Sal)
Please update your local `main` branch (`git switch main`; `git pull`) and then rebuild the project (`make clean && make`). Then run the simulation (`bin/sim`). Note the huge and nonsensical values, including occasional negative values. This looks like the possible result of one or more uninitialized variables.

### LINKED ISSUE #97: Add feedback mechanism for underproduction (Devin)
Currently, underproduction causes shortages of inventory and new orders having no workforce to allocate. 

Solution:
Every shopping event, have people buy only as much as they have in their account. The amount of each product bought will be proportional to the amount they would have bought if they had enough hours in their account. If they do not have enough, each amount is scaled down by some factor.

lennyyyyyyyy:
Resolve #86 

# Bugfixes - resolves #92 
 - Fixed rounding of reorder discrepancy up to the nearest order size, instead of down.
 - Fixed setting the initial inventories to be above initial thresholds at the beginning of time
 - Fixed demand signals being sent to every queried Producer when reordering, causing exploding demand at the beginning of time
 - Fixed double drafting when querying Producers
 
Resolve #97


Amittai-Aviram (commented):
LGTM
Please hold off on merging until we have fixed some bugs that have become evident through the `Firm` trace.  Accordingly, I will hold off on marking this PR as "approved" until that point.

Agent-Based Simulation/header/Constants.h
```diff
@@ -14,8 +14,8 @@ const int INITIAL_WORK_DAYS_WEEKLY = 5;
 const int INITIAL_WORK_WEEK = INITIAL_WORK_HOURS_DAILY * \
                               INITIAL_WORK_DAYS_WEEKLY;
 
-const int STARTING_NUM_PEOPLE = 1000;
-const int STARTING_NUM_PRODUCTS = 100;
+const int STARTING_NUM_PEOPLE = 100;
+const int STARTING_NUM_PRODUCTS = 10;
```
Amittai-Aviram: This change is temporary just to support debugging, isn't it? Shouldn't you restore the original values now and test? Or do you want to wait until you have fixed the overconsumption issue?

Agent-Based Simulation/header/Constants.h
```diff
@@ -29,18 +29,18 @@ const double AVERAGE_RETIREMENT_AGE = (
     INITIAL_AGE; // approx 62 years in hours
 
 const int PRODUCT_ORDER_SIZE_MIN = 1;
-const int PRODUCT_ORDER_SIZE_MAX = 1000;
+const int PRODUCT_ORDER_SIZE_MAX = 100;
```
Amittai-Aviram: See above.

Agent-Based Simulation/header/Constants.h
```diff
@@ -29,18 +29,18 @@ const double AVERAGE_RETIREMENT_AGE = (
...
 const int PRODUCT_NUM_INPUTS_MIN = 0;
-const int PRODUCT_NUM_INPUTS_MAX = 10;
+const int PRODUCT_NUM_INPUTS_MAX = 2;
```
Amittai-Aviram: Ditto.

Agent-Based Simulation/header/Constants.h
```diff
@@ -49,7 +49,7 @@ const int MACHINE_LIFETIME_MAX = MACHINE_LIFETIME_MIN * 10;
 const int FIRM_TRAINING_TIME = INITIAL_WORK_HOURS_DAILY * WEEK / DAY;
 const int FIRM_STOCKPILE_DURATION = MONTH;
 const int FIRM_DEMAND_WINDOW = WEEK;
-const double FIRM_INITIAL_INVENTORY_MULTIPLIER = 2;
+const double FIRM_INITIAL_INVENTORY_MULTIPLIER = 1;
```
Amittai-Aviram, line 53: So my comments on the first change in this file apply to the whole file. If you want to scale the society down for debugging purposes, that's OK, but either restore the original numbers within this same PR or else copy the original file as `Constants_orig.h`, add it to the repo, and write an issue to rename `Constants.h` as `Constants_dbg.h` and `Constants_orig.h` as `Constants.h`.

Agent-Based Simulation/src/Distributor.cpp
```diff
@@ -10,8 +10,8 @@ Distributor::Distributor() : Firm() {}
 
 Distributor::Distributor(std::unordered_set<Product *> initial_catalog) :
     Firm(initial_catalog) {
-    for (Product * p : initial_catalog) {
-        inventory[p] = p->order_size * FIRM_INITIAL_INVENTORY_MULTIPLIER;
+    for (Product * p : get_products_to_reorder()) {
```
Amittai-Aviram, line 13: Does this work because all the inventories are empty at this point in the execution of the constructor?

Agent-Based Simulation/src/Firm.cpp
```diff
@@ -65,11 +65,13 @@ Producer * Firm::send_order(Order * order) {
     int order_time = INT_MAX;
     Producer * chosen_producer = nullptr;
 
+    //std::cout << "Draft order times: " << std::endl;
```
Amittai-Aviram: Uncomment or remove.

Agent-Based Simulation/src/Firm.cpp
```diff
@@ -65,11 +65,13 @@ Producer * Firm::send_order(Order * order) {
...
     for (auto * producer : suppliers) {
         int draft_order_time = producer->draft_order(order);
+        //std::cout << draft_order_time << std::endl;
```
Amittai-Aviram: Ditto. Looks as if you should remove it.

Agent-Based Simulation/src/Firm.cpp
```diff
@@ -65,11 +65,13 @@ Producer * Firm::send_order(Order * order) {
     int order_time = INT_MAX;
     Producer * chosen_producer = nullptr;
 
+    //std::cout << "Draft order times: " << std::endl;
     for (auto * producer : suppliers) {
```
Amittai-Aviram: Please replace `auto` with `Producer`.

Agent-Based Simulation/src/Firm.cpp
```diff
@@ -65,11 +65,13 @@ Producer * Firm::send_order(Order * order) {
...
         int draft_order_time = producer->draft_order(order);
```
Amittai-Aviram, line 69: This is slightly out of scope, but why is `draft_order` called `draft_order`? Shouldn't it be called `draft_plan`? You're sending an order in to a potential supplier (producer), asking it to draft a plan, and getting back the proposed turnaround time according to that draft plan — correct?

Agent-Based Simulation/src/Firm.cpp
```diff
@@ -106,8 +108,8 @@ void Firm::reorder_product_to_threshold(
         ) {
     if (pending_inventory < threshold) {
         int discrepancy = product->order_size * 
-            ((int) std::ceil(threshold - pending_inventory) /
-             product->order_size);
+            ((int) std::ceil((threshold - pending_inventory) /
+             product->order_size));
```
Amittai-Aviram: Correct. Subtract, then divide, then take the ceiling.  It might also be helpful (documentation-wise) to cast (using `static_cast<double>`) `pending_inventory` to `double` to show that the subtraction and, therefore, the division are in rational-number arithmetic.

Agent-Based Simulation/src/Person.cpp
```diff
@@ -141,6 +146,12 @@ void Person::on_time_step() {
 	if (will_shop()) { shop(); }
 	if (will_retire()) { retire(); }
 	update_health_status();
+    //std::cout << "At time step " << Sim::get_current_time_step() 
+    //          << ", Person has:" << std::endl;
+    //std::cout << "\tAge: " << age << std::endl;
+    //std::cout << "\tAccount balance: " << account << std::endl;
+    //std::cout << "\tHealth status: " << 
+    //  (health_status == HEALTHY ? "Healthy" : "Unhealthy") << std::endl;
 }
```
Amittai-Aviram, line 167: Remove, right?

Agent-Based Simulation/src/Producer.cpp
```diff
@@ -13,31 +14,50 @@ Producer::Producer() : Firm() {}
...
     Firm(initial_catalog) {
-    for (Product * product : initial_catalog) {
-        for (auto& p : product->inputs_per_unit) {
-            inventory[p.first] += p.second *
-                                  product->order_size *
-                                  FIRM_INITIAL_INVENTORY_MULTIPLIER;
-        }
+    for (Product * p : get_products_to_reorder()) {
+        inventory[p] = get_reorder_threshold(p) * FIRM_INITIAL_INVENTORY_MULTIPLIER;
     }
```
Amittai-Aviram, line 19: Just checking: The idea is that all the inventories are empty to start with, so the firm has to "reorder" the products right from the start — correct?

Agent-Based Simulation/src/Producer.cpp
```diff
@@ -13,31 +14,50 @@ Producer::Producer() : Firm() {}
...
 
 int Producer::draft_order(Order * order) {
+    std::cout << "Drafting order for " << order->quantity << " units of "
+         << order->product->product_name << std::endl;
     bool enough_inputs = true;
     for (auto &p : order->product->inputs_per_unit) {
         if (inventory[p.first] < p.second * order->quantity) {
+            //std::cout << "Not enough inventory of "
+            //          << p.first->product_name << ". Has: " << inventory[p.first] 
+            //          << ", needs: " << p.second * order->quantity << std::endl;
```
Amittai-Aviram: Uncomment or remove.

Agent-Based Simulation/src/Producer.cpp
```diff
@@ -75,6 +97,9 @@ bool Producer::drop_order(Order * order) {
 }
 
 bool Producer::pursue_order(Order * order) {
+    for (auto &p : order->product->inputs_per_unit) {
```
Amittai-Aviram: Please replace `auto` with the actual type (`std::pair<Product *, double>&`). The `&` goes with the type name, not the variable name.

Amittai-Aviram (changes requested):
Excellent work so far!

Agent-Based Simulation/header/Constants.h
lennyyyyyyyy: Got it. I'll just revert it. I do think its better to have PRODUCT_CONSUMPTION_FREQUENCY_MIN as 0 and FIRM_INITIAL_INVENTORY_MULTPLIER as 1, however.

Agent-Based Simulation/src/Distributor.cpp
lennyyyyyyyy: get_products_to_reorder() gives the set of all products that are going to have a non-zero threshold. This is extended differently for Producer and Distributor.

For Distributor, looping over initial_catalog or get_products_to_reorder() is actually equivalent, but I just wanted it to be consistent with Producer.

Agent-Based Simulation/src/Producer.cpp
lennyyyyyyyy: I think that's pretty much correct. get_products_to_reorder() doesn't care about if the inventories are empty or not. It's just initializing the set of products it'll be reordering later to be higher than the threshold at the beginning of time. No real ordering happening here though.

Agent-Based Simulation/src/Firm.cpp
lennyyyyyyyy: Yeah. When I first named it, I honestly thought it fit nicer along with the other API functions `drop_order` and `pursue_order`. Let me know if you want me to change it though.

Agent-Based Simulation/src/Distributor.cpp
Amittai-Aviram: So get_products_to_reorder really means "get products that _might_ need to be reordered? So, for the Producer, it's the inputs, and, for the Distributor, it's the products to be distributed?

Agent-Based Simulation/src/Firm.cpp
Amittai-Aviram: Please write a new issue for a name change. Thanks.

Agent-Based Simulation/src/Firm.cpp
Amittai-Aviram: Please add the static cast before merging. (Be sure to build and test-run.)

Amittai-Aviram (approved):
Please see my 3 (I think) responses to your replies, and please add that static cast. Another response asks you to write up a (small) issue. Otherwise, LGTM.

Issue #72, opened 2025-12-26
Clean up Distributor (Sal)


Amittai-Aviram:
`Distributor` has the following functions, which all seem to be obsolete:

1. `get_output_ratio`
2. `planned_satisfaction_per_person`
3. `is_overproduced`

Right now, no code outside of these three calls any of these functions. Please either remove these functions or determine how they will be used in the near future.

PR #93, opened 2026-01-20
Draft Order Issues Fixed

### LINKED ISSUE #92: Firm produces nonsensical output from `on_time_step` (Sal)

Sal47506:
### Summary

- Stabilize draft order selection and worker assignment to avoid INT_MAX turnaround times.
- Clamp reorder calculations to prevent overflow and extreme reorder sizes.
- Ensure producers can draft orders consistently by initializing plans and updating suitability handling.

### Changes

- Fix producer selection in send_order() to use the computed draft time once.
- Guard total_suitability and ensure at least one worker is assigned when possible.
- Clamp reorder thresholds and discrepancy calculations to avoid overflow.
- Initialize Plan objects with zeroed fields.
- Obsolete methods as mentioned in #72 from Distributor are now removed 

IMG_4084

Resolves #92,  #72

### DIFF
--- a/Agent-Based Simulation/src/Firm.cpp
+++ b/Agent-Based Simulation/src/Firm.cpp
@@ -65,11 +65,11 @@ Producer * Firm::send_order(Order * order) {
     int order_time = INT_MAX;
     Producer * chosen_producer = nullptr;
 
-    for (auto * producer : suppliers) {
+    for (Producer * producer : suppliers) {
         int draft_order_time = producer->draft_order(order);
         if (draft_order_time != DRAFT_ORDER_REJECTED &&
-                draft_order_time < order_time) {
-            order_time = producer->draft_order(order);
+            draft_order_time < order_time) {
+            order_time = draft_order_time;
             chosen_producer = producer;
         }
     }
@@ -87,8 +87,9 @@ Producer * Firm::send_order(Order * order) {
 }
 
 double Firm::get_reorder_threshold(Product * product) {
-    return std::max((double) product->order_size, 
-            inventory_demands[product] * FIRM_STOCKPILE_DURATION);
+    double demand = std::min(inventory_demands[product], 10000.0);
+    return std::max((double) product->order_size,
+        demand * FIRM_STOCKPILE_DURATION);
 }
 
 int Firm::get_pending_inventory(Product * product) {
@@ -105,14 +106,19 @@ void Firm::reorder_product_to_threshold(
         int pending_inventory
         ) {
     if (pending_inventory < threshold) {
-        int discrepancy = product->order_size * 
-            ((int) std::ceil(threshold - pending_inventory) /
-             product->order_size);
+        double diff = threshold - pending_inventory;
+        if (diff > INT_MAX / 2) diff = INT_MAX / 2;
+        int units_needed = (int) std::ceil(diff);
+        int discrepancy = product->order_size > 0
+            ? ((units_needed + product->order_size - 1) / product->order_size) * product->order_size
+            : 0;
         std::cout << "Reordering " << discrepancy << " units of " 
             << product->product_name << std::endl;
 
-        Order * order = new Order{product, discrepancy, this,
-            (int) (pending_inventory / threshold * FIRM_STOCKPILE_DURATION),
+        int turnaround = threshold > 0
+            ? (int) (pending_inventory / threshold * FIRM_STOCKPILE_DURATION)
+            : 0;
+        Order * order = new Order{product, discrepancy, this, turnaround,
             Order::ORDER_REQUESTED};
 
         Producer * chosen_producer = send_order(order);
@@ -209,6 +215,16 @@ void Firm::assign_workers_by_suitability_threshold(
             draft_plan->workers.push_back(unemployed_person);
         }
     }
+    if (draft_plan->workers.empty()) {
+        std::vector<Person *> & unemployed_people =
+            Society::get_instance()->get_unemployed_people();
+        if (!unemployed_people.empty()) {
+            draft_plan->workers.push_back(unemployed_people.front());
+        }
+    }
+    if (draft_plan->workers.empty() && !workers.empty()) {
+        draft_plan->workers.push_back(workers.front());
+    }
 }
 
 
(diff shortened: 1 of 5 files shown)


Agent-Based Simulation/src/Firm.cpp
```diff
@@ -65,11 +65,11 @@ Producer * Firm::send_order(Order * order) {
...
     Producer * chosen_producer = nullptr;
 
-    for (auto * producer : suppliers) {
+    for (Producer * producer : suppliers) {
         int draft_order_time = producer->draft_order(order);
         if (draft_order_time != DRAFT_ORDER_REJECTED &&
-                draft_order_time < order_time) {
-            order_time = producer->draft_order(order);
+            draft_order_time < order_time) {
+            order_time = draft_order_time;
```
lennyyyyyyyy, line 72: This makes a lot of sense now. It was drafting an order twice, which was causing the Producer to reject the second one.

Agent-Based Simulation/src/Firm.cpp
```diff
@@ -87,8 +87,9 @@ Producer * Firm::send_order(Order * order) {
 }
 
 double Firm::get_reorder_threshold(Product * product) {
-    return std::max((double) product->order_size, 
-            inventory_demands[product] * FIRM_STOCKPILE_DURATION);
+    double demand = std::min(inventory_demands[product], 10000.0);
```
lennyyyyyyyy, line 90: I think this might only be a temporary solution to exploding demand.

Agent-Based Simulation/src/Firm.cpp
```diff
@@ -105,14 +106,19 @@ void Firm::reorder_product_to_threshold(
         int pending_inventory
         ) {
     if (pending_inventory < threshold) {
-        int discrepancy = product->order_size * 
-            ((int) std::ceil(threshold - pending_inventory) /
-             product->order_size);
+        double diff = threshold - pending_inventory;
+        if (diff > INT_MAX / 2) diff = INT_MAX / 2;
+        int units_needed = (int) std::ceil(diff);
+        int discrepancy = product->order_size > 0
```
lennyyyyyyyy, line 112: I think if the product's order size is ever zero, we should fix the constructor of Product instead of accounting for it here.

Agent-Based Simulation/src/Firm.cpp
```diff
@@ -105,14 +106,19 @@ void Firm::reorder_product_to_threshold(
...
+            ? ((units_needed + product->order_size - 1) / product->order_size) * product->order_size
+            : 0;
         std::cout << "Reordering " << discrepancy << " units of " 
             << product->product_name << std::endl;
 
-        Order * order = new Order{product, discrepancy, this,
-            (int) (pending_inventory / threshold * FIRM_STOCKPILE_DURATION),
+        int turnaround = threshold > 0
```
lennyyyyyyyy, line 118: Same thing here about product's order size (the threshold is bounded from below by the order size)

Agent-Based Simulation/src/Firm.cpp
```diff
@@ -105,14 +106,19 @@ void Firm::reorder_product_to_threshold(
         int pending_inventory
         ) {
     if (pending_inventory < threshold) {
-        int discrepancy = product->order_size * 
-            ((int) std::ceil(threshold - pending_inventory) /
-             product->order_size);
+        double diff = threshold - pending_inventory;
+        if (diff > INT_MAX / 2) diff = INT_MAX / 2;
```
lennyyyyyyyy, line 110: Same idea as line 90, it doesn't really fix the issue of exploding demand. I have a theory on why this is happening that I'll talk about in our meeting today.

Agent-Based Simulation/src/Producer.cpp
```diff
@@ -32,20 +33,19 @@ bool Producer::can_produce(Product * product) {
...
 
 int Producer::draft_order(Order * order) {
-    bool enough_inputs = true;
+	if (order_to_draft_plan[order] != nullptr) {
+		return DRAFT_ORDER_REJECTED;
+	}
     for (auto &p : order->product->inputs_per_unit) {
         if (inventory[p.first] < p.second * order->quantity) {
-            enough_inputs = false;
+            inventory[p.first] = p.second * order->quantity;
```
lennyyyyyyyy, line 41: How can we just change the inventory without receiving it from a supplier?The order should be rejected if there's not enough inputs.

Agent-Based Simulation/src/Firm.cpp
```diff
@@ -209,6 +215,16 @@ void Firm::assign_workers_by_suitability_threshold(
...
+    if (draft_plan->workers.empty()) {
+        std::vector<Person *> & unemployed_people =
+            Society::get_instance()->get_unemployed_people();
+        if (!unemployed_people.empty()) {
+            draft_plan->workers.push_back(unemployed_people.front());
+        }
+    }
+    if (draft_plan->workers.empty() && !workers.empty()) {
+        draft_plan->workers.push_back(workers.front());
+    }
```
lennyyyyyyyy, line 227: I'm not sure why this logic is here. If there aren't any workers in draft_plan, that suggests something wrong with the drafting process. It doesn't make sense to just take random worker from the Society to fill in for that plan either.

lennyyyyyyyy (changes requested):

Agent-Based Simulation/src/Firm.cpp
Amittai-Aviram: But probably `Producer::draft_order` should have been made idempotent, so that, if you call it a second time with the same order, it uses the same draft plan and returns the same draft turnaround time.

Agent-Based Simulation/src/Firm.cpp
Amittai-Aviram: Yeah, this is not how we want to approach such problems in general. Also, the bare numeric literal here sends up a red flag. This is precisely why we should always use symbolic constants as defined in `Constants.h`: in order to make us think twice about using an arbitrary constant rather than some kind of dynamic self-adjustment mechanism.

Agent-Based Simulation/src/Firm.cpp
Amittai-Aviram: Agreed. Also, please note that our style rules require curly braces and line breaks even for single-statement condition bodies:
```
if (diff > INT_MAX / 2) {
    diff = INT_MAX / 2;
}
````

File: Agent-Based Simulation/src/Firm.cpp (hunk 3 above)
Comment (Amittai-Aviram) [reply Agent-Based Simulation/src/Firm.cpp:112]: We should add an `assert` statement somewhere. It's clearly an error. We should also check the order size in a unit test.

File: Agent-Based Simulation/src/Firm.cpp (hunk 7 above)
Comment (Amittai-Aviram) [reply Agent-Based Simulation/src/Firm.cpp:227]: Was this necessary for the bug fix?

File: Agent-Based Simulation/src/Producer.cpp [hunk 8]
```diff
@@ -63,6 +63,7 @@ int Producer::draft_order(Order * order) {
     }
 
 	order_to_draft_plan[order] = draft_plan;
+	//std::cout << "Draft plan predicted turnaround time: " << draft_plan->predicted_turnaround_time << std::endl;
```
Amittai-Aviram, line 66: Either this needs to be uncommented and eventually logged (reasonable idea) or it should be removed.

Amittai-Aviram (commented):
This PR has already been closed, but I went through it anyway and made comments for both Sal and Devin to consider in future planning. I'm about to go through Devin's changes.

PR #99, opened 2026-01-24
Distinguish machines from other inputs; provide producers with necessary machines for initial products

### LINKED ISSUE #94: Distinguish machines from other inputs and provide producers with necessary machines at T0 (Amittai)
- Right now, machines are added to the `Society::products` vector and are therefore incorrectly counted as inputs.
- `Product::machines_needed` is not used to supply actual machines to producers.

Amittai-Aviram:
- Changed the name of `Society::products` to `Society::goods` to distinguish products (_goods_) that can serve as inputs (raw materials) to other products from machines.
  - Goods are kept in inventories and are re-ordered in bulk when an inventory level drops below its re-order threshold.
  - Firms hold only one of each machine, and machines are re-ordered one at a time when they reach the ends of their respective lifetimes.
  - Distributors only distribute goods.
- Propagated the name change to all concerned entities (Producer and Product).
- Kept `Society::products` as the union (or concatenation) of `goods` and `machines`.
- Continued to use `Society::products` for the input-output matrix in order to compute initial prices, including the prices of machines.
- Added logic to the `Producer` constructor (but  not to that of either `Distributor` or `Firm`) to add initial machines to the `machines` vector according to the needs of products in the initial catalogue.  Producers are assumed to need machines to make their initial (and later) products, whereas distributors, for now, are not.

Resolves #94

### DIFF
--- a/Agent-Based Simulation/src/Society.cpp
+++ b/Agent-Based Simulation/src/Society.cpp
@@ -3,6 +3,7 @@
 #include <Eigen/Eigenvalues>
 #include <iostream>
 #include <numeric>
+#include <sstream>
 #include <stdexcept>
 #include <unordered_map>
 
@@ -22,17 +23,17 @@ Society * Society::get_instance() {
 
 Society::Society() {
     set_initial_products();
-    // note: no way to assign products to producers or suppliers
+    // note: no way to assign goods to producers or suppliers
     // to distributors yet
     for (int i = 0; i < STARTING_NUM_PRODUCERS; i++) {
-        Producer * producer = new Producer({products[i %
+        Producer * producer = new Producer({goods[i %
                 STARTING_NUM_PRODUCTS]});
         producers.push_back(producer);
         firms.push_back(producer);
     }
     for (int i = 0; i < STARTING_NUM_DISTRIBUTORS; i++) {
         Distributor * distributor =
-            new Distributor({products[i % STARTING_NUM_PRODUCTS]});
+            new Distributor({goods[i % STARTING_NUM_PRODUCTS]});
         distributors.push_back(distributor);
         firms.push_back(distributor);
     }
@@ -44,7 +45,7 @@ Society::Society() {
             }
         }
     }
-    // people MUST come after products and distributors are created
+    // people MUST come after goods and distributors are created
     for (int i = 0; i < STARTING_NUM_PEOPLE; i++) {
         birth_person();	
     }
@@ -54,6 +55,7 @@ void Society::set_initial_products() {
     std::size_t i = 0;
     for (; i < STARTING_NUM_PRODUCTS; ++i) {
         Product * new_product = new Product("Product " + std::to_string(i));
+        goods.push_back(new_product);
         products.push_back(new_product);
         product_to_index[new_product] = i;
     }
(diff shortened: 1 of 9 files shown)


Agent-Based Simulation/src/Society.cpp
```diff
@@ -22,17 +23,17 @@ Society * Society::get_instance() {
 
 Society::Society() {
     set_initial_products();
-    // note: no way to assign products to producers or suppliers
+    // note: no way to assign goods to producers or suppliers
```
lennyyyyyyyy: I think this comment is mine, but is now outdated. Can you just remove it here?

Agent-Based Simulation/src/Society.cpp
```diff
@@ -68,9 +70,9 @@ void Society::set_initial_products() {
         products.push_back(new_machine);
         product_to_index[new_machine] = i;
     }
-    for (Product * product: products) {
-        product->set_inputs(products);
-        product->set_machines(machines);
+    for (Product * good: goods) {
+        good->set_inputs(goods);
+        good->set_machines(machines);
```
lennyyyyyyyy: I think it makes more sense to initialize all products with goods and machines (instead of only goods), since machines need a production process as well, right?

Agent-Based Simulation/src/Society.cpp
```diff
@@ -156,8 +158,10 @@ void Society::set_product_prices() {
     }
     Eigen::VectorXd values = get_leontief_function(A, l);
     for (std::size_t i = 0; i < dim; ++i) {
-        if (values(i) < 0.0) {
-            throw std::domain_error("Value < 0.");
+        if (values(i) <= 0.0) {
+            std::stringstream message;
+            message << "Value of item " << i << " <= 0.";
+            throw std::domain_error(message.str());
```
lennyyyyyyyy, line 161: I'm not really familiar with sstream, but I trust that this makes sense.

lennyyyyyyyy (changes requested):

Agent-Based Simulation/src/Society.cpp
Amittai-Aviram: Deleted. Thank you!

Agent-Based Simulation/src/Society.cpp
Amittai-Aviram: Good point! I have revised the code accordingly. We do not have any producers producing machines yet, but the sequence should be similar to reordering goods, except with an order of size 1.

Agent-Based Simulation/src/Society.cpp
Amittai-Aviram: Yeah. In C++, a string literal (in "quotation marks"), as you may recall from my lectures in CS2, is stored as a sequence of ASCII codes in a part of the executable file that gets loaded into read-only memory — so it cannot be concatenated at runtime with a variable value.  A stringstream is a convenient C++ abstraction for a resizable array containing a _copy_ of the string literal and any other value that you might wish to push onto it.  Basically, when you concatenate strings with + in Java or Python, they do the equivalent thing with the equivalent of stringstream behind the scenes.

Agent-Based Simulation/src/Society.cpp
```diff
@@ -68,8 +67,8 @@ void Society::set_initial_products() {
         products.push_back(new_machine);
         product_to_index[new_machine] = i;
     }
-    for (Product * product: products) {
-        product->set_inputs(products);
+    for (Product * product: goods) {
```
lennyyyyyyyy: Erm, I think this should be `for (Product * product : products)`, right? You only changed the element variable name.

Amittai-Aviram: Sorry — fixed now.

lennyyyyyyyy (approved):
LGTM. BTW, you can use `git commit --amend` followed by `git push -f` if you find that you made a small mistake that should've already been covered by your last commit.

PR #101, opened 2026-01-25
Trade function implemented for the CE.

Kkenize:
The CE section of the trade function is implemented.


Amittai-Aviram (approved):
Looks good so far.

Issue #96, opened 2026-01-20
Have the distributor create a plan for every product listed in its catalog (Amittai)


Amittai-Aviram:
In the `Distributor` constructor, have the distributor create a plan for `order_size` units of each product in its catalogue. Use a standard constant defined in `Constants.h` (say, 0.25 hours) as the cost of distribution per unit in living labor. There is only one input, the product to be distributed, and one unit of that product per unit of "output" in the plan. Machinery for now is 0.

We shall tackle the issue of re-ordering goods when the inventory goes down in a later issue.

PR #102, opened 2026-01-25
Support consumer goods whose price includes distribution costs

### LINKED ISSUE #91: Define consumer products as products sold by distributors (Amittai)
A consumer product is a product as obtained by a distributor from a producer and then sold to consumers. Hence the price of the consumer product reflects the cost of distribution _in addition to_ the cost of the product itself.

Amittai-Aviram:
- Define a new type, `ConsumerGood`, a specialization of `Product`.
  - A consumer good has a single product as its "input" and no required machinery.
  - The living labor per unit is a standard (arbitrary) quantity assumed to be the cost of distributing one unit of _any_ product to consumers.
- Keep a map from products to consumer goods as `Society::consumer_goods`. Each product serving as key is sold by distributors as the corresponding consumer good, which includes the cost of distributing the product.
- Have the distributor, in its constructor, create a plan to distribute each of the products in its initial catalog.
- When a consumer buys a product, the consumer pays the cost of the corresponding consumer good. The plan gets updated accordingly.

Resolves #91 , #96


Agent-Based Simulation/header/Constants.h
```diff
@@ -39,6 +39,7 @@ const int PRODUCT_NUM_MACHINES_MIN = 0;
 const int PRODUCT_NUM_MACHINES_MAX = 5;
 const int PRODUCT_LABOR_PER_UNIT_MIN = 1;
 const int PRODUCT_LABOR_PER_UNIT_MAX = INITIAL_WORK_WEEK; 
+const double DISTRIBUTION_LABOR_PER_UNIT = 0.2;
```
ghost, line 42: Just wondering what this is supposed to exactly represent

Agent-Based Simulation/header/Society.h
```diff
@@ -37,15 +40,16 @@ class Society : public Agent {
...
         std::vector<Person *> people;
-        std::vector<Firm *> firms;
-        std::vector<Producer *> producers;
-        std::unordered_map<Product *, std::vector<Distributor *>>
-            product_to_distributors;
         std::vector<Product *> goods;
-        std::unordered_map<Product *, std::size_t> product_to_index;
         std::vector<Machine *> machines;
         std::vector<Product *> products;
+        std::unordered_map<Product *, std::size_t> product_to_index;
```
ghost, line 46: What is product_to_index and what is its purpose?

Sal47506 (approved):
LGTM, just a couple of questions

Agent-Based Simulation/header/Society.h
Amittai-Aviram: Thanks, Sal. This is just used in the computation of initial prices (in the execution of the `Society` constructor), using the Leontief input-output matrix.  In order to build the matrix, you need to know which row and column correspond to a given product. The column is that product as output, and the values in the rows of that column are the quantities of the respective products that serve as inputs to the output product. To populate the matrix, we go through each product in order (giving us the columns — the products as outputs) and read the input map (Product * to double quantity), which gives us a quantity for each input product — but then we need the index for each of those input products to find its row and enter the quantity there as the value.

Since this data structure is only used at the beginning of time, it probably should not be part of the `Society` class definition, but, for now, it's convenient to keep it as a private member. At some point later, I will try to refactor the code to make it just a local variable that is discarded after the prices are computed.

PR #103, opened 2026-01-26
Build basic logging mechanism

### LINKED ISSUE #98: Rename draft_order to draft_plan in Producer (Devin)

### LINKED ISSUE #100: Create logging subsystem (Amittai)
We need a logging subsystem that can be switched between
- The accumulation of data to be used in graphs
- Printing informative messages to the command line.

Here are some notes in this direction:
- We define an enumeration type, with each enumerator representing a distinct key for which values are to be logged.
- Every logging call passes in a triple:
  - The appropriate enumerator
  - The value of the datum for that key
  - A CLI string.
For this mechanism to be robust and general enough for various use cases, typically, the programmer would use a `std::stringstream` to compose the CLI string so as to include the variable value.

Amittai-Aviram:
Resolves #100 
Resolves #98

### DIFF
--- a/Agent-Based Simulation/header/Firm.h
+++ b/Agent-Based Simulation/header/Firm.h
@@ -57,7 +57,7 @@ class Firm : public Agent {
   public:
 	Firm(Society * society);
     Firm(Society * society, std::unordered_set<Product *> initial_catalog);
-    void on_time_step() override;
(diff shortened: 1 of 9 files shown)


lennyyyyyyyy (approved):
LGTM.

PR #105, opened 2026-01-26
Add documentation for Person, give Persons an initial account

### LINKED ISSUE #104: Document `Person` (Devin)

### LINKED ISSUE #106: Give `Person`s an initial account for their first shopping trip (Devin)
In `Society`, after all `Product` initial prices have been computed, calculate the average cost of the initial shopping trip, and store this in an `initial_account`, with an accessor.

`Person`'s can then give themselves this `initial_account` on construction and kick-start demand.

lennyyyyyyyy:
Resolve #104
Resolve #106

### DIFF
--- a/Agent-Based Simulation/src/Person.cpp
+++ b/Agent-Based Simulation/src/Person.cpp
@@ -122,7 +122,7 @@ void Person::shop() {
         ideal_purchase_quantities[p.first] =
             p.second * PERSON_SHOPPING_PERIOD * std::abs(dist(Sim::gen));
         total_price += ideal_purchase_quantities[p.first] * 
-            Society::get_instance()->get_consumer_good(p.first)->price_per_unit;
+            society->get_consumer_good(p.first)->price_per_unit;
     }
     for (std::pair<Product *, double> p : ideal_purchase_quantities) {
         int affordable_quantity = (int) (account / total_price *
--- a/Agent-Based Simulation/src/Person.cpp
+++ b/Agent-Based Simulation/src/Person.cpp
@@ -5,6 +5,7 @@
 #include <random>
 
 #include "Constants.h"
+#include "ConsumerGood.h"
 #include "Distributor.h"
 #include "Firm.h"
 #include "Logger.h"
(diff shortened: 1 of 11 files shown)


Amittai-Aviram (commented):
The doc looks great now, from my POV. But Alex is still reviewing, and you might still be making changes to set up that initial account issue.

Agent-Based Simulation/src/Person.cpp
```diff
@@ -119,7 +121,8 @@ void Person::shop() {
     for (std::pair<Product *, double> p : purchase_frequencies) {
         ideal_purchase_quantities[p.first] =
             p.second * PERSON_SHOPPING_PERIOD * std::abs(dist(Sim::gen));
-        total_price += ideal_purchase_quantities[p.first] * p.first->price_per_unit;
+        total_price += ideal_purchase_quantities[p.first] * 
+            Society::get_instance()->get_consumer_good(p.first)->price_per_unit;
```
Amittai-Aviram: `Person` now has a pointer to his/her `Society`, so we don't need to use this accessor. (The member is slightly cheaper.)  At some point, `get_consumer_good` should really belong to `Distributor`, not `Society`, I think.

Agent-Based Simulation/src/Society.cpp
```diff
@@ -203,6 +204,22 @@ int Society::get_current_work_days_weekly() {
 	return current_work_days_weekly;
 }
 
+void Society::set_initial_account() {
+    initial_account = 0.0;
+    for (Product * product : products) {
+        ConsumerGood * consumer_good = get_consumer_good(product);
+        if (consumer_good) {
+            initial_account += consumer_good->price_per_unit*
```
Amittai-Aviram: Space before *, please.

Amittai-Aviram (changes requested):
Just a couple of small changes. Otherwise, this is GOOD!

Amittai-Aviram (commented):
Sorry!  There's one more issue here:  when a person goes to buy something from a distributor and there's no enough in the inventory, right now now, the distributor just prints a message to stderr and returns. Shouldn't the distributor, instead, _log_ the shortfall, and also put in a new order? Meanwhile, should the person look for another distributor? I'll write an issue up.

alexbcreiner0 (approved):

Amittai-Aviram (approved):
LGTM

Issue #107, opened 2026-01-26
Improve distributor's handling of a shortfall (Sal, Devin)


Amittai-Aviram:
With the new logging mechanism in place, and with Devin's latest changes (still under review — PR #105  for Issue #106 ), when you run the simulation, you can see that, if the distributor does not have enough of an item, it simply prints an error message and bails.
- The shortfall should trigger the distributor to place an order immediately (Sal).
- The person should try to get that item from another distributor (Devin).

We should try to figure out why we get into this situation in the first place. Why does the distributor ever fall short of a product?

Amittai-Aviram:
Resolved by PR #108

PR #108, opened 2026-01-27
Fix distributor shortfall ordering and producer revenue transfer

### LINKED ISSUE #72: Clean up Distributor (Sal)

### LINKED ISSUE #95: Transfer hours from buyer to seller producers upon delivery of a shipment (Sal)
Whenever a producer's plan is complete and the entity that has ordered the goods has received them, the recipient must transfer the cost of the goods in hours from the `raw_materials` account of the relevant plan to the `prd` account of the producer. The cost is the price per unit times the quantity of the order.

### LINKED ISSUE #112: Handle dropping a plan correctly (Sal)
`Producer::drop_plan` currently drops a plan by setting the value for the plan as key to NULL.  Shouldn't we remove the mapping by means of `erase`?

```
bool Producer::drop_order(Order * order) {
    log_dropped_order(order);
	if (order_to_draft_plan[order] == nullptr) {
		return false;
	}
	order_to_draft_plan[order] = nullptr;
	return true;
}
```

Also, right now, the firm that calls this function never tests the return value, so there's no point in returning a value.  So, If I am not mistaken, this should simply be as follows:

```
void Producer::drop_order(Order * order) {
    log_dropped_order(order);
    order_to_draft_plan.erase(order);
}
```

If, for some reason, we later wish to test whether there was an order (and draft plan) that is indeed dropped, we can use the return value of `erase`, which is 0 if the key does not occur in the map and 1 if it does. You can simply cast that integer to a bool.

### LINKED ISSUE #113: Do not allow a draft plan with no workers (Sal)
In `Producer::draft_plan` we currently have this:
```
    if (!draft_plan->workers.empty()) {
        draft_plan->machinery_cost = machinery_cost_per_hour *
                        (static_cast<double>(draft_plan->labor_hours) /
                         draft_plan->workers.size());
    } else {
        draft_plan->machinery_cost = 0.0;
    }
```
But why would `draft_plan->workers` ever be empty? Isn't that an error condition, and shouldn't we throw an exception to catch that condition?

Even as things stand, you can (I believe) simply multiply by the size anyway — if the vector is empty, you'll get 0:
```
    draft_plan->machinery_cost = machinery_cost_per_hour *
                    (static_cast<double>(draft_plan->labor_hours) /
                     draft_plan->workers.size());
```
I.e., we should either test for an empty `worker` vector _only_ if it's an error condition, and throw an exception in that case, or drop the test.

Sal47506:
### Summary

- Ensure distributors still place orders when an item is missing/short, instead of bailing after logging.
- Transfer finished-order revenue from `raw_materials` to `prd` in producer plans after delivery.
- Remove unused Distributor helpers (`get_output_ratio`, `planned_satisfaction_per_person`, `is_overproduced`).
- Added pooling account logic for firms
- Added code for not drafting a plan if no workers are present
- Handle dropping a plan correctly

Resolves #72
Resolves #95 
Resolves #112 
Resolves #113

Resolved Distributor's end for #107


Agent-Based Simulation/header/Firm.h
```diff
@@ -45,6 +45,19 @@ struct Order {
...
+    Order(Product * p,
+          int q,
+          Firm * c,
+          int turnaround)
+        : product(p),
+          quantity(q),
+          customer(c),
+          requested_turnaround_time(turnaround),
+          status(ORDER_REQUESTED)
+    {}
```
Amittai-Aviram: THis is good, but could you please put only the signature here and move the definition to out of line in Firm.cpp?  Please avoid single-character parameter and variable names except for ubiquitous conventions such as `i` and `j` as iteration variables. Replace `q` with `quantity` and `c` with `customer`.

Agent-Based Simulation/src/Distributor.cpp
```diff
@@ -21,12 +22,7 @@ Distributor::Distributor(
         society->add_consumer_good(product);
         int quantity =
             get_reorder_threshold(product) * FIRM_INITIAL_INVENTORY_MULTIPLIER;
-        Order * order = new Order;
-        order->product = product;
-        order->quantity = quantity;
-        order->customer = this;
-        order->requested_turnaround_time = 0;
-        order->status = Order::ORDER_REQUESTED;
+        Order * order = new Order(product, quantity, this, 0);
```
Amittai-Aviram, line 27: Yay! Excellent!

Agent-Based Simulation/src/Distributor.cpp
```diff
@@ -55,52 +51,44 @@ void Distributor::on_time_step() {
...
-
-double Distributor::planned_satisfaction_per_person(Product& product, Person& person) {
-    return get_output_ratio(product) * person.get_purchase_frequencies()[&product];
-}
-
 void Distributor::sell_goods(Product& product, int quantity, Person * person) {
     add_demand_signal(&product, quantity);
 
-    if (!inventory[&product]) {
+    auto inventory_it = inventory.find(&product);
```
Amittai-Aviram: What would this type be if it were not `auto`? Anyway, please use `count` instead of `find` in order to find out whether the entry for the product exists in the inventory. If it does not, `available` is 0. Otherwise, `available = inventory[&product]`. That is simpler.

Agent-Based Simulation/src/Distributor.cpp
```diff
@@ -55,52 +51,44 @@ void Distributor::on_time_step() {
...
-        return;
     }
-    int available = inventory[&product];
+    int remainder = 0;
+    int sell_quantity = std::min(available, quantity);
     if (available < quantity) {
-        std::cerr << "Shortfall in product " << product.product_name 
-            << " of " << quantity - available << " units. " << std::endl;
+        remainder = quantity - available;
+        std::cout << "Shortfall in product " << product.product_name
```
Amittai-Aviram: Use the Logger or else do not record at all. Do not print to `cout` unless you are doing so temporarily to help identify a bug.

Agent-Based Simulation/src/Firm.cpp
```diff
@@ -117,9 +117,12 @@ void Firm::reorder_product_to_threshold(
                 threshold - static_cast<double>(pending_inventory)
             ) / product->order_size));
         Logger::log_firm_reorder(product->product_name, discrepancy);
-        Order * order = new Order{product, discrepancy, this,
-            (int) (pending_inventory / threshold * FIRM_STOCKPILE_DURATION),
-            Order::ORDER_REQUESTED};
+        Order * order = new Order(
+                product,
+                discrepancy,
```
lennyyyyyyyy: Can you change this so that it orders the complete threshold size? I think this is also causing Distributor to run out of goods prematurely.

Agent-Based Simulation/src/Firm.cpp
```diff
@@ -117,9 +117,12 @@ void Firm::reorder_product_to_threshold(
...
+                this,
+                (int) (pending_inventory / threshold * FIRM_STOCKPILE_DURATION)
```
lennyyyyyyyy: Also, add a constant < 1.0, called something like DEADLINE_SAFETY_MULTIPLIER, that causes the goods to arrive a little safer, a bit before we actually run out.

lennyyyyyyyy (changes requested):

Agent-Based Simulation/src/Firm.cpp
```diff
@@ -253,22 +257,38 @@ void Firm::assign_plan_dependent_fields(
...
                     worker->get_current_productivity());
         }
-        draft_plan->predicted_turnaround_time =
-            draft_plan->training_time +
-            predict_turnaround_time(draft_plan->order, total_suitability);
-        draft_plan->labor_hours =
-            draft_plan->labor_hours_remaining =
-            draft_plan->training_time * draft_plan->workers.size() +
-            predict_labor_hours(draft_plan->order, total_suitability);    
+        if (total_suitability <= 0) {
```
lennyyyyyyyy, line 278: Can you revert this error trapping?

Agent-Based Simulation/docs/Distributor.MD
```diff
@@ -0,0 +1,23 @@
...
+
+## Selling goods
+`sell_goods(Product& product, int quantity, Person * person)`:
+- Adds a demand signal for the requested quantity.
+- If inventory is short, creates a new order for the remainder and sends it to producers.
+- Charges the buyer for the portion actually sold and records revenue on the product's plan.
+- Reduces inventory by the quantity sold.
+
+## Reordering
+The Distributor reorders only for products in its catalog.
```
Amittai-Aviram: Good so far. We need details on the reorder threshold and how it is changed in response to demand. Ho also need to mention that a plan is just like any other firm's plan, but, here, the "raw material" is the good to be distributed.  Does a distributor always have an initial workforce at construction? (I don't think so — it gets one when it creates a plan — but how?)

Agent-Based Simulation/src/Distributor.cpp
```diff
@@ -55,52 +51,44 @@ void Distributor::on_time_step() {
...
-        std::cerr << "Shortfall in product " << product.product_name 
-            << " of " << quantity - available << " units. " << std::endl;
+        remainder = quantity - available;
+        std::cout << "Shortfall in product " << product.product_name 
+            << " of " << remainder << " units. " << std::endl;
+        
+        int turnaround_time = product.living_labor_per_unit * remainder;
+        Order * order = new Order(&product, remainder, this, turnaround_time);
+        send_order(order);
+    }
```
Amittai-Aviram, line 81: OK, we talked about this. The upshot of our discussion is that the distributor does not re-order the good right away in this scenario, since it will re-order it anyway because the inventory has fallen below the threshold. Meanwhile, the person will try to buy the rest of the good from some other distributor.

Agent-Based Simulation/src/Distributor.cpp
```diff
@@ -55,52 +51,44 @@ void Distributor::on_time_step() {
...
     ConsumerGood * consumer_good = society->get_consumer_good(&product);
     if (!consumer_good) {
         std::cerr << "No consumer good for product " << product.product_name << std::endl;
         return;
     }
-    double cost = quantity * consumer_good->price_per_unit;
+    double cost = sell_quantity * consumer_good->price_per_unit;
     if (!person->charge(cost)) {
-        std::cerr << "Person cannot afford " << quantity 
+        std::cerr << "Person cannot afford " << sell_quantity
```
Amittai-Aviram, line 92: This and the previous write to `std::cerr` are good for now, since these are probably bug conditions — a person should not try to buy more goods than that person can afford.

Agent-Based Simulation/src/Distributor.cpp
```diff
@@ -55,52 +51,44 @@ void Distributor::on_time_step() {
...
             << " units of " << product.product_name << " costing " 
             << cost << std::endl;
         return;
     } 
     Plan * plan = product_to_plan[&product];
-    plan->outgoing_units_consumed += quantity;
+    plan->outgoing_units_consumed += sell_quantity;
     product_to_plan[&product]->prd += cost;
```
Amittai-Aviram: You've already looked the plan up, so use `plan`.

Agent-Based Simulation/src/Firm.cpp
```diff
@@ -211,6 +214,7 @@ void Firm::assign_workers_by_suitability_threshold(
             draft_plan->workers.push_back(unemployed_person);
         }
     }
+
```
Amittai-Aviram: Please remove this vertical space.

Agent-Based Simulation/src/Firm.cpp
```diff
@@ -253,22 +257,38 @@ void Firm::assign_plan_dependent_fields(
...
         }
-        draft_plan->predicted_turnaround_time =
-            draft_plan->training_time +
-            predict_turnaround_time(draft_plan->order, total_suitability);
-        draft_plan->labor_hours =
-            draft_plan->labor_hours_remaining =
-            draft_plan->training_time * draft_plan->workers.size() +
-            predict_labor_hours(draft_plan->order, total_suitability);    
+        if (total_suitability <= 0) {
+            total_suitability = 1.0;
```
Amittai-Aviram, line 279: I'm confused about this. Maybe this is the same issue as what Devin meant. Here (l. 260), you check for a non-positive `total_suitability` and set it to 1 in that case. Then, later, you handle the (no longer possible) case in the `else` clause of where `total_suitability > 0` is not true.  So you've already eliminated the possibility handled by that `else` clause.

Amittai-Aviram (changes requested):
Good work so far, but, as you can see, I have questions and issues. Also, I have _not_ gotten through Producer or Society yet, so comments on those will come separately. I just did not want to hold this up any longer.

Agent-Based Simulation/src/Producer.cpp
```diff
@@ -11,56 +10,44 @@
...
         std::unordered_set<Product *> initial_catalog
         ) :
-    Firm(society, initial_catalog)
-{
-    std::unordered_set<Machine *> initial_machines;
+    Firm(society, initial_catalog) {
     for (Product * product : initial_catalog) {
-        for (Machine * machine : product->machines_needed) {
-            initial_machines.insert(machine);
+        for (auto& p : product->inputs_per_unit) {
```
Amittai-Aviram: Please avoid `auto` as much as possible. Here, please replace `auto` with `std::pair<Product *, double>`.

Agent-Based Simulation/src/Producer.cpp
```diff
@@ -11,56 +10,44 @@
...
+            inventory[p.first] += p.second *
```
Amittai-Aviram: Why are you scaling the initial stockpile of the raw material (or input product) by the proportion needed per unit of output product? Why not just fill the inventory up to `order_size * INVENTORY_MULTIPLIER`? I do not understand this.

Agent-Based Simulation/src/Producer.cpp
```diff
@@ -11,56 +10,44 @@
 #include "Sim.h"
 #include "Society.h"
 
-Producer::Producer(Society * society) : Firm{society} {}
+Producer::Producer(Society * society) : Firm(society) {}
```
Amittai-Aviram: Do we actually need a default constructor? If not, please eliminate it.

Agent-Based Simulation/src/Producer.cpp
```diff
@@ -11,56 +10,44 @@
...
         Society * society,
         std::unordered_set<Product *> initial_catalog
         ) :
-    Firm(society, initial_catalog)
-{
-    std::unordered_set<Machine *> initial_machines;
+    Firm(society, initial_catalog) {
     for (Product * product : initial_catalog) {
-        for (Machine * machine : product->machines_needed) {
-            initial_machines.insert(machine);
```
Amittai-Aviram, line 23: Why have you gotten rid of this logic? The idea is that, if a producer has a product listed in its catalogue, then it should start its existence in the simulation already possessing all the machines needed to produce that product.  This is as opposed to the producer acquiring the machines upon receiving its first order for the product. We can do it that way, but have you implemented that logic?  That requires ordering a single machine from a producer of that machine. Right now, producers do not list machines in their catalogues.

Agent-Based Simulation/src/Producer.cpp
```diff
@@ -11,56 +10,44 @@
...
+                                  FIRM_INITIAL_INVENTORY_MULTIPLIER;
         }
     }
-    for (Machine * machine : initial_machines) {
-        machines.push_back(machine);
-    }
-    for (Product * product : get_products_to_reorder()) {
-        inventory[product] =
-            get_reorder_threshold(product) * FIRM_INITIAL_INVENTORY_MULTIPLIER;
-    }
```
Amittai-Aviram, line 32: See above. I do not understand why you have gotten rid of this.

Agent-Based Simulation/src/Producer.cpp
```diff
@@ -11,56 +10,44 @@
...
 }
 
 void Producer::on_time_step() {
     Firm::on_time_step();
 	execute_plans();
```
Amittai-Aviram: Please rename the function `execute_plans` to `move_plans_forward_one_step`. Likewise, please rename `Producer::execute_plan` to `move_plan_forward_one_step`.

Agent-Based Simulation/src/Producer.cpp
```diff
@@ -11,56 +10,44 @@
...
-        Logger::log_producer_plans(plans_in_progress);
-    }
 }
 
 bool Producer::can_produce(Product * product) {
     return catalog.count(product);
 }
 
 int Producer::draft_plan(Order * order) {
-    Logger::log_producer_draft_plan(order->product->product_name, order->quantity);
```
Amittai-Aviram: You do not want to log this event?

Agent-Based Simulation/src/Producer.cpp
```diff
@@ -11,56 +10,44 @@
...
-    bool enough_inputs = true;
+	bool enough_inputs = true;	
     for (auto &p : order->product->inputs_per_unit) {
```
Amittai-Aviram: s/`auto &p`/`std::pair<Product *, double>& input`/

Agent-Based Simulation/src/Producer.cpp
```diff
@@ -11,56 +10,44 @@
...
+			enough_inputs = false;
+			return DRAFT_ORDER_REJECTED;
         }
+        add_demand_signal(p.first, p.second * order->quantity);
     }
-	if (!enough_inputs || order_to_draft_plan[order] != nullptr) {
-		return DRAFT_ORDER_REJECTED;
-	}
 	if (!can_produce(order->product)) {
 		catalog.insert(order->product);
```
Amittai-Aviram: What entitles this producer to add this product to its catalogue? Has it acquired the machinery and necessary inputs?

Agent-Based Simulation/src/Producer.cpp
```diff
@@ -78,11 +65,11 @@ int Producer::draft_plan(Order * order) {
     }
 
 	order_to_draft_plan[order] = draft_plan;
+	//std::cout << "Draft plan predicted turnaround time: " << draft_plan->predicted_turnaround_time << std::endl;
```
Amittai-Aviram: Please remove.

Agent-Based Simulation/src/Producer.cpp
```diff
@@ -78,11 +65,11 @@ int Producer::draft_plan(Order * order) {
...
 	return draft_plan->predicted_turnaround_time;
 }
 
 bool Producer::drop_order(Order * order) {
-    Logger::log_producer_dropped_order(order->product->product_name, order->quantity);
```
Amittai-Aviram: Why do you not want to log this event? And, if you decide that an event is not worth logging, please remove the corresponding function and vector from the Logger.

Agent-Based Simulation/src/Producer.cpp
```diff
@@ -91,9 +78,6 @@ bool Producer::drop_order(Order * order) {
 }
 
 bool Producer::pursue_order(Order * order) {
-    for (std::pair<Product *, double> p : order->product->inputs_per_unit) {
-        add_demand_signal(p.first, p.second * order->quantity);
-    }
 	if (order_to_draft_plan[order] == nullptr) {
```
Amittai-Aviram: You're checking whether this producer has already created a draft plan for this order, correct? So that means checking whether the order is maped to a draft plan — correct? If so, then use `count`:
`if (!order_to_draft_plan.count(order)) {`.

Agent-Based Simulation/src/Producer.cpp
```diff
@@ -105,18 +89,17 @@ bool Producer::pursue_order(Order * order) {
 			workers.erase(it);
 		}
 		it = std::find(
```
Amittai-Aviram, line 97: Instead of having to call `Society::get_instance` over and over, just set a local variable: `Society * society = Society::get_instance();`.

Agent-Based Simulation/src/Producer.cpp
```diff
@@ -105,18 +89,17 @@ bool Producer::pursue_order(Order * order) {
...
-		if (it != society->get_unemployed_people().end()) {
-			society->get_unemployed_people().erase(it);
+		if (it != Society::get_instance()->get_unemployed_people().end()) {
+			Society::get_instance()->get_unemployed_people().erase(it);
 		}
 	}
 	// move draft_plan to plans_in_progress
 	order_to_draft_plan[order] = nullptr;
 	plans_in_progress.push_back(draft_plan);
-    Logger::log_producer_pursued_plan(draft_plan);
```
Amittai-Aviram: See earlier comments on the Logger. Why do you not want to log this event?

Agent-Based Simulation/src/Producer.cpp
```diff
@@ -161,8 +144,10 @@ void Producer::end_plan(Plan * plan) {
 	inventory[plan->order->product] += plan->order->quantity;
 	// simplification: product shipped instantly
 	inventory[plan->order->product] -= plan->order->quantity;
-	plan->order->customer->receive_shipment(plan->order);
-    plan->prd += plan->order->product->price_per_unit * plan->order->quantity;
+    plan->order->customer->receive_shipment(plan->order);
+    double revenue = plan->order->product->price_per_unit * plan->order->quantity;
```
Amittai-Aviram: I don't want us to model the transaction this way. We need a `transfer` function from `Agent` to `Agent` to transfer a quantity of labor hour credits. The customer should call this function to transfer labor hour credits to the producer.  In this case, the customer should transfer `product->price_per_unit * order->quantity` credits to the producer. Or else have a `charge` function — the producer charges the credits from the customer, the way we do now with consumers. Then the producer ships the order to the customer. The transfer should make it clear who is deducting the credits from which account and who is reeiving the credits to which account. This is critically important.

Agent-Based Simulation/src/Producer.cpp
```diff
@@ -161,8 +144,10 @@ void Producer::end_plan(Plan * plan) {
...
+    plan->raw_materials -= revenue;
```
Amittai-Aviram: This makes no sense.

Agent-Based Simulation/src/Society.cpp
```diff
@@ -24,15 +24,14 @@ Society * Society::get_instance() {
 
 Society::Society() {
     set_initial_products();
+    std::unordered_set<Product *> all_products(products.begin(), products.end());
     for (int i = 0; i < STARTING_NUM_PRODUCERS; i++) {
-        Producer * producer = new Producer(this, {goods[i %
-                STARTING_NUM_PRODUCTS]});
+        Producer * producer = new Producer(this, all_products);
```
Amittai-Aviram: I don't understand this. This makes every producer produce _all products_. Why?

Agent-Based Simulation/src/Society.cpp
```diff
@@ -24,15 +24,14 @@ Society * Society::get_instance() {
...
         producers.push_back(producer);
         firms.push_back(producer);
     }
     for (int i = 0; i < STARTING_NUM_DISTRIBUTORS; i++) {
-        Distributor * distributor =
-            new Distributor(this, {goods[i % STARTING_NUM_PRODUCTS]});
+        Distributor * distributor = new Distributor(this, all_products);
```
Amittai-Aviram: Likewise — does every distributor distribute _all products_? BTW, that also means that every distributor distributes _all machines_ in addition to _all goods_. Why?

Agent-Based Simulation/src/Society.cpp
```diff
@@ -42,6 +41,9 @@ Society::Society() {
                 firm->add_supplier(producer);
             }
         }
+        if (!machines.empty()) {
+            firm->machines.push_back(machines[std::rand() % machines.size()]);
```
Amittai-Aviram: So I originally had a logic where every producer would come into existence (through its constructor) having all machines needed to produce all of the products in its catalogue. At that point, each producer's catalogue started out with only one product, and we left until later the question of how a producer might grow to start producing other products. We could have had each producer start out with a few products — but producing _all_ products seems to me inappropriate. Anyway, now, you are supplying each producer (firm) with a single _randomly-selected_ machine. Why have you made this change?

Amittai-Aviram (changes requested):
As you can see, I have a lot of questions about some of the changes you have made. In several cases, I do not understand what you have in mind.

Agent-Based Simulation/src/Producer.cpp
ghost: Removed

Agent-Based Simulation/src/Society.cpp
ghost: Removed

Agent-Based Simulation/header/Firm.h
ghost: Moved constructor defs to Firm.cpp and used descriptive parameter names.

Agent-Based Simulation/src/Distributor.cpp
ghost: no changes made here

Agent-Based Simulation/src/Distributor.cpp
ghost: Changed to:

`bool has_inventory = inventory.count(&product) > 0;
 int available = has_inventory ? inventory[&product] : 0;`

Agent-Based Simulation/src/Distributor.cpp
ghost: Removed cout and used logger instead

Agent-Based Simulation/docs/Distributor.MD
ghost: Still needs more work (we should discuss this more on the next meeting)

Agent-Based Simulation/src/Distributor.cpp
ghost: Removed the immediate reorder on shortfall and rely on the normal threshold-based reorder instead

Agent-Based Simulation/src/Distributor.cpp
ghost: Kept cerr logs

Agent-Based Simulation/src/Distributor.cpp
ghost: Used the cached plan variable

Agent-Based Simulation/src/Firm.cpp
ghost: Space removed

Agent-Based Simulation/src/Firm.cpp
ghost: Remove the dead `else` branch after clamping `total_suitability`

Agent-Based Simulation/src/Producer.cpp
ghost: Replaced `auto` with `std::pair<Product * const, double>&`

Agent-Based Simulation/src/Producer.cpp
ghost: Changed the initial input inventory to use the standard reorder/threshold approach (`order_size * multiplier`)

Agent-Based Simulation/src/Producer.cpp
ghost: Removed constructor

Agent-Based Simulation/src/Producer.cpp
ghost: Restored logic

Agent-Based Simulation/src/Producer.cpp
ghost: Restored

Agent-Based Simulation/src/Producer.cpp
ghost: Changed it so the customer pays the producer before shipment with clear account updates.

Agent-Based Simulation/src/Society.cpp
ghost: Reverted so each distributor starts with a limited catalog (goods only).

Agent-Based Simulation/src/Society.cpp
ghost: Removed the random machine assignment and restored the catalog-based machine initialization.

Agent-Based Simulation/src/Producer.cpp
ghost: Renamed `execute_plans` to `move_plans_forward_one_step`

Agent-Based Simulation/src/Producer.cpp
ghost: Removed catalogue insertion on draft

Agent-Based Simulation/src/Producer.cpp
ghost: `pursue_order` now uses `count`

Agent-Based Simulation/src/Producer.cpp
ghost: Set a local variable `Society * society = Society::get_instance();` instead of calling `Society::get_instance` over and over again

Agent-Based Simulation/src/Distributor.cpp
Amittai-Aviram: I'm afraid this has fallen behind WRT Logger developments, but it should work. For consistency, we need to wrap the call to `Logger::get_instance` into a private function belonging to `Distributor`. You do not need to do that for this PR, though.

Agent-Based Simulation/header/Firm.h
```diff
@@ -44,6 +44,15 @@ struct Order {
     Firm * customer;
     int requested_turnaround_time;
     OrderStatus status;
+    
+    Order(
+            Product * product,
+            int quantity,
+            Firm * customer,
+            int requested_turnaround_time
+    );
```
Amittai-Aviram, line 53: Please move this constructor declaration to second position (right after the enumeration type definition), and please remove vertical spaces.

Agent-Based Simulation/src/Firm.cpp
Amittai-Aviram: I still want to know whether `total_suitabiliy` could ever be <= 0, but we'll leave that for another day.

Agent-Based Simulation/docs/Distributor.MD
```diff
@@ -0,0 +1,34 @@
...
+is the max of `order_size` and demand-based stockpiling. Demand is tracked via
+`add_demand_signal()` and decays in `apply_demand_window()` over
+`FIRM_DEMAND_WINDOW`, resulting in `inventory_demands[product] *
+FIRM_STOCKPILE_DURATION` as the demand component.
+
+## Plans and workforce
+Each catalog product has a distribution `Plan` created at initialization to
+track labor hours, outgoing units consumed, and revenue; the "raw material" for
+distribution is the product itself. Plans start with zero workers and are
+staffed via the shared firm worker-assignment logic.
```
Amittai-Aviram, line 34: I actually did not review this documentation carefully. I will do that in a separate pass, after you have merged this PR into main.

Amittai-Aviram (approved):
I have requested one small change (in `Distributor`).  Otherwise, LGTM.
Good work! This is ready to be merged with main — but please read my new comments. Except for the one change request, the others are responses to your replies and mention things we need to change in the future.

Amittai-Aviram (approved):

Amittai-Aviram (approved):

PR #110, opened 2026-01-30
Completed Documentation and Updated Shortage/Surplus Detection

Kkenize:
Completed Documentation for the event-based simulation and the trade interface while updated shortage/surplus detection to use supply instead of output based on my discussion with Alex on Wednesday. The documentation is on the README.md file in the trade interface folder.


Trade Interface/README.md
```diff
@@ -0,0 +1,46 @@
+# Simulation of Trade Between Capitalist and Labor-Time Economies
+
+## The Event-Based Simulation: A Dynamic Capitalist Economy Backend for the Trade Simulation
+### Description / Overview
+
+The Event-Based Simulation implements a dynamic capitalist economy model designed to serve as the price- and supply-generating backend for this trade interface. Its purpose is to generate internally consistent trajectories of prices, outputs, supplies, labor usage, and value magnitudes under capitalist production relations, which can then be used by this trading system. It is built as a production-grounded and class-structured model.
```
Amittai-Aviram: implements a dynamic model of a capitalist economy
"this trade interface" -> "an interface for trade between this capitalist economy and another economy based on labor-time accounting."
"labor usage" -> "labor usage rages"

Trade Interface/README.md
```diff
@@ -0,0 +1,46 @@
...
+
+### Simulation Design
+
+- Evaluates the commodity states of both economies at each time step to identify missing products and unmet needs
+- Constructs import demand lists for each economy based on detected shortages
+- Constructs export supply lists for the capitalist economy based on detected surpluses
+- Removes import and export entries automatically when shortages or surpluses disappear
+- Restricts the labor-time economy to trading only when it is missing required goods
+- Treats overproduction in the labor-time economy as conditional on trade necessity
```
Amittai-Aviram: Maybe you should explain this point:
- Assumes that the labor-time economy would only overproduce (with respect to its own needs) _in order_ to trade exports with imports that it cannot currently produce and import from the CE, such as scarce raw materials.

Amittai-Aviram (changes requested):
I only focused on the documentation, which is pretty good, but on which I am requesting a few small changes. I am leaving your code changes for Alex to review.

Trade Interface/README.md
Kkenize: Replaced with your suggested words. (I will discuss it with you today.)

Trade Interface/README.md
Kkenize: Replaced with your suggested words. (I will discuss it with you today.)

Amittai-Aviram (approved):

PR #114, opened 2026-02-01
Support logging to CSV files for graphs

### LINKED ISSUE #111: Improve Logging (Amittai)
The current logging mechanism is awkward because it requires the definition of a new, context-specific function _and_ container for every client call site. Instead, we need to generalize the logging API to make it easily extensible. In addition, right now, the mechanism anticipates transfer of data into Python code by means of a third-party library such as Pybind11, but a more versatile implementation would instead allow the option to write the data to CSV files. Finally, the logging mode (print to CLI or store in containers) requires recompilation with or without a preprocessor definition flag.  It would be more convenient to use a CLI argument to the `sim` command, as Joshua did in his project in summer 2025.
- Define an enumeration type whose enumerators correspond to the possible logging client classes, which are all the entity classes (not only agents) in the ABSM.
- Create a new API that takes the client enumerator as the first argument and a text string as the second. This should provide overloads for various sequences of data — e.g., nothing (for an event), a string product name and an integer quantity, a string product name and a double price, or a single integer hours.
- Replace the individual vectors with a map hierarchy: client -> ID string -> time step -> tuple, where _tuple_ has type `std::variant`, polymorphically representing any of the possible tuple types needed by the various call sites.
- Define command-line arguments to determine the behavior of the logger. These should set a variable in the `Sim` object.
- For the -graph option, use a `std::visitor` to iterate over the values of each tuple and render the values appropriately.

Amittai-Aviram:
- Replaced the previous need-specific logging API with a more extensible one that expects a client (an enumerator in an enumeration of possible logging clients — basically, the other classes in the simulation model), a label (the subject matter of the log), an id number uniquely identifying the instance of the client class logging its data, the current time stamp, and the case-specific data (such as a product name and quantity for an order, a plan, etc.).
- Implemented the API with advanced C++ 17 features so as to accommodate more or less arbitrary data sets.
- Defined two command-line options:
  - `--trace` — print data to the command line as well as storing the data internally.
  - `--write` — write data to CSV files in the new `data` subdirectory.

The CSV files are named by the client name concatenated (by `_`) with the label. Each entry in each CSV file has the general format
```
id, time_step, values ...
```
Resolves #111


Amittai-Aviram:
@lennyyyyyyyy — I am asking you to review this because I thought you might be interested in the advanced C++ features. These allow type-safe ways of handling tuples of various types and lengths polymorphically.

Agent-Based Simulation/header/Logger.h
```diff
@@ -1,42 +1,91 @@
...
+        template<typename TupleT>
+            static void trace_tuple(const TupleT& values);
+        template<typename TupleT>
+            static void write_tuple(std::ofstream& file, const TupleT& values);
+        static const char * clients[];
+        static const char * logging_dir;
+        std::unordered_map<Client,
+            std::unordered_map<std::string,
+            std::map<unsigned int,
+            std::map<int, Tuple>>>> data;
```
lennyyyyyyyy, line 90: Is it possible to make this a `std::map<std::tuple<Client, std::string, unsigned int, int>, Tuple>` instead? This could be expensive.

Or make a struct, and do `std::map<LogInfo, Tuple>`.

lennyyyyyyyy: Ok, I guess it pays off when you open and close the file streams in `write_data()`. I guess you could group it like 
`std::unordered_map<ClientAndLabel, std::map<IdAndTimestep, Tuple>>` 
But what you have is probably fine too.

lennyyyyyyyy (approved):

Agent-Based Simulation/header/Logger.h
Amittai-Aviram: Possibly a good idea. I'll think about it. Thank you!

PR #126, opened 2026-02-14
`PriceController`: compute new price based on updated average labor and current input costs

### LINKED ISSUE #122: `PriceController`: set the new price per unit of a product to include the (indirect) costs of inputs (Amittai)
Currently, a product's price per unit is set initially in the `Society` constructor, and then adjusted on the completion of any plan to produce that product, by the Price Controller. However, the Price Controller only adjusts the price based on reported actual living labor, not taking into account the possibly revised costs of the upstream inputs. This needs to be corrected: the Price Controller should compute the new price per unit of the product based on all three kinds of input: machinery, "raw materials" (upstream products), and living labor.

### LINKED ISSUE #123: `PriceController`: recompute average living labor over all units within a time window (Amittai)
Currently, we recompute the average living labor based on a window of _orders_, without weighting the average by the size of the order. This leads to an inaccurate under-weighting of large orders and over-weighting of small ones.  The average should be over all _units_ (regardless of order size) produced within a time window. This requires maintenance of a history of orders (a queue) with time stamps, so that the time of the oldest order can be checked and discarded if it has fallen outside of the window.

Amittai-Aviram:
This PR reconceives the computation of an updated product price:
- `PriceController` is now a singleton class, in order to support instance data (the plan history — see below).
- As before, a firm calls (now indirectly, through the singleton accessor) `PriceController::update_price` when it completes a plan for an order of goods. 
- Plans are saved in a queue representing a fixed-size time window of history, including the current time step.
- If the oldest plan in the history window is older than the set window time period, it is removed.
- The average work hours per unit of the output product and the average hours worked by a team per unit of output product are computed. The latter is the number of hours per unit that each machine is assumed to be in use, on the assumptions that
  - A single machine is used by only one worker at a time, i.e., that the machine is used the same number of hours as the whole team works together; and
  - Each machine is in continuous use so long as the workers are working.
- The updated price is the sum of the newly-computed average work per output unit, the costs of the input goods (materials) per output unit (based on their respective prices, proportioned to the quantity per unit), and the sum of the machine depreciation costs.

Resolves #122 
Resolves #123


Amittai-Aviram:
@alexbcreiner0 — Please feel free to limit your review to PriceController.h and PriceController.cpp . Most of the other changes included into this PR are related to logging to a SQLite database, which was work that I had started before our conversation about the price controller. I got the first step of the database stuff working and then switched over to working on the price controller, since the latter takes higher priority. Thanks!

PR #129, opened 2026-02-14
Export Capacity Ratio For Trade Measurements 

Kkenize:
Records equilibrium total demand from the event based simulation for the export capacity calculation of CE products. 
The equilibrium total demand is the reproduction requirement, as it measures how much the system can safely give up without harming reproduction. 
The equation for calculating the export capacity ratio for a given product type i is to divide its supply by its equilibrium total demand. 
We then subtract 1 from the ratio. If the ratio is positive, then the CE is open to trade since it has surplus. If it's negative, then the CE will not trade this product type (it would instead want to import). 
We can time product type i's (export capacity ratio - 1) by the order size of product type i of the LTE to get the specific amount the CE can export.


Kkenize:
> Records equilibrium total demand from the event based simulation for the export capacity calculation of CE products. The equilibrium total demand is the reproduction requirement, as it measures how much the system can safely give up without harming reproduction. The equation for calculating the export capacity ratio for a given product type i is to divide its supply by its equilibrium total demand. We then time product type i's export capacity ratio by the order size of product type i of the LTE to get the specific amount the CE can export.

More changes to come, this is not the final version yet of this pull request.

Event-Based Simulation/simulation.py
```diff
@@ -216,13 +216,62 @@ def export_trajectories_with_equilibrium(
...
+    # Extract equilibrium state vectors
+    eq_prices = np.array(traj["p"])[eq_index]   # equilibrium prices
+    eq_output = np.array(traj["q"])[eq_index]   # equilibrium output
+
+    # Worker consumption at equilibrium
+    # b = (alpha_w * m_w) / (p · b_bar) * b_bar
+    # where: alpha_w = worker propensity to consume,
+    #        m_w = worker share of total money,
+    #        p = equilibrium price vector,
+    #        b_bar = baseline worker consumption bundle
```
Amittai-Aviram: I have no earthly idea what these symbols mean. Alex's code depended on a separate presentation of the math, where the names of variables corresponded one-to-one. That is not the situation here, unless you somehow copy the definitions of these variables. Computing variable names are generally informative, unlike math variable names, for this reason.

Amittai-Aviram (commented):
Looks OK so far. One comment about this code in general is that it seems to have what we call the _business logic_ mixed up with the _presentation logic_. The _business logic_ is the actual computation of your model; the _presentation logic_ is the creation of your graphs. It's better to separate these two kinds of code as much as possible. Note how, in the ABSM, logging is handled on its own, and the business logic just calls into the logger with one function call at a time here and there. This way, the presentation logic has a minimal appearance within the business logic.

Event-Based Simulation/simulation.py
Kkenize: Explained in our meeting last week.

Kkenize:
Finished simulation and Graphing. I would like to present the graphs tomorrow to notify the group of my progress if possible.

Amittai-Aviram (approved):

Issue #130, opened 2026-02-15
Add `input_inventory` for Producer


ghost:
Producers should have an input and output inventory while Distributors should only have an output inventory. Currently, there is only inventory (output inventory) for the `Firm` class

PR #132, opened 2026-02-17
Reorder Raw Materials Logic Added for Producers and Fixed Transfer of PRD and Pooled Account

### LINKED ISSUE #125: Create a Reorder Method for Raw Materials (`Distributor` Class)
- Reordering raw materials will deplete the `Distributor`'s `pooled account` by the quantity of raw materials it reorders in labor hours
- Must reorder from another `Distributor`

Sal47506:
Previously the pooled account had a logic issue where pooled account did not get transferred to PRD, the reorder of raw materials came out of "thin air", and the pooled account would decrease for every time a unit of raw material was used.

Now, the updated logic to fix these logical issues feature:
- An `unordered_map` for the input inventory was made called   `std::unordered_map<Product *, int> input_inventory`
- When a plan is being started/executed (in `void Producer::start_plan(Plan * plan)`) then `pooled_account` is initialized to `plan->raw_materials` and `plan->raw_materials = 0` 
- Pooled account value gets deducted by the amount needed to restock an input inventory (when it is under the `RAW_MATERIAL_THRESHOLD`) every time a reorder of raw materials occur and that value is then added to PRD
- If no upstream `Producer` meets the requirement of having enough inputs to fulfill the reorder of raw materials to the downstream `Producer` then a random `Producer` is chosen from the society's list of Producers to fulfill this reorder by calling the helper method `reorder_product_to_threshold`

Resolves #125, #130


Agent-Based Simulation/header/Constants.h
```diff
@@ -53,6 +53,8 @@ const int FIRM_DEMAND_WINDOW = WEEK;
 const double FIRM_INITIAL_INVENTORY_MULTIPLIER = 2;
 const double DEADLINE_SAFETY_MULTIPLIER = 0.75;
 
+const int RAW_MATERIAL_THRESHOLD = 50;
```
Amittai-Aviram, line 56: Is this in units? And does it apply to each stockpile of an input product?

Agent-Based Simulation/header/Logger.h
```diff
@@ -15,6 +13,7 @@ using TupleStringDouble = std::tuple<std::string, double>;
 using TupleIntDoubleInt = std::tuple<int, double, int>;
 using TupleStringStringInt = std::tuple<std::string, std::string, int>;
 using Tuple = std::variant<TupleNone,
+      TupleInt,
```
Amittai-Aviram, line 18: You have to define the symbol `TupleInt` first, before you include it as one of the types in the `variant`. I do not see any definition above. If you mean to allow a "tuple" consisting of just one integer of type `int`, the previous definition would be
`using TupleInt = std::tuple<int>;`

Agent-Based Simulation/header/Producer.h
```diff
@@ -23,9 +23,18 @@ class Producer : public Firm {
 	int draft_plan(Order * order);
 	void drop_order(Order * order);
 	bool pursue_order(Order * order);
+  void reorder_raw_materials(Plan * plan, Product * product);
+  bool has_raw_materials();
+  void initialize_input_inventory(std::unordered_map<Product *, int>& inventory_items); 
+  void get_pending_input_inventory(Product * product);
+  void log_pending_input_inventory_level(Product * product);
```
Amittai-Aviram: Please list all the log-related functions together — low down among the declarations. Let's try to keep the declarations reasonably ordered, going down from most to least important.

Agent-Based Simulation/header/Producer.h
```diff
@@ -23,9 +23,18 @@ class Producer : public Firm {
...
+  int get_pooled_account();
 
   private:
+  std::unordered_map<Product *, int> input_inventory;
 	std::unordered_map<Order *, Plan *> order_to_draft_plan;
+  int pooled_account;
```
Amittai-Aviram: I hate to pile the work on, but could we please change this name to a more informative and specific one? Maybe `raw_materials_account`. You'd change the name of the accessor as well. But, at some point, I'd also like to change "raw materials" everywhere to "input products," which is more accurate, so this would be `input_products_account`, and the accessor would be `get_input_products_account`.

Agent-Based Simulation/header/Firm.h
```diff
@@ -77,7 +77,6 @@ class Firm : public Agent {
 
 
   protected:
-    double pooled_account;
```
Amittai-Aviram, line 80: Why are you removing this here? Oh! Were you thinking at this point that only producers would have this account? And then we changed that coming out of our conversation yesterday. It makes sense that even distributors pool all their input products accounts together from the various ongoing plans to distribute goods.  (BTW, we should rename this as `input_products_account` — see my comment on `Producer.h`.)

Agent-Based Simulation/src/Firm.cpp
```diff
@@ -58,10 +59,11 @@ double Firm::get_avg_productivity() {
 }
 
 int Firm::get_inventory(Product * product) {
-    if (!inventory.count(product)) {
+    auto it = inventory.find(product);
```
Amittai-Aviram: Please _no auto!_. Let's avoid using `auto` everywhere unless the type is _insanely_ complicated and takes up a lot of space to write out. Also, please do not check for presence in a map this way. Please just use `count`:
```
if (!inventory.count(product)) {
    return 0;
}
```

Agent-Based Simulation/src/Firm.cpp
```diff
@@ -58,10 +59,11 @@ double Firm::get_avg_productivity() {
...
+    if (it == inventory.end()) {
         return 0;
     }
-    return inventory[product];
+    return it->second;
```
Amittai-Aviram: `return inventory[product];`

Agent-Based Simulation/src/Firm.cpp
Amittai-Aviram, line 120: It's not clear to me whether this function will pertain to the input or the output inventory. It refers to "outbound orders," which means orders from purchasers (downstream) to buy a product from this firm. So the orders taken from `product_to_output_orders` are, presumably, orders placed by those purchasers, which means that "inventory" is the _output_ inventory. But then you use this quantity in `order_product`, which orders a product from _some other_ producer. So the code is mixing up input and output products.

Agent-Based Simulation/src/Firm.cpp
```diff
@@ -141,14 +143,6 @@ void Firm::reorder_product_to_threshold(
...
-                order->product->inputs_per_unit.begin(),
-                order->product->inputs_per_unit.end(),
-                0.0,
-                [](double acc, const std::pair<Product *, double>& p) {
-                    return acc + p.second;
-                }
-                ) * order->quantity;
         }
     }
 }
```
Amittai-Aviram, line 147: OK, it looks as if this function is about reordering to replenish the _input_ inventory — but then it should not be using the "pending inventory" value computed by `get_pending_inventory`.

Agent-Based Simulation/src/Firm.cpp
```diff
@@ -141,14 +143,6 @@ void Firm::reorder_product_to_threshold(
         Producer * chosen_producer = send_order(order);
         if (chosen_producer) {
```
Amittai-Aviram, line 142: You have removed this, but now where do the credits get transferred from the plan to product the output product to the debt line item of the upstream producer's plan to produce the upstream product (i.e., the upstream producer's output product, which is the current firm's input product)?

Agent-Based Simulation/src/Person.cpp
```diff
@@ -114,16 +114,26 @@ float Person::avg_productivity_over_time_step(std::string product_name) {
...
+        return;
+    }
+    Logger::get_instance()->log(
+            Logger::PERSON,
+            "no_distributor",
+            id,
+            p->product_name,
+            quantity
+            );
 }
```
Amittai-Aviram, line 126: i think Devin has also made changes to this logic, so please check, and, if necessary, merge carefully, making sure that the result makes sense at every point.

Agent-Based Simulation/src/Producer.cpp
```diff
@@ -114,6 +114,8 @@ void Producer::start_plan(Plan * plan) {
             plan->order->product->inputs_per_unit) {
 		inventory[input.first] -= input.second * plan->order->quantity;
 	}
+    pooled_account += plan->raw_materials;
+    plan->raw_materials = 0;
```
Amittai-Aviram, line 118: Yes — correct.

Agent-Based Simulation/src/Producer.cpp
```diff
@@ -131,7 +133,7 @@ void Producer::move_plan_forward_one_step(Plan * plan) {
         }
 	} else {
 		raw_materials_used =
-            plan->raw_materials *
+            pooled_account *
             labor_hours_done /
             (plan->labor_hours - plan->workers.size() * plan->training_time);
```
Amittai-Aviram, line 137: I do not understand this. Could you please explain how this gives you the raw materials used? What exactly does that measure?

Agent-Based Simulation/src/Producer.cpp
```diff
@@ -131,7 +133,7 @@ void Producer::move_plan_forward_one_step(Plan * plan) {
...
         inputs_used = std::accumulate(
```
Amittai-Aviram: How is `inputs_used` different from `raw_materials_used`?  And what will we use this for? This aggregates the total number of units of all the inputs combined — but the inputs products themselves, which are things, are not pooled into a single stockpile.

Agent-Based Simulation/src/Producer.cpp
```diff
@@ -147,10 +149,33 @@ void Producer::move_plan_forward_one_step(Plan * plan) {
 	for (Person * worker : plan->workers) {
 		worker->register_hours_worked((double) labor_hours_done / plan->workers.size());
 	}
-	plan->labor_hours_remaining -= labor_hours_done;
-	plan->raw_materials_remaining -= raw_materials_used;
+    plan->labor_hours_remaining -= labor_hours_done;
+    plan->raw_materials_remaining -= raw_materials_used;
+    for (auto& input : input_inventory) {
```
Amittai-Aviram: Please replace `auto`.

Agent-Based Simulation/src/Producer.cpp
```diff
@@ -147,10 +149,33 @@ void Producer::move_plan_forward_one_step(Plan * plan) {
...
+        if (input.second < RAW_MATERIAL_THRESHOLD) {
+            reorder_raw_materials(plan, input.first);
+        }
```
Amittai-Aviram, line 148: OK, this part makes sense.

Agent-Based Simulation/src/Producer.cpp
```diff
@@ -147,10 +149,33 @@ void Producer::move_plan_forward_one_step(Plan * plan) {
...
+
+int Producer::get_pooled_account() {
+    return pooled_account;
+}
+
+void Producer::reorder_raw_materials(Plan * plan, Product * product) {
+    Society * society = Society::get_instance();
+    for (Producer * producer : society->get_producers()) {
+        if ((producer->inventory[product] > RAW_MATERIAL_THRESHOLD * 1.5)) {
+            producer->inventory[product] -= RAW_MATERIAL_THRESHOLD * 1.2;
```
Amittai-Aviram: What are the 1.5 and 1.2? And why aren't they defined as symbolic constants?

Agent-Based Simulation/src/Producer.cpp
```diff
@@ -147,10 +149,33 @@ void Producer::move_plan_forward_one_step(Plan * plan) {
...
+            plan->prd += RAW_MATERIAL_THRESHOLD * 1.2;
```
Amittai-Aviram: The value should go to the upstream producer's plan's PRD account, not to the current firm's plan's PRD account.

Agent-Based Simulation/src/Producer.cpp
```diff
@@ -147,10 +149,33 @@ void Producer::move_plan_forward_one_step(Plan * plan) {
...
+            pooled_account -= RAW_MATERIAL_THRESHOLD * 1.2;
+            break;
+        }
+    }
+    double threshold = get_reorder_threshold(product);
+    reorder_product_to_threshold(product, RAW_MATERIAL_THRESHOLD, RAW_MATERIAL_THRESHOLD * 1.2);
```
Amittai-Aviram: Again, move the 1.2 to a symbolic constant.

Amittai-Aviram (changes requested):
I think you're headed in the right direction, but this will all make better sense once you implement the distinct input and output inventories and the distinct algorithms for replenishing the inventories of those two distinct kinds.

Agent-Based Simulation/header/Constants.h
ghost: This is in units and it does apply to each stockpile of an input product

Agent-Based Simulation/header/Logger.h
ghost: Reset those changes

Agent-Based Simulation/src/Producer.cpp
ghost: Changed to `std::pair<Product *const, int>& input : input_inventory`

Agent-Based Simulation/src/Producer.cpp
ghost: No changes

Agent-Based Simulation/src/Firm.cpp
ghost: Changed to  `std::unordered_map<Product *, int>::iterator it = inventory.find(product);`

Agent-Based Simulation/src/Firm.cpp
ghost: Changed to `return inventory[product];`

Agent-Based Simulation/src/Producer.cpp
Amittai-Aviram: Sorry — I still see the `auto` in the code I just pulled from your branch.

Agent-Based Simulation/src/Producer.cpp
Amittai-Aviram: But could you please replace `auto` here?

Amittai-Aviram (changes requested):

Agent-Based Simulation/src/Producer.cpp
ghost: I haven't pushed yet into my branch, just writing out all the changes from my local changes

Agent-Based Simulation/src/Person.cpp
ghost: Reset changes

Agent-Based Simulation/src/Producer.cpp
ghost: no changes

Agent-Based Simulation/src/Producer.cpp
ghost: when the plan gets started for a producer, the `plan->raw_materials` becomes 0 and that original value gets transferred over to `pooled_account` which is now changed to `inputs_account`

Agent-Based Simulation/src/Producer.cpp
ghost: Changed to:

 `const double RAW_MATERIAL_SURPLUS_FACTOR = 1.5;
const double RAW_MATERIAL_ORDER_MULTIPLIER = 1.2;`

Agent-Based Simulation/src/Producer.cpp
ghost: Changed to deduct from upstream producer's plan's PRD account:

`plan->firm = producer; 
            plan->prd += product->price_per_unit * (RAW_MATERIAL_THRESHOLD * RAW_MATERIAL_ORDER_MULTIPLIER); // This is the upstream producer's plan
            producer->start_plan(plan);
            input_products_account -= product->price_per_unit * (RAW_MATERIAL_THRESHOLD * RAW_MATERIAL_ORDER_MULTIPLIER); `

Agent-Based Simulation/src/Producer.cpp
ghost: Updated to a symbolic constant

Agent-Based Simulation/header/Producer.h
ghost: Changed to the appropriate field names

Agent-Based Simulation/header/Producer.h
ghost: Moved it down below to the other log methods

Agent-Based Simulation/header/Firm.h
ghost: Changed to `input_products_account`

Agent-Based Simulation/src/Firm.cpp
ghost: Changed function name to `get_pending_output_inventory(Product * product)`

Agent-Based Simulation/src/Firm.cpp
ghost: Firm (mainly distributor) reorders output products from producers. Changed function name to `reorder_output_product_to_threshold`

Amittai-Aviram: I'm sorry — that name change makes matters worse, I think. For any firm, "output" should always refer to that firm's output. In the special case of the distributor, there is no output inventory, because the distributor does not produce anything — the consumers draw from the input inventory.  So, anyway, it does not make sense to me to say that a firm orders _output_ products from another firm. From the firm's point of view, those are input products. It's only from the upstream firm's POV that they are output products. I think I will let this slide so as not to hold you up any more, but I will review the code later (during the break) and try to figure out what things should be called what.

Agent-Based Simulation/src/Producer.cpp
Amittai-Aviram: I'm sorry. Your explanation does not match what I see above in lines 134 – 137, including parts of the arithmetic that are _not_ specifically part of this PR. I'm afraid we probably need to discuss this.

Amittai-Aviram (approved):
Sal, at least two problems remain, to my mind, in this PR, but I do not want to hold the changes up, so I am approving the PR but have also created new issues #139 and #140 and assigned them to you.  Please review those.
Please don't forget to switch to main (locally), pull, switch to your branch, merge from main, resolve conflicts, commit, and push.

Amittai-Aviram:
BEFORE YOU MERGE: Please remember to switch to main (on your machine), pull, switch to your branch, merge main into your branch, resolve conflicts, commit, and push.

PR #138, opened 2026-02-26
Updated README for the trade interface. 

Kkenize:
The README file now contains the full information for what's happening in the simulation.


Amittai-Aviram (approved):
Zhejun, this is mostly very good for a README-type full documentation of the project. I say "mostly" because there are a few particular language issues, but I'd rather suggest edits on that level later on. Meanwhile, i'm afraid that this is way too long and too detailed for a poster abstract. The poster abstract should be limited to 250 words. It is your "elevator pitch" — you would have to be able to recite it during the time available from when you get onto an elevator to when you get off (in a realistic building).
Also, you present the project as if you had developed the whole CE simulation model from scratch, but that was mostly Alex's work, inspired by work by Ian Wright.  For you rposter, you need to concentrate on _your contribution_ to this project, but figure out how to explain _that contribution_ to a non-expert who knows nothing about the LTEWG project, nothing about the LTE concept, only a little about international trade and finance, etc.

PR #142, opened 2026-03-03
Devins bugfix branch

### LINKED ISSUE #133: Major issues preventing a stable simulation (Devin)
These are ordered in order of importance:
 - Workers are not being reallocated to a Firm or Society after a plan is finished. After a while, this leaves a pool of zero workers for Producers to choose from, causing them all to decline.
 - People always consume up to their account, effectively ignoring Product consumption frequencies. This causes demand spikes that Distributors can't keep up with, and since People are left with an empty account, there is no demand after the first purchase.
 - The initial inventories set in Producer and Distributor are still running out long before demand signals stabilize. 
    - Currently, its set as a multiplier of the initial threshold, which has no information on how long it might take for demand to stabilize. In fact, I had to set `FIRM_INITIAL_INVENTORY_MULTIPLIER = 1500` for it to work as intended. Changing this to look like to `[some measure of initial product demand] * STARTING_NUM_PEOPLE * (FIRM_STOCKPILE_DURATION + k * FIRM_DEMAND_WINDOW)` will hopefully fix this issue in all cases.
 - Math in drafting a Plan - `labor_hours_total` is mistakenly calculated as hours done by a single worker, not a total for all workers.
    - This is causing large orders with large amounts of workers allocated to be completed much faster than the deadline, and small orders with small amounts of workers allocated to be completed much slower, when in reality they should take the same amount of time. In short, the predicted turnaround time is completely inaccurate.
    - Multiplying by a factor of `plan->workers.size()` seems to solve the issue
 - Constants tuning
    - People consume far more than they are able to produce. We should think of the consumption frequency as a person's bare minimum needs, so its ok if it is small. Having `PRODUCT_LABOR_PER_UNIT * PRODUCT_CONSUMPTION_FREQUENCY = 0.03` is working good for me.

After informally writing fixes to these on my own, the Distributors in the simulation seem to accurately track the demand and reorder accordingly, and the Producer's plans are being finished at the times the draft plans specify. Looking at the logs of Distributor inventory, it has a cyclical nature, with regular shipments coming in that match when the Distributor is soon to run out of a product. The simulation seems to be stable up to at least 50,000 timesteps.

One bug still unexplained:
 - Products' `living_labor_per_unit` seems to mysteriously rise steadily, causing a noticeable steady increase in price as time goes on. This quantity should be a constant for each Product, and I also can't find anywhere where this field is set besides the constructor.

lennyyyyyyyy:
Fixes major bugs described in #133. Resolves #133 

Also fixes some bugs not mentioned in the above issue:
- Floating-point rounding errors causing plans not to be completed when hours are used up
- Plans being started (and consuming the required inputs) multiple times, redundantly.

Also adds some non-bug changes:
- Separate constants for Producer's and Distributor's initial inventory multiplier 
- Useful logs
- Create a demand window minimum and maximum length, changing the time averaged over depending on the datapoints currently available


Agent-Based Simulation/header/Constants.h
```diff
@@ -29,29 +30,31 @@ const double AVERAGE_RETIREMENT_AGE = (
     INITIAL_AGE; // approx 62 years in hours
 
 const int PRODUCT_ORDER_SIZE_MIN = 1;
-const int PRODUCT_ORDER_SIZE_MAX = 1000;
+const int PRODUCT_ORDER_SIZE_MAX = 10;
 const int PRODUCT_NUM_INPUTS_MIN = 0;
-const int PRODUCT_NUM_INPUTS_MAX = 10;
-const double PRODUCT_INPUT_PER_UNIT_MIN = 0.01;
+const int PRODUCT_NUM_INPUTS_MAX = 5;
```
Amittai-Aviram: Why this change?

Agent-Based Simulation/header/Constants.h
```diff
@@ -29,29 +30,31 @@ const double AVERAGE_RETIREMENT_AGE = (
...
+const double PRODUCT_INPUT_PER_UNIT_MIN = 0.1;
 const double PRODUCT_INPUT_PER_UNIT_MAX = 0.25;
 const double PRODUCT_INPUT_EPSILON = 0.0001;
 const int PRODUCT_NUM_MACHINES_MIN = 0;
 const int PRODUCT_NUM_MACHINES_MAX = 5;
-const int PRODUCT_LABOR_PER_UNIT_MIN = 1;
-const int PRODUCT_LABOR_PER_UNIT_MAX = INITIAL_WORK_WEEK; 
+const double PRODUCT_LABOR_PER_UNIT_MIN = 0.05;
+const double PRODUCT_LABOR_PER_UNIT_MAX = 0.1;
```
Amittai-Aviram, line 44: The most a worker will spend making one unit is 6 minutes? Why is this change necessary?

Agent-Based Simulation/src/Society.cpp
```diff
@@ -160,19 +160,17 @@ void Society::adjust_io_matrix(
...
 }
 
-Eigen::VectorXd get_leontief_function(
-        Eigen::MatrixXd io_matrix, 
-        Eigen::VectorXd labor
+Eigen::MatrixXd get_leontief_inverse(
+        Eigen::MatrixXd io_matrix
         ) {
-    Eigen::MatrixXd io_matrix_transpose = io_matrix.transpose();
+    // Eigen::MatrixXd io_matrix_transpose = io_matrix.transpose();
```
Amittai-Aviram: Please delete.

Agent-Based Simulation/src/Society.cpp
```diff
@@ -234,10 +242,11 @@ void Society::set_initial_account() {
     initial_account = 0.0;
     for (Product * product : products) {
         ConsumerGood * consumer_good = get_consumer_good(product);
-        if(!consumer_good) {
-            std::cerr << "consumer good DNE" << std::endl;
+        if (consumer_good) {
```
Amittai-Aviram: Under what condition would a consumer good for this product not exist? Is it indeed an error condition?

Agent-Based Simulation/src/Society.cpp
```diff
@@ -234,10 +242,11 @@ void Society::set_initial_account() {
...
+            initial_account += consumer_good->price_per_unit *
+                consumer_good->mean_consumption_frequency *
+                FIRM_DEMAND_WINDOW_MIN * INITIAL_ACCOUNT_MULT;
```
Amittai-Aviram: This arithmetic — with the last two factors — stikes me as opaque. The last two factors seem like pure fudge factors. Are they? Is there any way that we can avoid them?

Agent-Based Simulation/src/Firm.cpp
```diff
@@ -369,18 +369,31 @@ void Firm::train_workers(
...
 }
 
 void Firm::apply_demand_window() {
-    while (!demand_signals.empty() && 
-            demand_signals.front().timestep <= 
-            Sim::get_current_time_step() - FIRM_DEMAND_WINDOW) {
-        inventory_demands[demand_signals.front().product] -= 
-            (double) demand_signals.front().quantity / FIRM_DEMAND_WINDOW;
-        demand_signals.pop();
+    for (auto& product : demand_signals) {
```
Amittai-Aviram: Please DO NOT USE AUTO unless the type is impossibly complicated!  ☹️

Agent-Based Simulation/src/Firm.cpp
```diff
@@ -369,18 +369,31 @@ void Firm::train_workers(
...
+            total_demands[product.first] -= 
+                signals.front().quantity;
+            signals.pop();
+        }
+    }
+}
+
+double Firm::get_demand(Product * product) {
+    int window_start = Sim::get_current_time_step();
+    if (!demand_signals[product].empty()) {
```
Amittai-Aviram, line 388: Can we be sure that `demand_signals` will have an entry for `product`? If not, even if that is conceptually an error condition, we probably want to throw an exception with a meaningful message (including the product name or ID) rather than just let the program crash with a segmentation fault:
```
if (!demand_signals.count(product)) { ...
```

Agent-Based Simulation/src/Firm.cpp
```diff
@@ -369,18 +369,31 @@ void Firm::train_workers(
...
+        window_start = demand_signals[product].front().timestep;
```
Amittai-Aviram, line 389: `front` has the most recent time step, right? Is that what you meant? Did you mean `back`?

Agent-Based Simulation/src/Firm.cpp
```diff
@@ -396,7 +409,7 @@ void Firm::log_shipment_received(std::string product_name, int quantity) {
 void Firm::log_inventory_level(std::string product_name, int quantity) {
     Logger::get_instance()->log(
             Logger::FIRM,
-            "inventory_level",
+            "inventory_level " + product_name,
```
Amittai-Aviram: If we write CSV files, this will result in one CSV file for each product listed in any inventory, with the file name containing a space — **Firm\_inventory\_level Product1.csv**, etc. Instead, you should probably use the Logger API that takes a string and a numeric value, so that the resulting file (and DB table) will have all the inventory levels together, with each row distinguishing the product. The same applies to the JSON logging output.

Agent-Based Simulation/src/Firm.cpp
```diff
@@ -416,9 +429,19 @@ void Firm::log_reorder(std::string product_name, int quantity) {
...
             id,
             product_name,
             requested_turnaround_time
             );
 }
+
+void Firm::log_demand(std::string product_name, double demand) {
+    Logger::get_instance()->log(
+            Logger::FIRM,
+            "demand " + product_name,
```
Amittai-Aviram, line 429: See comment above re logging.

Agent-Based Simulation/src/Person.cpp
Amittai-Aviram: It might be more efficient to keep this as a member rather than a local variable and then update the values on each call to `shop`.

Agent-Based Simulation/src/Person.cpp
```diff
@@ -141,11 +135,10 @@ void Person::shop() {
     }
     log_shopping_deficit(std::max(0.0, (total_price - account) / total_price)); 
     for (std::pair<Product *, double> p : ideal_purchase_quantities) {
-        int affordable_quantity = (int) (account / total_price *
-                ideal_purchase_quantities[p.first]);
-        if (affordable_quantity > 0) {
-            purchase_good(p.first, affordable_quantity);
-            log_shopping(p.first->product_name, affordable_quantity);
+        int quantity = (int) (std::min(account / total_price, 1.0) * p.second);
```
Amittai-Aviram, line 137: What is the rationale here? In real life, the quantity would be determined independently by need.

Agent-Based Simulation/src/Producer.cpp
```diff
@@ -29,7 +29,9 @@ Producer::Producer(
     }
     for (Product * product : get_products_to_reorder()) {
         inventory[product] =
-            get_reorder_threshold(product) * FIRM_INITIAL_INVENTORY_MULTIPLIER;
+            (society->get_initial_production()[product] - product->mean_consumption_frequency) * 
+            (FIRM_STOCKPILE_DURATION + FIRM_DEMAND_WINDOW_MIN * PRODUCER_INITIAL_INVENTORY_MULT) * 
+            STARTING_NUM_PEOPLE;
```
Amittai-Aviram, line 34: This arithmetic is, again, a bit opaque.  Maybe you can just explain it to me in person or in a document.

Amittai-Aviram (approved):
Thank you for the great work! I am asking for a few explanations and also a few (fairly small) changes.  For where I ask for an explanation, if you _can_ explain the point in writing, that's probably better, but, if it's difficult or takes too much time, we can just discuss it.

Amittai-Aviram (changes requested):
Sorry! I clicked on "Approve," but I had meant to select "Request changes."  Please see my previous comments.

Agent-Based Simulation/header/Constants.h
alexbcreiner0: Having max and min constants which are separate from the number of products is a bad idea. These should be dependent variables, i.e. they should be functions of the number of products. Maybe something like NUM_PRODUCTS / 3 for the min, or something like that.

lennyyyyyyyy: Done.

Agent-Based Simulation/header/Constants.h
lennyyyyyyyy: I was just tuning I did to make production at least greater than consumption. I think we will use a better solution later: see #144

Agent-Based Simulation/src/Society.cpp
lennyyyyyyyy: Done.

Agent-Based Simulation/src/Society.cpp
lennyyyyyyyy: You're right, it is an error condition. I can't remember why I got rid of that.

Agent-Based Simulation/src/Society.cpp
lennyyyyyyyy: I don't think so. These sort of make sure that people are able to buy what they need for FIRM_DEMAND_WINDOW_MIN amount of time before they run out of money, to give a good starting value for demand signals. I think I'll move `* FIRM_DEMAND_WINDOW_MIN * INITIAL_ACCOUNT_MULT` to the end of the function to make it more transparent.

Agent-Based Simulation/src/Firm.cpp
lennyyyyyyyy: Done.

Agent-Based Simulation/src/Firm.cpp
lennyyyyyyyy: Because it is unordered map, it'll create a new entry for a `std::queue` anyways, before it calls `.empty()` (so there's no seg fault). There's no problem with this, because we want all the queues for products in `get_products_to_reorder()` to start out empty from the beginning of time.

Agent-Based Simulation/src/Firm.cpp
lennyyyyyyyy: C++ queues use the `push_back` and `pop_front` methods of the underlying container (if it's compatible), so front will have the oldest demand signal still in the queue.

Agent-Based Simulation/src/Firm.cpp
lennyyyyyyyy: I did this originally, the problem is that the Logger's map keys only by client, label, id, and timestep. The product name (Product1, Product2, etc.) is put in the tuple and set as the value. This basically meant that for every timestep, each Product entry for a specific Firm (which would have the same client, label, id, and timestep) would overwrite the last, and I could only see the inventory of the last product for that timestep. 

So, I added the product name to the label, which gave me separate files for each Product. It was kind of a hasty solution, so we might have to change how this works.

Agent-Based Simulation/src/Firm.cpp
lennyyyyyyyy: See above.

Agent-Based Simulation/src/Person.cpp
lennyyyyyyyy: Done.

Agent-Based Simulation/src/Person.cpp
lennyyyyyyyy: This has the effect of buying up to the account, if the account is less than the total price of needed goods, which is the same as what we were doing before.

However, if the person has more than enough money to buy needed goods, they don't spend their entire account (which was the bug that was causing problems before). Now, the left side of the product goes to `1.0` and they buy exactly what they need.

Agent-Based Simulation/src/Producer.cpp
lennyyyyyyyy: Yeah, I can explain this on Sunday.

Amittai-Aviram (approved):
Please remove the logging call for inventory levels, pending changes soon to be made and a possible redesign of part of the logging API.

Issue #139, opened 2026-02-26
Clarify names in `Firm`


Amittai-Aviram:
Some of the names of functions and variables are still confusing, in particular in relation to to the opposition of _input_ to _output_. The confusion caused by these names may be bleeding into the actual logic — it's hard to tell until we make our names more explanatory and precise.

Issue #131, opened 2026-02-17
Initialize Input Inventory for Producers


ghost:
This is a small adjustment. I made a new `unordered_map` for inputs in the inventory initialized as `std::unordered_map<Product *, int> input_inventory` which holds the Producer's inputs. All that is needed in `Society.cpp`:
- [ ] Initialize an input inventory for each Producer
- [ ] When Producer starts a plan, it checks whether the inputs exist for that input inventory (first time around it won't be that's when you add it to the unordered_map and make the qt. 0) and if it's already there then it needs to check if it's already up to the threshold otherwise order from a `Producer`
- [ ] Make sure that on each time step the Producer's inputs decrease

Amittai-Aviram:
@Sal, this is a high-priority item.

PR #147, opened 2026-03-08
Initialize Inputs to Society and Inputs Inventory for Both Distributor and Producer

### LINKED ISSUE #124: Create graphs for transactions between `Producer` and `Distritbutor`
Make graphs for the simulation to show the relation between `Distributor` and `Producer`:

- Must include details for how many goods the `Producer` made over time and how many goods were sold by the `Distributor`
- Change in pooled accounts over time
- If time permits, include graphs of consumption by `Person` in society

(Need more time to revise this issue)

Sal47506:
- Producer and Distributor now initializes their input inventory in their constructor.
- Society also initializes the input inventory `unordered_map` in the constructor with the quantity of each input being a random number using `Sim::get_random_generator`
- Initial code for making graphs in `cpp` have been added (still a lot more work need to be done)
- Variable and function names were also changed in Firm to address the previous lack of clarity
- Logging added in `Firm` to log input inventory

Resolves #124 #131 #130 #139


alexbcreiner0:
Hey Sal, with respect to the plotting, why are we putting energy into that when my wrapper software exists to facilitate that purpose? If you want to spend some time meeting with me, I can show you how to use it to plot whatever you want using it. I think it's better if we try to keep the code for plotting separated from the code for the simulation itself (not to mention it's better written in Python since the C++ matplot library is just a wrapper around it if I recall correctly from our work in Summer).

Agent-Based Simulation/src/Producer.cpp
```diff
@@ -15,9 +15,11 @@
 
 Producer::Producer(
         Society * society,
-        std::unordered_set<Product *> initial_catalog
+        std::unordered_set<Product *> initial_catalog,
```
Amittai-Aviram: Shouldn't this be passed in by reference?

Agent-Based Simulation/src/Producer.cpp
```diff
@@ -15,9 +15,11 @@
...
+        std::unordered_map<Product *, int> input_inventory
```
Amittai-Aviram: Ditto. (Reference.)

Agent-Based Simulation/src/Producer.cpp
```diff
@@ -15,9 +15,11 @@
...
         ) :
-    Firm(society, initial_catalog) {
+    Firm(society, initial_catalog, input_inventory),
```
Amittai-Aviram: And please change `Firm`'s constructor to take the catalogue and inventory by reference.

Agent-Based Simulation/src/Society.cpp
```diff
@@ -31,15 +31,33 @@ Society::Society() {
...
         Logger::get_instance()->log(Logger::SOCIETY, "order_size", product->id, product->order_size);
     }
+    static std::uniform_int_distribution<> initial_inventory_dist(
+        PRODUCT_ORDER_SIZE_MIN, PRODUCT_ORDER_SIZE_MAX);
     for (int i = 0; i < STARTING_NUM_PRODUCERS; i++) {
-        Producer * producer = new Producer(this, {goods[i %
-                STARTING_NUM_PRODUCTS]});
+        Product * product = goods[i % STARTING_NUM_PRODUCTS];
+        int initial_quantity = initial_inventory_dist(Sim::get_random_generator());
+        std::unordered_map<Product *, int> initial_input_inventory = {
```
Amittai-Aviram, line 39: Wait. Why is the product that the producer will produce (which is the first entry in its catalog) the same as the product in its _input_ inventory? Doesn't it need the product's _inputs_ to be in its input inventory?

Agent-Based Simulation/src/Society.cpp
```diff
@@ -31,15 +31,33 @@ Society::Society() {
...
         producers.push_back(producer);
         firms.push_back(producer);
     }
     for (int i = 0; i < STARTING_NUM_DISTRIBUTORS; i++) {
-        Distributor * distributor =
-            new Distributor(this, {goods[i % STARTING_NUM_PRODUCTS]});
+        Product * product = goods[i % STARTING_NUM_PRODUCTS];
+        int initial_quantity = initial_inventory_dist(Sim::get_random_generator());
+        std::unordered_map<Product *, int> initial_input_inventory = {
+            {product, initial_quantity}
```
Amittai-Aviram, line 54: Yes. This is OK for a _distributor_, because a distributor has only one inventory, its _input_ inventory. This holds the products that come _into_ the distributor, which are also the products that go _out_ to consumers.  But this logic does not make sense for producers.

Agent-Based Simulation/src/Producer.cpp
```diff
@@ -15,9 +15,11 @@
...
+    input_inventory(input_inventory) {
```
Amittai-Aviram: Why are you initializing both the superclass's and the subclass's input inventory? There should be only one input inventory, which should belong to the superclass (`Firm`), and, as I noted in `Society`, it should be the _input_ inventory — it should hold _inputs_,  not output products.  (It's possible for a product to have itself as an input, but that is determined by the product's input map in any event.)

Agent-Based Simulation/src/Producer.cpp
```diff
@@ -50,7 +52,7 @@ bool Producer::can_produce(Product * product) {
 int Producer::draft_plan(Order * order) {
     for (std::pair<Product * const, double>& input :
             order->product->inputs_per_unit) {
-        if (inventory[input.first] < input.second * order->quantity) {
+        if (input_inventory[input.first] < input.second * order->quantity) {
```
Amittai-Aviram, line 54: Correct.

Agent-Based Simulation/src/Producer.cpp
```diff
@@ -119,9 +121,9 @@ void Producer::start_plan(Plan * plan) {
 	// simplification: consume all raw materials at start of plan
 	for (std::pair<Product * const, double>& input :
             plan->order->product->inputs_per_unit) {
-		inventory[input.first] -= input.second * plan->order->quantity;
+		input_inventory[input.first] -= input.second * plan->order->quantity;
 	}
-    input_products_account += plan->raw_materials;
+    pooled_input_value_account += plan->raw_materials_remaining;
```
Amittai-Aviram: Why do you use `raw_materials_remaining` rather than simply `raw_materials`?  (Typo.)

Agent-Based Simulation/src/Producer.cpp
```diff
@@ -119,9 +121,9 @@ void Producer::start_plan(Plan * plan) {
...
     plan->raw_materials = 0;
     plan->total_hours_remaining -= 1e-6;
```
Amittai-Aviram: I know that this is not part of your current change set, but why are we doing this? I don't understand it.

Agent-Based Simulation/src/Producer.cpp
Amittai-Aviram, line 132: This function has gotten to be too big and complicated. It should be raised to a higher level of abstraction and then call a sequence of helper functions. Please write an issue to do this throughout the code base for all big and complicated functions.

Agent-Based Simulation/src/Producer.cpp
```diff
@@ -139,38 +141,63 @@ void Producer::move_plan_forward_one_step(Plan * plan) {
...
-            labor_hours_done /
-            (plan->labor_hours - plan->workers.size() * plan->training_time);
+        // Consume this plan's remaining raw-material value in proportion
+        // to productive labor completed this step.
+        if (labor_hours_done > 0 && plan->labor_hours_remaining > 0) {
+            raw_materials_used =
+                pooled_input_value_account *
+                static_cast<double>(labor_hours_done) /
+                plan->labor_hours_remaining;
+        }
```
Amittai-Aviram: I do not understand this logic.

Agent-Based Simulation/src/Producer.cpp
```diff
@@ -139,38 +141,63 @@ void Producer::move_plan_forward_one_step(Plan * plan) {
...
+    return pooled_input_value_account;
+}
+
+int Producer::get_pending_input_inventory(Product * product) {
+    int pending_inventory = input_inventory[product];
+    for (Order * order : product_to_outbound_orders[product]) {
+        pending_inventory += order->quantity;
+    }
+    return pending_inventory;
 }
```
Amittai-Aviram: I do not understand this function. It looks to me as if it's mixing inputs with outputs.

Agent-Based Simulation/src/Producer.cpp
```diff
@@ -139,38 +141,63 @@ void Producer::move_plan_forward_one_step(Plan * plan) {
...
 
 void Producer::reorder_raw_materials(Plan * plan, Product * product) {
     Society * society = Society::get_instance();
```
Amittai-Aviram: Do you need this? Isn't Society's address already a member?

Agent-Based Simulation/src/Producer.cpp
```diff
@@ -139,38 +141,63 @@ void Producer::move_plan_forward_one_step(Plan * plan) {
...
     for (Producer * producer : society->get_producers()) {
-        if ((producer->inventory[product] > RAW_MATERIAL_THRESHOLD * RAW_MATERIAL_SURPLUS_FACTOR)) {
-            producer->inventory[product] -= RAW_MATERIAL_THRESHOLD * RAW_MATERIAL_ORDER_MULTIPLIER;
+        if ((producer->input_inventory[product] > RAW_MATERIAL_THRESHOLD * RAW_MATERIAL_SURPLUS_FACTOR)) {
+            producer->input_inventory[product] -= RAW_MATERIAL_THRESHOLD * RAW_MATERIAL_ORDER_MULTIPLIER;
+            input_inventory[product] += RAW_MATERIAL_THRESHOLD * RAW_MATERIAL_ORDER_MULTIPLIER;
             plan->firm = producer;
```
Amittai-Aviram: ??

Agent-Based Simulation/src/Producer.cpp
```diff
@@ -139,38 +141,63 @@ void Producer::move_plan_forward_one_step(Plan * plan) {
...
 void Producer::reorder_raw_materials(Plan * plan, Product * product) {
     Society * society = Society::get_instance();
     for (Producer * producer : society->get_producers()) {
-        if ((producer->inventory[product] > RAW_MATERIAL_THRESHOLD * RAW_MATERIAL_SURPLUS_FACTOR)) {
-            producer->inventory[product] -= RAW_MATERIAL_THRESHOLD * RAW_MATERIAL_ORDER_MULTIPLIER;
+        if ((producer->input_inventory[product] > RAW_MATERIAL_THRESHOLD * RAW_MATERIAL_SURPLUS_FACTOR)) {
+            producer->input_inventory[product] -= RAW_MATERIAL_THRESHOLD * RAW_MATERIAL_ORDER_MULTIPLIER;
+            input_inventory[product] += RAW_MATERIAL_THRESHOLD * RAW_MATERIAL_ORDER_MULTIPLIER;
             plan->firm = producer; 
             plan->prd += product->price_per_unit * (RAW_MATERIAL_THRESHOLD * RAW_MATERIAL_ORDER_MULTIPLIER); // This is the upstream producer's plan
```
Amittai-Aviram: Yes, it should be the upstream producer's plan, but it's just passed in as a parameter.

Please implement the firm-to-firm transaction through APIs:
Downstream firm: gives value from pooled inputs account, receives input goods.
Upstream firm: receives value and enters it into PRD account, gives (output) goods.

Agent-Based Simulation/src/Society.cpp
```diff
@@ -31,15 +31,33 @@ Society::Society() {
         Logger::get_instance()->log(Logger::SOCIETY, "price", product->id, product->price_per_unit); 
         Logger::get_instance()->log(Logger::SOCIETY, "order_size", product->id, product->order_size);
     }
+    static std::uniform_int_distribution<> initial_inventory_dist(
+        PRODUCT_ORDER_SIZE_MIN, PRODUCT_ORDER_SIZE_MAX);
     for (int i = 0; i < STARTING_NUM_PRODUCERS; i++) {
-        Producer * producer = new Producer(this, {goods[i %
-                STARTING_NUM_PRODUCTS]});
+        Product * product = goods[i % STARTING_NUM_PRODUCTS];
```
Amittai-Aviram: Why are you using the iteration variable defined for producers to identify products?

Agent-Based Simulation/src/Producer.cpp
```diff
@@ -139,38 +141,63 @@ void Producer::move_plan_forward_one_step(Plan * plan) {
...
     plan->labor_hours_remaining -= labor_hours_done;
-    plan->raw_materials_remaining -= raw_materials_used;
+    plan->raw_materials_remaining = std::max(
+            0.0,
+            plan->raw_materials_remaining - raw_materials_used
+            );
+    pooled_input_value_account = std::max(
+            0.0,
+            pooled_input_value_account - raw_materials_used
+            );
```
Amittai-Aviram: The pooled input value account is only reduced when the producer replenishes an input inventory by buying the new supply batch from an upstream producer. So the pooled account should not be touched in this function. It should only be touched when you start a plan — where you add the raw materials budget to the pooled account — and when you reorder an input from an upstream producer.

Amittai-Aviram (changes requested):
Good progress so far, but I see a number of problems here, as noted. (My comments represent a _partial_ review.) Please respond to the comments as you fix the issues. Thanks!

Agent-Based Simulation/src/Producer.cpp
ghost: So 2 separate functions for this upstream and downstream exchange?

Agent-Based Simulation/src/Producer.cpp
ghost: I'm going to leave a comment soon explaining this part

ghost: (This explanation is regarding my newest update)

Basically, when a producer has an input product below the threshold then it subtracts the threshold constant by the current number of inputs it has for that product to find how many more goods it needs to meet the threshold, however you would need more than just that difference in order to not fall below the threshold immediately within the next few time steps so thats where the RAW_MATERIAL_ORDER_MULTIPLIER comes in. This means that reordering the difference by that multiplier would better ensure that the input product doesn't run out immediately and we also have to iterate over upstream producers that have enough of that input product where if they were to sell it to the downstream producer by `(RAW_MATERIAL_THRESHOLD - input_inventory[product]) * RAW_MATERIAL_ORDER_MULTIPLIER` then they should be able to have enough quantity of that input where that quantity doesn't fall below the threshold.

Agent-Based Simulation/src/Producer.cpp
ghost: Updated

Agent-Based Simulation/src/Producer.cpp
ghost: Updated

Agent-Based Simulation/src/Producer.cpp
ghost: Updated

Agent-Based Simulation/src/Society.cpp
ghost: Producer only has input inventory while Distributor only has output inventory

Agent-Based Simulation/src/Society.cpp
ghost: Distributor changed to only have output inventory

Agent-Based Simulation/src/Producer.cpp
ghost: Removed

Agent-Based Simulation/src/Society.cpp
ghost: Updated

Agent-Based Simulation/src/Producer.cpp
ghost: Removed

ghost:
Just made an update for firms to have a pooled account as a higher level of abstraction so I don't have to redefine them in Distributor and Producer. In addition, I have included the calculations for pooled account deductions when reordering goods in `Firm.cpp`

Amittai-Aviram:
Small change: `Firm::check_and_reorder_inputs` only calls `Firm::reorder_input_product_to_threshold` if the inventory level is below the threshold:
```
            reorder_input_product_to_threshold(product, threshold, pending_inventory);
        }
```
So we should not duplicate this test in `Firm::reorder_input_product_to_threshold`:
```
        // ...
    }
```
If you like, as a matter of keeping the functions somewhat independent, you can check for an error condition in the latter function instead:
```
    if (pending_inventory >= threshold) {
        std::cerr << "Reordering product " << product->name << " unnecessarily." << std::endl;
        return;
    }
```

Agent-Based Simulation/src/Firm.cpp
Amittai-Aviram, line 108: I'm sorry! This does not make sense. We have already set `order_time` to `MAX_INT`. _Any_ integer is going to have a value, by definition, <= `MAX_INT`. I think what should happen here is that `order_time` should be the order deadline, and we should check whether the draft plan time is <= order_time. If not, we've got an error condition that should not happen. If so, it's still possible that draft_plan_time is strictly less than order_time, in which case order_time should be revised downward  (I suppose — I'm not sure what the consequences of that would be anyway).

Agent-Based Simulation/src/Firm.cpp
```diff
@@ -97,6 +103,7 @@ Producer * Firm::send_order(Order * order) {
     }
     if (chosen_producer) {
         chosen_producer->pursue_order(order);
+        chosen_producer->plans_in_progress.back()->prd += order->product->price_per_unit * order->quantity;
```
Amittai-Aviram, line 115: I think this quantity should be subtracted from the downstream firm's pooled input value account and added to the upstream firm's PRD when the former receives the shipment from the latter.

Agent-Based Simulation/src/Firm.cpp
```diff
@@ -144,20 +151,21 @@ void Firm::reorder_output_product_to_threshold(
         if (chosen_producer) {
             log_reorder(product->product_name, reorder_quantity);
             log_accepted_order(product->product_name, order->requested_turnaround_time);
+            this->pooled_input_value_account -= product->price_per_unit * reorder_quantity;
```
Amittai-Aviram: This should happen at the end of `receive_shipment`. (See my comment there.) And you do not need `this`.

Agent-Based Simulation/src/Producer.cpp
```diff
@@ -139,51 +139,41 @@ void Producer::move_plan_forward_one_step(Plan * plan) {
...
+    return pooled_input_value_account;
 }
 
 void Producer::end_plan(Plan * plan) {
     log_ended_plan(plan);
     plan->order->status = Order::ORDER_FINISHED;
 	// simplification: whole product amount is added to inventory at the end of
     // a plan
-	inventory[plan->order->product] += plan->order->quantity;
+	input_inventory[plan->order->product] += plan->order->quantity;
```
Amittai-Aviram, line 182: This adds the new batch of products to its producer's own inventory! That doesn't make sense. The upstream producer, which is the one that calls this function (`end_plan`), is the one who will _ship_ the batch of products to the _downstream_ producer, who will receive the batch of goods into its inventory.

Agent-Based Simulation/src/Producer.cpp
```diff
@@ -139,51 +139,41 @@ void Producer::move_plan_forward_one_step(Plan * plan) {
...
 	// simplification: product shipped instantly
-	inventory[plan->order->product] -= plan->order->quantity;
+	input_inventory[plan->order->product] -= plan->order->quantity;
```
Amittai-Aviram, line 184: This just undoes what you did one statement earlier (on line 174).

Agent-Based Simulation/src/Producer.cpp
```diff
@@ -139,51 +139,41 @@ void Producer::move_plan_forward_one_step(Plan * plan) {
...
     plan->order->customer->receive_shipment(plan->order);
     double total = plan->order->product->price_per_unit * plan->order->quantity;
     plan->prd += total;
```
Amittai-Aviram: This should happen in the customer's (i.e., downstream firm's) `receive_shipment`.  This will require a new API function `Producer::receive_payment(Order * order, double payment)`, so please write an issue to create one, and, for now, leave this update of PRD here.

Agent-Based Simulation/src/Firm.cpp
```diff
@@ -78,9 +71,22 @@ void Firm::receive_shipment(Order * order) {
             << std::endl;
         return;
     }
-    inventory[order->product] += order->quantity;
+    input_inventory[order->product] += order->quantity;
```
Amittai-Aviram, line 73: Just before this the firm adds the batch (`order->quantity`) to its inventory is where it should take the value out of its pooled input value account to pay for this batch. It should also pay the upstream producer, which will require a new API. See my comment about that on `Producer::end_plan` (`producer.cpp`:179). The sequence in `receive_shipment` should be that the customer receives the batch and pays the upstream producer, and the upstream producer adds that payment to the plan that it had followed to produce that batch of the product.

Amittai-Aviram (commented):
We need to go over this together. I have added a few comments. There is one place where I see what looks like a bug: the upstream producer is adding its product to its own inventory.

Amittai-Aviram (approved):
LGTM

PR #151, opened 2026-03-15
Locally track Firm's living labor, add tuning and normalization between consumption and production rates, initiate shopping from Person inventory deficits

### LINKED ISSUE #134: Make Person more realistic and varied (Devin)
### Consumption 
 - Change product purchase frequencies to product consumption frequencies
 - Give each person a personal inventory vector and a deficit vector
 - On every time step:
    - Consume from the deficit vector
    - Subtract the minimum of the inventory or deficit of each product from both vectors
    - Log the deficit as a measure of personal satisfaction
 - Launch a **shopping event** on one of the following conditions
     - It is the correct time every week determined by `shopping_offset`
     - The magnitude of the deficit vector exceeds a threshold

### Shopping
- First, buy the amounts specified by the deficit vector, up to the price of the Person's current account. 
- Then, if the shopping event is "regular" (the one triggered weekly):
   - Buy a scaled bundle of "wants" (or maybe "tastes"), uniquely generated for each Person in its constructor, up to a constant percentage of the account the Person at the beginning of the shopping event.
   - Log the hours spent on these "wants" as another measure of satisfaction

### LINKED ISSUE #137: Rethink productivity, suitability, and updates to Products' `living_labor_per_unit` (Devin)
Something is causing `living_labor_per_unit` to drift up, which is tied to the average suitability of a draft plan consistently being centered around one.

### LINKED ISSUE #144: Outline for determining consumption coefficients, consumption behavior of persons, and shopping behavior of persons.
Even once we fix our full employment problem, we are going to have to view our consumption frequencies as semi-dependent variables which can't be fully random in their determination if we want a stable reproductive society.

In particular, let $\mathbf{C}$ denote the vector of consumption frequencies for all commodities. That is $C[i]$ is the average number of units of commodity $i$ which are consumed per hour all members of society (combined). Then for reproduction of the society to be possible, it must be the case that the total labor value of this bundle, i.e. the total gross labor time required to produce that bundle, is less than the total labor which the society can perform in an hour. This latter quantity is equal to the number of persons in the society, which we'll denote $N$. Letting $\mathbf{\lambda}$ denote the vector of unit labor values, we have the condition

$$ \mathbf{\lambda} \cdot \mathbf{C} = \sum_{i=1}^N \lambda[i]c[i] < N $$

If we approximate people's consumption habits by viewing $c[i] := \frac{C[i]}{N}$ as the number of units of commodity $i$ consumed by an individual each hour, then we have the simpler condition

$$ \mathbf{\lambda} \cdot \mathbf{c} < 1$$

There are a lot of probabilistic ways we could go about calculating such a vector $\mathbf{c}$. However, finding one which does not severely restrict the possibility space while still being tractable with respect to $N$ seems difficult. Here's a cheap suggestion which makes that compromise. We can revise this with something better later if anyone thinks of anything.

Let $\epsilon$ be some new constant parameter of the system. Its function is very similar to the scalar that we are dividing our I/O matrix by in order to force productivity. Call it the "degree of surplus tendency" of society, maybe. Consider $\mathbf{c} = \frac{1}{\epsilon|\\mathbf{\lambda}|^2}\mathbf{\lambda}$, where $|\mathbf{\lambda}|$ is the Euclidean norm of $\mathbf{\lambda}$. Then of course

$$ \mathbf{\lambda} \cdot \mathbf{c} =  \frac{1}{\epsilon}\frac{\mathbf{\lambda}\cdot \mathbf{\lambda}}{|\mathbf{\lambda}|^2} = \frac{1}{\epsilon}\frac{|\mathbf{\lambda}|^2}{|\mathbf{\lambda}|^2} = \frac{1}{\epsilon} < 1$$

The actual per-person consumption frequencies could safely be determined by sampling from a normal distribution whose means are the actual consumption frequencies calculation according to the above equation in the constructor for Society, right after setting the initial initial prices ($\mathbf{\lambda}$). It will likely require some trial and error testing to determine just how large the variance of this distribution is allowed to be. Something small for now and we can inch up. 

On a related note, this issue should be seen as bound at the hip to  [Devin's issue](https://github.com/Amittai-Aviram/BC-LTEWG/issues/134) on revising how the Person class goes about consuming and shopping. Purchases need to be accompanied by an actual transfer of the products from the distributor to the consumers. To do this, we should give them an endowment vector $\mathbf{e}$, basically an inventory like what a firm would have. For every hour that passes, if the modulo of the current hour with a consumption frequency $c[i]$ ever equals zero, then we should decrement 1 from the appropriate entry of this vector.

I suggested in addition that we have a deficit vector $\mathbf{d}$ which we increment when the endowment vector goes to zero but the person wishes to consume, and then have a person go shopping when $|\mathbf{d}| > \gamma$ where $\gamma > 0$ is some agitation threshold. Devin rightfully pointed out that this wouldn't be realistic, because people go out to the store to stock up on something before running out of it, i.e. pre-emptively. Here is a revised version of my suggestion which I believe addresses this while also binding the shopping system more cleanly to the consumption system. Instead of having a deficit vector at all, just allow the endowment vector $\mathbf{e}$ to go negative in its entries. In addition to an agitation threshold $\gamma$, have a vector lower bound $\mathbf{g}$ representing the minimal bundle that a person wishes to possess at a given moment. Instead of monitoring the norm of the deficit vector, have a person decide to shop when

$$\sum_{i = 1}^{N} d(e[i], g[i]) > \gamma$$

where $d(e[i], g[i]) = g[i] - e[i]$ if $e[i] < g[i]$ and $d(e[i], g[i]) = 0$ otherwise. 

Note what this does: If $e[i]$ is below $g[i]$, then this will contribute to a person's desire to shop. However, $d(e[i], g[i])$ being zero when $e[i]$ exceeds $g[i]$ means that surpluses of one product don't deter them from going shopping for another that they are low on. Thus a person having 600 rolls of toilet paper doesn't stop them from needing to shop when their life-saving medication runs low. When a person shops, they should try to purchase to minimize this overall quantity, i.e. bring $\mathbf{e}$ as close to $\mathbf{g}$ as possible. 

How should $\mathbf{g}$ be determined? Let's have another constant, say $s$, representing how many hours worth of each commodity a person wishes to have at a given moment (obviously they will want to think in terms of days or weeks, but we choose to measure $s$ in hours). I believe we have something similar already which we are calling the shopping frequency. Here it precisely represents how much time a person wants to go in between shopping errands. Then for a person to stock $s$ hours of each good is to ensure that they possess exactly $s\mathbf{c}$ of it, where $\mathbf{c}$ is the consumption frequency bundle from earlier. Thus we have

$$ \mathbf{g} = s\mathbf{c} $$

I.e. $\mathbf{g}[i] = sc[i]$ for each $i$. I think this system works pretty well and is very clean, but let me know what you all think and tell me if there is anything about the above protocol that I can help clarify. $\gamma$ and $s$ can be determined randomly per person or they can be constants.

lennyyyyyyyy:
Resolves #134 
Resolves #137 
Resolves #144


Agent-Based Simulation/src/Society.cpp
```diff
@@ -189,7 +192,14 @@ void Society::set_product_prices_and_production() {
         }
         products[i]->price_per_unit = values(i);
     }
-
+    double consumption_scalar = 0.0;
+    for (Product * product : products) {
+        consumption_scalar += product->living_labor_per_unit * product->mean_consumption_frequency;
```
alexbcreiner0: First term on the right here should be the initial price per unit, i.e. the total socially necessary labor, not the living labor per unit.

Agent-Based Simulation/header/Constants.h
```diff
@@ -43,8 +44,9 @@ const int PRODUCT_NUM_MACHINES_MAX = STARTING_NUM_MACHINES / 5;
 const double PRODUCT_LABOR_PER_UNIT_MIN = 0.05;
 const double PRODUCT_LABOR_PER_UNIT_MAX = 0.1; 
 const double DISTRIBUTION_LABOR_PER_UNIT = 0.2;
-const double PRODUCT_CONSUMPTION_FREQUENCY_MIN = 0.05;
-const double PRODUCT_CONSUMPTION_FREQUENCY_MAX = 0.2;
+const int PRODUCT_CONSUMPTION_GROUP_SIZE = 5;
+const double PRODUCT_CONSUMPTION_FALLOFF = 0.9;
```
alexbcreiner0: I think this is too steep, should be closer to 0.97 I think to start.

Agent-Based Simulation/src/Person.cpp
```diff
@@ -102,39 +89,49 @@ float Person::get_current_productivity() {
...
+    }
 }
 
 bool Person::will_shop() {
-    return Sim::get_current_time_step() % PERSON_SHOPPING_PERIOD == shopping_offset;
+    double total_deficit = 0.0;
+    for (Product * product : society->get_goods()) {
+        total_deficit += std::max(0.0, 
+            PERSON_STOCKPILE_DURATION - 
+            inventory[product] / product->mean_consumption_frequency
```
alexbcreiner0, line 114: Okay, I think I see why you're doing this. It's not about normalizing units. It's about making it so that the decision to shop isn't 'unfairly' weighted by higher consumption frequencies, right? This way, all deficits are weighted equally. I like it.

Agent-Based Simulation/src/Person.cpp
```diff
@@ -102,39 +89,49 @@ float Person::get_current_productivity() {
...
+            (int) (PERSON_STOCKPILE_DURATION * product->mean_consumption_frequency) - 
+            inventory[product]
+        );
+        total_price += purchase_quantities[product] * 
+            society->get_consumer_good(product)->price_per_unit;
     }
-    log_shopping_deficit(std::max(0.0, (total_price - account) / total_price)); 
-    for (std::pair<Product *, double> p : ideal_purchase_quantities) {
-        int quantity = (int) (std::min(account / total_price, 1.0) * p.second);
+    double price_scalar = std::min(account / total_price, 1.0);
```
alexbcreiner0, line 131: This is fine for now. (Much) later on, we can maybe replace this logic with a utility maximization algorithm in which the person chooses an optimal bundle subject to their account constraint.

Agent-Based Simulation/header/Constants.h
```diff
@@ -43,8 +44,9 @@ const int PRODUCT_NUM_MACHINES_MAX = STARTING_NUM_MACHINES / 5;
...
+const double PRODUCT_CONSUMPTION_MULT = 0.25;
```
alexbcreiner0: I think that the way you are calculating the consumption frequencies, the dot product of the value vector with the consumption frequency vector will always be exactly this number. If that's the case, I think that this should be much higher, something like 0.95 or 0.97. At 0.25, that means that society is capable of spending 4/5 of every hour producing surplus. That's excessive.

Agent-Based Simulation/src/Producer.cpp
```diff
@@ -156,6 +157,7 @@ void Producer::move_plan_forward_one_step(Plan * plan) {
         }
     }
 	plan->total_hours_remaining -= labor_hours_done + raw_materials_used;
+    plan->quantity_remaining -= quantity_produced;
 }
```
alexbcreiner0, line 161: Overall on this function, I'm not sure why we are suddenly choosing to keep track of the quantity produced as our measure of when the plan is finished and not the total (socially necessary) hours remaining. As it stands, the way you are adjusting the quantity produced according to worker suitability and the price controller's reported living labor per unit requirements is fine. But I'm not seeing the labor time reporting reflect these adjustments at all. I think we should stick to keeping track of the plan's progress in terms of labor hours exclusively. Monitoring quantity produced is fine, but to replace our progress measure to progress in-kind as you are doing here is not something I was aware we were planning to do, and I think it is complicating things further. We should make sure that the issues we had before are gone before making such changes.

alexbcreiner0, line 161: I just checked the price controller changes and I _do think this all works as expected_. Initially, the plan's total labor hours LH and the plan's labor hours remaining LHR are the same. The plan is done when the quantity remaining reaches below zero (allowing for some partial overproduction which is fine). If this takes longer than expected, then LHR will go below zero, leading to LH - LHR > LH, indicating to the price controller that the time taken was longer than expected and accurate. So the price will adjust upward the way we want. So I think this is all fine for now, despite my initial complaint above. It's up to Amittai whether we should switch to using the quantity remaining as an indicator that the plan is done or the labor hours remaining as we did before.

Agent-Based Simulation/src/Producer.cpp
```diff
@@ -187,6 +189,13 @@ void Producer::end_plan(Plan * plan) {
     plan->order->customer->receive_shipment(plan->order);
     double total = plan->order->product->price_per_unit * plan->order->quantity;
     plan->prd += total;
+
+    // update local labor time
+    local_living_labor_per_unit[plan->order->product] = 
+        local_living_labor_per_unit[plan->order->product] * FIRM_LOCAL_LABOR_LERP + 
+        (plan->labor_hours - plan->labor_hours_remaining) * 
+        (1 - FIRM_LOCAL_LABOR_LERP) / plan->order->quantity;
```
alexbcreiner0: This is a little more complex than what I suggested, but I think it is okay and more realistic than my suggestion. Sure. It _does_ introduce yet another constant parameter though. We eventually have a big boring job to do of reviewing all of these constants and deciding which ones actually contribute to interesting emergent behaviors. We want our model to be as simple as possible, so I am tempted to argue that we simply update the local living labor based on what it was strictly during the last order. 

We need to always ask ourselves why we're adding some complication to our model: are we actually going to alter the fundamental qualitative results of the model by introducing this? If not, we shouldn't add it, because our model is a proof of concept for an economic system which has flexibility in its implementation.

Agent-Based Simulation/src/Society.cpp
```diff
@@ -189,7 +192,14 @@ void Society::set_product_prices_and_production() {
...
+    }
+    consumption_scalar = PRODUCT_CONSUMPTION_MULT / consumption_scalar;
+    for (Product * product : products) {
+        product->mean_consumption_frequency *= consumption_scalar;
+    }
```
alexbcreiner0, line 205: An oversight on my part: In my write up, my equation was that $\mathbf{\lambda} \cdot \mathbf{C} < N$. As Ivan pointed out in my office yesterday, this assumes implicitly that workers work every hour, because they certainly are consuming every hour (that's by definition of the $\mathbf{C}$ vector, which is an average. The RHS here needs to be modified to $\frac{T'}{T}\frac{D'}{D}N$ where $T'$ is the number of hours in the _working day_, and $D'$ is the number of days in the _working week_. I think that we need to adjust our equations to account for this working reality, or we will get overconsumption. Only after multiplying by these two ratios do we get the actual average number of labor hours worked by the population _per hour_.

alexbcreiner0 (changes requested):
I made a bunch of comments, but overall this is great work and only a few things actually need to be fixed before we can merge:

- The living labor needs reference in the consumption scalar calculation needs to be replaced with the price_per_unit, see my comment [here](https://github.com/Amittai-Aviram/BC-LTEWG/pull/151/changes#r2937020451).
- PRODUCT_CONSUMPTION_MULT should be multiplied by the ratio $\frac{T'}{24}\frac{D'}{7}$ where $T'$ is the number of hours in the working day and $D'$ is the number of days in the working week. (This is something that Ivan noticed in my office after the meeting, but Devin noticed too.)
- The new constants require some adjustment. PRODUCT_CONSUMPTION_MULT should be much higher, between 0.9 and 1. PRODUCT_CONSUMPTION_FALLOFF should also be higher, something closer to 1, like 0.97.

Agent-Based Simulation/src/Society.cpp
lennyyyyyyyy: Fixed.

Agent-Based Simulation/header/Constants.h
lennyyyyyyyy: Updated.

Agent-Based Simulation/header/Constants.h
lennyyyyyyyy: Updated.

Agent-Based Simulation/src/Producer.cpp
lennyyyyyyyy: Simplified.

Agent-Based Simulation/src/Society.cpp
lennyyyyyyyy: Fixed.

alexbcreiner0 (approved):

PR #154, opened 2026-03-21
Worker deallocation, updates to hours in the working day

### LINKED ISSUE #44: Improve efficiency of worker allocation handling (Devin)
Instead of storing workers in the society's and various firms' vectors, keep all of them in the society's vector, but add fields to the Person class to designate where that person is working, if at all. Then, worker allocation involves changing those fields.

Also, fix the math for predicting the total turnaround time after training when drafting a plan.

### LINKED ISSUE #145: Proposal for a worker reallocation system
Here is a proposed broad outline of a system which would handle the reassignment of workers from one firm to another. Suppose that orders for a particular product that Firm A specializes in are slowing down, while simultaneously orders for a product that Firm B specializes in are piling up faster than it can handle. Have Firm B, if demand exceeds its capacity for handling orders beyond a certain margin, send out a signal to other firms requesting a 'solidarity trip', i.e. a temporary reassignment of some workers from other firms to Firm B. Workers from any firm with the relevant skills and experience could choose to transfer themselves to Firm B, improving its capacity to fill orders. The temporary assignment can be permanent if the worker chooses to stay.

That's really it. I think for now, we can make the transfers permanent, but later we can think harder about incentives which would actually make these transfers go smoothly in a more realistic circumstance. The bottom line is that what we really need is 

1. A system for firms to determine if demand for their product exceeds their capacity by a certain threshold.
2. A system for broadcasting a request for more work to other firms.
3. A system for firms to determine if demand for their product is sufficiently low that they would be willing to send their workers elsewhere. 

This should be enough to see a dynamically division of labor which evolves according to need.

### LINKED ISSUE #163: Adjusting the length of the work day (Devin)
The measure of Society's average busyness gives us an way to tune the length of the working day. In a perfect world, we want the average busyness over the Society to be evenly spread amongst all workers, such that every Person is working all the time, for exactly the length of the work day. This situation looks like:

$$b_S=\frac{5\cdot h}{7\cdot 24}$$

where $$h$$ is the number of the hours in the work day. However, $$h$$ must be an integer, so we can't always have equality. Also, this describes a situation where there is no downtime or inefficiency, so there also can't be any time for workers to be on standby, which is unreasonable. So we introduce a constant for inefficiency ($$c_e\ge 1$$) and use:

$$b_S\cdot c_e\le\frac{5\cdot h}{7\cdot24}$$
$$h\ge\frac{b_S\cdot c_e\cdot7\cdot24}{5}$$

Periodically, we can set the new length of the work day to the smallest integer value of $$h$$ that satisfies the inequality. As mentioned in #145 , plans should also store the value of $$h$$ at the time they were created, and stick with that value for the duration of the plan, so as to not mess up the predicted vs. actual turnaround times when the global $$h$$ updates.

lennyyyyyyyy:
Resolve #163 
Resolve #145 
Resolve #44


Agent-Based Simulation/header/Constants.h
```diff
@@ -59,6 +59,9 @@ const unsigned int FIRM_DEMAND_WINDOW_MAX = MONTH;
 const double PRODUCER_INITIAL_INVENTORY_MULT = 10;
 const double DISTRIBUTOR_INITIAL_INVENTORY_MULT = 2;
 const double DEADLINE_SAFETY_MULT = 0.6;
+const double REJECTED_ORDER_DEMAND_EXCESS = 0.5;
+const double BUSYNESS_DEALLOCATION_THRESHOLD = 0.01;
```
Amittai-Aviram: These two names are pretty opaque. In general, the names of our constants seem to be getting more and more opaque — presumably because the logic that requires them is getting more complex. This is not good.

Agent-Based Simulation/src/Firm.cpp
```diff
@@ -84,23 +85,52 @@ Producer * Firm::send_order(Order * order) {
...
+    for (Producer * producer : primary_producers) {
         int draft_plan_time = producer->draft_plan(order);
-        if (draft_plan_time != DRAFT_ORDER_REJECTED &&
-                draft_plan_time < order_time) {
+        if (draft_plan_time == DRAFT_ORDER_REJECTED) {
+            producer->drop_order(order);
+            rejecting_primary_producers.push_back(producer);
+            continue;
+        }
+        if (draft_plan_time < order_time) {
```
Amittai-Aviram, line 145: I wrote a comment on this on Sal's recent PR (#147 ): it does not make sense.  `order_time` is initialized at 85 to `INT_MAX`, so, by definition, it must be true that `draft_plan_time <= order_time`. And if `draft_plan_time == order_time`, (a) that would be crazy, and (b) this would not tell you anything, i.e., it would not tell you what the intended turnaround time is. Instead, don't you want to compare the draft plan's turnaround time to the order's deadline?

Agent-Based Simulation/src/Firm.cpp
```diff
@@ -84,23 +85,52 @@ Producer * Firm::send_order(Order * order) {
...
             chosen_producer = producer;
         }
     }
+    if (!chosen_producer) {
+        for (Producer * producer : secondary_producers) {
+            int draft_plan_time = producer->draft_plan(order);
+            if (draft_plan_time == DRAFT_ORDER_REJECTED) {
+                producer->drop_order(order);
+                continue;
+            }
```
Amittai-Aviram, line 159: Code duplication — please refactor.

Agent-Based Simulation/src/Firm.cpp
```diff
@@ -187,6 +217,31 @@ void Firm::assign_workers(
...
 
+void Firm::onboard_worker(Person * worker) {
+    busyness[worker] = 0.0;
+    onboard_time[worker] = Sim::get_current_time_step();
+}
+
+void Firm::record_busyness(Person * worker, bool busy) {
+    double duration_proportion = 1.0 / DEALLOCATION_SAFETY_DURATION;
+    busyness[worker] = busyness[worker] * (1 - duration_proportion) + 
+        busy * duration_proportion;
```
Amittai-Aviram: To choose clarity over cleverness, maybe this should be two statements:
```
// ... double possible_idleness = busyness[worker] * duration_proportion;
busyness[worker] -= possible_idleness;
if (busy) {
    busyness[worker] += duration_proportion;
}
```
That said, this is still pretty opaque.

Amittai-Aviram (commented):
I really appreciate the thought, imagination, and work involved in this PR, but I'm afraid I am not quite "there" with it. The approach seems a little on the complicated side, and I'm not sure whether this is the best way to model the idea of workers being either tied up or available for reallocation. Intuitively, I tend to think that our modeling should be more closely connected to _plans_. The plan has the information we need about how many people are needed for how many hours. On the GIC model, the firm does not commit to paying workers a certain number of labor hour credits no matter what; it _does_ commit to paying workers for the hours set forth in the plan. Let's discuss this further.

Agent-Based Simulation/src/Firm.cpp
lennyyyyyyyy: `order_time` is just a running minimum, so we just set it as `INT_MAX` to start. Then we just use the running minimum to select the Producer that can accomplish the order the fastest. We don't have to check against the deadline because that information is given to Producers when they're drafting the plans. All of them try to meet the deadline, but if none of them can, then the downstream Producer has to make it work with the fastest Producer anyway,

Amittai-Aviram: OK, never mind. I did not see that `order_time` gets updated inside the loop that iterates over producers. This is one thing about GitHub's representation of changes that makes reviewing difficult. Sorry.

Amittai-Aviram (approved):
LGTM
This is great! Alas, you now have some merge conflicts, which may have arisen from my recent changes having gone into _main_.  The merge conflict (AFAIK) is only in src/Producer.cpp, and I think you just need to accept "HEAD" over "main" _except_ for `add_input_inventory`, which should be removed.

I have a few more questions about busyness that we should discuss in person.

PR #159, opened 2026-03-23
Set maximum number of inputs for a product reasonably

### LINKED ISSUE #149: Bug: the assignment of inputs to a product can cause the program to hang
If the number of inputs exceeds the overall number of products in the society, the `while`-loop in `Product::set\_inputs` will run forever.  This logic needs to be hardened to handle any number of products.

### LINKED ISSUE #158: Bug: Crash in `Society::set_initial_account`
`Society::set_initial_account` iterates over all `products`, prints an error message if there is no `ConsumerGood` for the current `Product *` — and continues.  Now, `products` includes machines, which are not sold as consumer goods, so, when the ensuing code touches `consumer_good->mean_consumption_frequency`, it dereferences NULL.

This bug was introduced in PR #142 .

Amittai-Aviram:
1. This change revises Devin's previous solution to the problem described in Issue #149 , in which the program would hang if the number of products was smaller than the number of inputs — so that the loop that populates the set of inputs would never finish. Instead of setting the maximum number of inputs to a fraction of the total number of products (e.g., 1/10), this change sets it to half, and then checks whether this would result in 0 inputs (i.e., if there is only 1 product), in which case we set it to the total number of products.  A maximum of half of all products seems reasonable, though it is theoretically possible for a real-world product to require even more inputs.
2. This change also fixes a bug introduced by PR #142  into `Society::set_initial_account`, where we iterate over all products instead of iterating only over the products designated as potential consumer goods (i.e., not machines). This bug could cause a crash, since there is no consumer good corresponding to a machine.

Resolves #149 
Resolves #158


Agent-Based Simulation/src/Product.cpp
```diff
@@ -33,8 +33,10 @@ Product::Product(int id, const std::string& name) :
 }
 
 void Product::set_inputs(std::vector<Product *>& goods) {
+    int max_num_inputs = PRODUCT_NUM_INPUTS_MAX ? PRODUCT_NUM_INPUTS_MAX :
```
lennyyyyyyyy: This works for STARTING_NUM_PRODUCTS = 1, but I'm kind of confused why its not instead:
`int max_num_inputs = std::max(PRODUCT_NUM_INPUTS_MIN, PRODUCT_NUM_INPUTS_MAX);`
`max_num_inputs = std::min(max_num_inputs, goods.size());`
, which accomplishes this in all cases.

Like, I could potentially see a scenario where we say "Products should always have, on average, 20 inputs, regardless of the total number of products." Then we might just set PRODUCT_NUM_INPUTS_MAX = 40, and then possibly vary STARTING_NUM_PRODUCTS to above or below 40, in which case this would break. 

If you're certain that we'll keep PRODUCT_NUM_INPUTS_MAX as a fractional proportion of STARTING_NUM_PRODUCTS, then this is fine.

lennyyyyyyyy (approved):
Approved, but it'd be good to look at my comment first.

Agent-Based Simulation/src/Product.cpp
Amittai-Aviram: Yeah, I actually thought of something like that (the max/min solution), but then discarded it. But, yeah, that is a more general solution.

Amittai-Aviram: Changed accordingly — see latest commit. Thanks!

PR #160, opened 2026-03-24
Clean Up and Shorten Functions in Firm and Producer

### LINKED ISSUE #152: Refactor Functions in Firm and Producer to Use Helper Functions
Some functions in `Firm.cpp` and `Producer.app` are too long and unreadable, therefore helper functions need to be made as a means of a higher level of abstraction that can be called in these methods to simplify the code. Additionally, the names of the helper functions need to be specific in order to be able to understand what exactly the helper function does.

This is a later priority issue BTW.

### LINKED ISSUE #161: Ensure That There is Only One Constructor
This is for Producer and Distributor (there are 2 constructors)

### LINKED ISSUE #162: Fix Initialization of Producers and Distributors
Producers should not have an initial input inventory consisting of their own products.

(Starts from Society in line 36)

Sal47506:
Resolves #152 

Resolves #162 

Resolves #161


Agent-Based Simulation/header/Firm.h
```diff
@@ -94,6 +94,15 @@ class Firm : public Agent {
...
 
     Producer * send_order(Order * order);
+    Producer * select_fastest_supplier_for_order(Order * order);
+    void pursue_order_with_selected_supplier(
+        Order * order,
+        Producer * chosen_producer
+    );
+    void drop_order_from_unselected_suppliers(
+        Order * order,
+        Producer * chosen_producer
```
lennyyyyyyyy: For consistency, the argument should be `unselected_supplier` or the method name should be `drop_order_from_unchosen_producer`. Same logic applies to the previous method.

Agent-Based Simulation/header/Firm.h
```diff
@@ -112,6 +121,11 @@ class Firm : public Agent {
     );
 	int predict_turnaround_time(Order * order, std::vector<Person*>& workers); 
 	int predict_labor_hours(Order * order, std::vector<Person*>& workers);
+    int calculate_raw_material_cost_for_order(Order * order);
+    void initialize_plan_progress_metrics(
+        Plan * draft_plan,
+        int raw_material_cost
```
lennyyyyyyyy: Does it have to be passed `raw_material_cost`? You can just call `calculate_raw_material_cost_for_order` on `draft_plan->order`.

Agent-Based Simulation/header/Producer.h
```diff
@@ -31,6 +31,23 @@ class Producer : public Firm {
 
   private:
     std::unordered_map<Order *, Plan *> order_to_draft_plan;
+    bool has_sufficient_inputs_for_order(const Order * order);
+    Plan * create_draft_plan_for_order(Order * order);
```
lennyyyyyyyy, line 34: I think it'd be better to directly change `draft_optimal_plan` in Firm, to avoid redundancy. There's only like 4 or 5 extra lines in this method anyway on top of the old one.

Agent-Based Simulation/src/Producer.cpp
```diff
@@ -195,6 +199,58 @@ void Producer::end_plan(Plan * plan) {
...
+        total_worker_suitability +=
+            worker->suitability(plan->order->product->required_abilities);
+    }
+    if (total_worker_suitability <= 0.0) {
+        return 0.0;
+    }
+    return total_worker_suitability / plan->order->product->living_labor_per_unit;
+}
+
+void Producer::register_worked_hour_for_plan_workers(Plan * plan) {
```
lennyyyyyyyy: Does this little really need to be pulled out?

Agent-Based Simulation/src/Producer.cpp
```diff
@@ -195,6 +199,58 @@ void Producer::end_plan(Plan * plan) {
...
+    plan->labor_hours_remaining -= labor_hours_done;
+    plan->raw_materials_remaining = std::max(
+        0.0,
+        plan->raw_materials_remaining - raw_materials_used
+    );
+    pooled_input_value_account = std::max(
+        0.0,
+        pooled_input_value_account - raw_materials_used
+    );
+    check_and_reorder_inputs();
```
lennyyyyyyyy: This doesn't make sense here, because the Plan doesn't deplete inputs as it's progressing. It consumes all the necessary inputs on `start_plan()`, so you just need to check once there.

Checking here defeats the purpose of getting rid of it in `on_time_step()`, because this is called on every time step for every Plan.

Agent-Based Simulation/src/Producer.cpp
```diff
@@ -195,6 +199,58 @@ void Producer::end_plan(Plan * plan) {
...
+bool Producer::is_within_work_schedule() const {
+    return Sim::get_current_time_step() % DAY <
+        Society::get_instance()->get_current_work_hours_daily() &&
+        Sim::get_current_time_step() / DAY % 7 <
+        static_cast<unsigned int>(
+            Society::get_instance()->get_current_work_days_weekly()
+        );
+}
+
+bool Producer::is_plan_finished(const Plan * plan) const {
```
lennyyyyyyyy: This seems unneccessary. If the statement is readable by itself, than the method isn't needed to clear the functionality up to the reader.

lennyyyyyyyy (dismissed):

Agent-Based Simulation/src/Producer.cpp
lennyyyyyyyy: I think you misunderstood my comment here, because it looks like you just dissolved the helper function back into the larger function. I liked how the helper function was pulled out, it was just that the `check_and_reorder_inputs` call was wrong.

Amittai-Aviram: I think `check_and_reorder_inputs` is gone from the predecessor to this code. Just please make sure not to _reintroduce_ it!

Agent-Based Simulation/header/Producer.h
Amittai-Aviram: Actually, `draft_optimal_plan` is only relevant to producers, so it should  be moved to `Producer`, in which case we do not need the wrapper function `create_draft_plan` at all.

Amittai-Aviram (approved):
Thank you for your hard work!
LGTM

PR #165, opened 2026-03-26
Call `Firm::check_and_reorder_inputs` only when an input inventory level can drop

### LINKED ISSUE #164: Call `Firm::check_and_reorder_inputs` only when an input inventory level drops
We are currently calling `Firm::check_and_reorder_inputs` in two places in the code base:
- In `Firm::on_time_step`
- In `Producer::move_plan_forward_one_step` — which, in turn, is called on every time step in which any plan is being executed.

Accordingly, every firm calls this function at least once on every time step, and every producer calls it at least twice on every time step, but possibly more times, if it has more than one plan in progress.  This is wasteful.

Instead, we should only call `check_and_reorder_inputs` when there is reason to believe that the inputs may need to be checked and reordered, which is when input products are drawn out of the inventory. This happens, specifically, in two places:
1. In  `Producer::start_plan`, right after the loop that calls `Firm::remove_input_inventory` for each input to make the current plan's product.
2. In `Distributor::try_sell_goods` (or perhaps `Distributor` should call `remove_input_inventory` in `try_sell_goods` to create a common pathway for all kinds of firm).

In addition, `remove_input_inventory` should be renamed to `remove_input_from_inventory`. There is also a `Firm::add_input_inventory`, but this is apparently dead code and should be removed.

Amittai-Aviram:
With this change, instead of checking the input inventory on every time step for every firm _and again_ for every plan of every producer, we check, and reorder if necessary, only
- When a producer removes products from its input inventory at the start fo a plan
- When a customer buys a product from a distributor.
In the latter case, we only check the level (and possibly reorder) for the product in question.

Resolves #164

### DIFF
--- a/Agent-Based Simulation/header/Firm.h
+++ b/Agent-Based Simulation/header/Firm.h
@@ -103,6 +103,7 @@ class Firm : public Agent {
         int pending_inventory
     );
     void check_and_reorder_inputs();
+    void check_and_reorder_input(Product * product);
 
 	int predict_workers_needed(Order * order);
     void assign_workers(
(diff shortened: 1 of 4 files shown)


ghost:
LGTM

Sal47506 (approved):
LGTM

PR #168, opened 2026-04-04
Expand logging and parameter-setting capabilities

### LINKED ISSUE #146: Additional logging requests, parameters as options when running sim, and other things
A variety of additional logging is needed for my Python application to track what we need to track. Also several other small things. A lot of this I've already fixed myself in my personal branch. Those which I've already implemented will be marked checked.

# Logger
- [x] The I/O matrix and the living labor vector need to be logged at the start of the simulation. I need this for a variety of different reasons.
- [x] Firms should log their initial inventories. I could theoretically infer this from the above information, but it would be more robust if I did not have to, in case our decisions for initial inventories changes later down the line and we forget that I was doing that. 
- [x] Firms should log when they deduct from their inventories for any reason. This includes both goods removed for pursuing a new plan, as well as goods removed when selling to a Person. Without this, it is currently very difficult to track individual firm inventories (i.e. it is only possible to track overall supply). 
- [x] Persons need to have assigned firms as a property. When their assigned firm changes from a null pointer to a firm, this should be logged. When it changes from a firm to a null pointer, this should also be logged. Without this, I can't track employment numbers.
- [x]  Total initial employment should be logged. Again, with the assumptions our model makes right now, this can be inferred (everyone is initially unemployed), but this inference is dangerous and unnecessary in the case that our scenarios change.
- [x]  My Python wrapper should also not have to infer whether a firm is a producer or distributor from their id alone. Right now I can simply check if their id is greater than or equal to the number of products, but this is a scenario based assumption which is subject to change. I don't have a specific preference as to how this is done, but in general whenever a firm logs anything, it should easy to infer from the log message alone whether they are a distributor or a producer. (I need this because I can already see the importance of distinguishing between accessible supply (i.e. products which are actually available for purchase by citizens, and total supply, which might be sitting fallow in a producer's inventory.)
- [x]  Persons should log their initial consumption frequencies (assuming that these are personalized when those changes are made), their initial inventories (whenever that gets implemented), and their initial skill sets.
- [x]  When their skills change or improve, this should be logged.
- [ ]  New demand signals should be logged, so that demand can be tracked as well as supply. This might induce too much slowdown because one demand signal goes to every producer at once. To make sure this doesn't cause problems later when we have multiple firms producing a product, we should have one producer of every product dynamically marked as the 'designated demand logger' and have just that firm log its demand signals. That is enough to infer demand for everyone.
- [x]  I need a log_shopped function in addition to log_shopping which only gets called when shopping is successful and goods actually transfer. Otherwise, it is impossible to track firm inventories. I've already implemented this one myself. I've done so by altering the charge function so that it takes three inputs - a product_name, a quantity, and a cost, and then logs those things together after their account has been decremented. I needed a log signal which would give me all three of these things together - having to combine log messages together in order to infer things like this is too difficult and subject to error. This also required me to create a new overloaded log function which takes a string name, an int quantity, and another double cost.

# Constants and parameters
- [x] Right now we can add the -j flag when we run sim to specify how the logger behaves. This needs to be extended to allow for the execution of the simulation with different parameter settings. We are not doing a good job distinguishing in our Constants.h file between what the actual independent parameters are characterizing the system and what are the dependent variables. A minimal set of parameters which I think we should have as arguments are the following: NUM_SIM_RUNS, STARTING_NUM_PEOPLE, INITIAL_WORK_HOURS_DAILY, INITIAL_WORK_HOURS_WEEKLY, STARTING_NUM_PRODUCTS, STARTING_NUM_MACHINES (not the actual number, but the divisor which bases this on the number of products, currently 5), STARTING_NUM_PRODUCERS, STARTING_NUM_DISTRIBUTORS, NUM_ABILITIES (this isn't currently a constant in Constants.h but it absolutely should be considered a parameter), and PRODUCT_INPUT_EPSILON.
    - If anyone adds even just one of these as an argument, I can probably just copy the pattern and implement the others myself. 
    - More thought is probably necessary about what the essential parameters of our model actually are. Ideally, a conceptual ABM should have as few parameters characterizing it as possible. 
- Our simulation is not stable with respect to changes to the constants in our Constant.h file.
    - [x] If STARTING_NUM_PEOPLE is close to STARTING_NUM_PRODUCERS + STARTING_NUM_DISTRIBUTORS, then the floored division we do in order to determine average team size is zero. This should be clamped to at least one. I don't know for sure if this causes problems, but should still be done.
    - [x] Right now, PRODUCT_NUM_INPUTS_MIN and PRODUCT_NUM_INPUTS_MAX are independent of the STARTING_NUM_PRODUCTS. This creates possibly infinite loops in the products set_inputs and set_machines functions, because indices.size can never exceed the number of inputs. In my own branch I've fixed this by setting the number of inputs to the minimum of what the distribution actually yields and the number of products, but the more robust solution would be to change these constants to actually depend on the STARTING_NUM_PRODUCTS. This is what I mean about the importance of minimizing the number of independent parameters we have.

### LINKED ISSUE #148: Support logging inventory levels
Logging inventories presents a problem with respect to the current logging API, because inventories really require an additional nesting level of keys and values.  Currently, every logging entry has a client (the particular class of agent), an id (distinguishing the individual instance of the client or agent class), a label (saying what is being logged), and a time step, as well as a corresponding single value or pair or triple of values. Thus, in effect, we have four nested levels of keys pointing to the corresponding value or values:

client -> id -> label -> time step -> values

But, in inventories, we need the individual product stored in the inventory to serve as an additional key. A typical entry should look like this:

Firm -> 3 -> "input_inventory" -> 57 -> "Product 7" -> 31

But this introduces a more general problem of the appropriate data structure to use to represent this sort of logging function — i.e., it is a function from an arbitrary ordered set (tuple) of values serving as the key to an ordered set of values per se.

Amittai-Aviram:
This change
- Allows the user to set several key parameters when running simulation, such as the number of people in the society, the number of products, etc.
- Logs changes in inventory levels and several other events useful for tracking the system.

Resolves #146 
Resolves #148


Agent-Based Simulation/src/Society.cpp
```diff
@@ -290,3 +313,24 @@ Person * Society::birth_person() {
 void Society::retire_person(Person *person) {
     // unimplemented until hiring/reallocation is done
 }
+
+void Society::log_io_matrix(Eigen::MatrixXd& A, size_t dim) {
+    Logger::get_instance()->log(Logger::SOCIETY, "A_dim", id, static_cast<int>(dim));
+    for (size_t i = 0; i < dim; ++i) {
+        for (size_t j = 0; j < dim; ++j) {
+            if (A(i, j)) {
+                std::string index = std::to_string(i) + "," + std::to_string(j);
```
alexbcreiner0: Right now the logged outputs here look like 

`{"t":0,"client":"Society","id":0,"label":"A","0,25":0.0586717} `

Can it instead be something like
`{"t":0,"client":"Society","id":0,"label":"A","coord":[0,25],"value":0.0586717} `?

or alternatively

`{"t":0,"client":"Society","id":0,"label":"A","value":[(0,25), 0.0586717]} `?

I need to be able to easily access the coordinate via hashing. Right now, I have to search for an item that looks like a coordinate-value pair, which will slow everything down.

Amittai-Aviram: Done:
{"t":0,"client":"Society","id":0,"label":"A","coords":[0,1],"value":0.0244905}
Note that I have "coords" rather than "coord."
Did you need this format for the labor vector as well?

Agent-Based Simulation/src/main.cpp
```diff
@@ -5,49 +5,89 @@
...
-                            print_usage();
-                            exit(EXIT_SUCCESS);
+            switch (argv[i][j]) {
+                case 'n':
+                case 'p':
+                case 'h':
+                case 'w':
+                case 'o':
+                case 'r':
+                case 'd':
```
alexbcreiner0, line 34: I think case 'm' is missing here

Agent-Based Simulation/src/Society.cpp
alexbcreiner0: It's not as important, but that would be nice.

Agent-Based Simulation/src/Firm.cpp
```diff
@@ -400,7 +400,7 @@ void Firm::log_shipment_received(std::string product_name, int quantity) {
 
 void Firm::log_inventory_level(std::string product_name, int quantity) {
     Logger::get_instance()->log(
-            Logger::FIRM,
+            get_client_type(),
             "inventory_level",
             id,
             product_name,
```
alexbcreiner0: Same thing as an earlier comment here. Currently it looks like

`{'t': 0, 'client': 'producer', 'id': 0, 'label': 'inventory_level', 'product_2': 10639}`

It would be a lot better if it were something like

`{'t': 0, 'client': 'producer', 'id': 0, 'label': 'inventory_level', 'product_id': 2, 'amount': 10639}`

So that it's easier to find the relevant values.

Agent-Based Simulation/src/Person.cpp
```diff
@@ -214,9 +220,22 @@ void Person::log_shopping_deficit(const double deficit) {
...
 
-void Person::log_state() {
-    Logger::get_instance()->log(Logger::PERSON, "age", id, age);
-    Logger::get_instance()->log(Logger::PERSON, "account", id, account);
-    Logger::get_instance()->log(Logger::PERSON, "health_status", id, health_status);
+void Person::log_placement() {
+    Logger::get_instance()->log(Logger::PERSON, "firm", id, firm ? static_cast<int>(firm->get_id()) : -1);
+}
+
+void Person::log_abilities() {
```
alexbcreiner0, line 240: Same request as my prior comments. Instead of

`{'t': 0, 'client': 'Person', 'id': 29, 'label': 'ability', 'Ability_3': 0.971524}`

it would be better to have

`{'t': 0, 'client': 'Person', 'id': 29, 'label': 'ability', 'ability': 3,  'value': 0.971524}`

Agent-Based Simulation/src/Person.cpp
```diff
@@ -214,9 +220,22 @@ void Person::log_shopping_deficit(const double deficit) {
...
+    for (std::pair<Ability, double> ability : abilities) {
+        Logger::get_instance()->log(Logger::PERSON,"ability", id, ability_names[ability.first], ability.second);
+    }
 }
 
+void Person::log_inventory() {
```
alexbcreiner0, line 253: Same request as my log_inventory request for firms

Amittai-Aviram:
@alexbcreiner0 Note that I already inserted a logging call for whenever a person changes firms: `Person::set_firm` calls `Person::log_placement`. Note that this logging call records a nonexistent (NULL) firm as -1. (Right now, I don't think we'd ever see that.)

Agent-Based Simulation/src/Person.cpp
```diff
@@ -202,21 +216,46 @@ double Person::suitability(std::vector<Ability>& required_abilities) {
...
+    }
+}
+
+void Person::log_account() {
     Logger::get_instance()->log(Logger::PERSON, "account", id, account);
-    Logger::get_instance()->log(Logger::PERSON, "health_status", id, health_status);
 }
 
+void Person::log_health_status() {
+    Logger::get_instance()->log(Logger::PERSON, "health_status", id, health_status_names[health_status]);
```
alexbcreiner0: This creates logs of the form

`{"t":44,"client":"Person","id":22,"label":"health_status","values":[Unhealthy]}`

That's not valid json though, and leads to decoder errors. There need to be quotes wrapped around the health status, i.e.

``{"t":44,"client":"Person","id":22,"label":"health_status","values":["Unhealthy"]}``

alexbcreiner0 (changes requested):
Things I still need (recapping what was said in our signal chat):
- Producers should log deductions from their inventory, either when plans are pursued or in the course of a plan being carried out.
- Consumption frequencies should be logged at initialization in the same way that A and l currently are
- Distributors need to log deductions from inventory when consumers make purchases
- Assuming the above, _firms do not need to report inventories when they receive shipments_. I should be able to infer inventory from keeping track of individual micro-events like orders received and transfers made between agents. 
- Changes in employment need to be logged. (Still not seeing anything there). 
 
Person health reporting creates a json parsing error. Currently messages look like 
`{"t":44,"client":"Person","id":22,"label":"health_status","values":[Unhealthy]}
`
Unhealthy needs to be wrapped in quotes, so it looks like
`{"t":44,"client":"Person","id":22,"label":"health_status","values":[Unhealthy]}
`

Agent-Based Simulation/src/Society.cpp
Amittai-Aviram: Done:
`{"t":0,"client":"Society","id":0,"label":"l","index":1,"value":0.0815339}`
Note use of "index" rather than "coords."
If you are satisfied with this, please mark this conversation as "resolved." Thanks!

Amittai-Aviram:
> * Producers should log deductions from their inventory, either when plans are pursued or in the course of a plan being carried out.

Done.

> * Consumption frequencies should be logged at initialization in the same way that A and l currently are

Done.

> * Distributors need to log deductions from inventory when consumers make purchases

Done.

> * Assuming the above, _firms do not need to report inventories when they receive shipments_. I should be able to infer inventory from keeping track of individual micro-events like orders received and transfers made between agents.

The inventory logging call comes right after the shipment received logging call, and is specific to the product received. We can take this out once we are sure that everything is working correctly.

> * Changes in employment need to be logged. (Still not seeing anything there).

As far as I can tell, this is because `Person::set_firm` never gets called. (I tested this by itself over 1000 runs.) This is probably related to the bug Devin and I discussed at yesterday's meeting. Devin is on it, so I think we shall see different behavior once his bug fix is in.

> 
> Person health reporting creates a json parsing error. Currently messages look like `{"t":44,"client":"Person","id":22,"label":"health_status","values":[Unhealthy]} ` Unhealthy needs to be wrapped in quotes, so it looks like `{"t":44,"client":"Person","id":22,"label":"health_status","values":[Unhealthy]} `

Done:
`{"t":0,"client":"Person","id":896,"label":"health_status","values":["Unhealthy"]}`

Agent-Based Simulation/src/main.cpp
Amittai-Aviram: Fixed.

Agent-Based Simulation/src/Firm.cpp
Amittai-Aviram: Done.
I did the same for shipments received, inventory reductions (on plan start), reorders, and recorder failures (where a producer cannot be found).
Also for `Person::log_inventory`.

Agent-Based Simulation/src/Person.cpp
Amittai-Aviram: Done:
`{"t":0,"client":"Person","id":0,"label":"ability","ability":2,"value":0}`

Agent-Based Simulation/src/Person.cpp
Amittai-Aviram: Done:
`{"t":0,"client":"Person","id":0,"label":"inventory","product_id":98,"amount":0}`

Agent-Based Simulation/src/Person.cpp
Amittai-Aviram: Fixed.

alexbcreiner0 (approved):

PR #169, opened 2026-04-05
Trade Interface Improved Step 1 

Kkenize:
I created a new project for the improved trade interface with a live ce instance created from the event based simulation. Step 1 of the new issue is being done this week. Next week I will perhaps work on step 2. 

This PR refers to Issue #167 .


Trade Interface Improved/README.md
```diff
@@ -0,0 +1,197 @@
...
+
+### Simulation Design
+
+- Constructs a multi-commodity capitalist economy using a generated input–output matrix and labor coefficients
+- Normalizes technical coefficients and the labor vector to ensure internally consistent production costs
+- Initializes prices, supplies, outputs, wages, profits, and inventories from parameterized initial conditions
+- Evolves the economy forward in discrete time by repeatedly applying a single event-based update step
+- Records trajectories of prices, supplies, and related state variables at each time step
+- Computes total reproduction demand at each time step as:
+  - Intermediate production demand (A · q)
```
Amittai-Aviram: Maybe just write $Aq$ (matrix-vector multiplication) rather than $A \dot q$ (impossible dot product).

Trade Interface Improved/README.md
```diff
@@ -0,0 +1,197 @@
...
+  - Worker consumption demand
+  - Capitalist consumption demand
+  - Total reproduction requirement = A·q + b + c
```
Amittai-Aviram: See above. $Aq$ is a (column) vector. Each element $i$ of $Aq$ is the dot product of row $i$ of $A$ with all of $q$. It represents a sequence of transformations of the $n$-dimensional "point" $q$ into a different $n$-dimensional space. More specifically here, $Aq$ is what we need to produce of each product in order to have our $q$ products, given that any of those products may be inputs for the production of other products. $b$ and $c$ are vectors of the same size as $q$ and (therefore) $Aq$, so the addition makes sense.

Amittai-Aviram (approved):
Looks good so far. Please see my few comments and revise before merging into main.

Issue #170, opened 2026-04-05
Bugfixes (Devin)


lennyyyyyyyy:
### Non-bug changes:
- After reverting the shared general inventory from an output inventory to an inventory, Distributor's `get_inventory()` method override is redundant of Firm's `get_inventory()`.
- Constants `PRODUCTION_THRESHOLD`, `RAW_MATERIAL_THRESHOLD`, `RAW_MATERIAL_SURPLUS_FACTOR`, `RAW_MATERIAL_ORDER_MULTIPLIER` are completely unused. In particular, the last three were used in the redundant reordering logic when inventories were changed.
- Person consumption is completely deterministic, causing everyone to shop and consume the same amounts at the same time. Varying the initial inventory level causes consumption and demand to be less spiky.
- Created a system for Firms to attempt smaller but faster orders when its larger orders are universally rejected, as well as Producers to limit larger orders that deplete too much of its inventory at once (see #176)
### Bugs fixed so far:
- `drop_orders_with_unchosen_producers()` drops orders from all producers that aren't the chosen producer, regardless of whether or not it even drafted a plan. Instead, dropping should occur at the same time as sending out draft requests.
- ConsumerGoods didn't adopt their respective products' `mean_consumption_frequency` (defaulting to zero), causing the predicted `initial_account` for people in the society to be zero, since they expected never purchasing the products.
- The PriceController didn't update the price of the ConsumerGood in updating the price of the base product, so the ConsumerGood's price ends up taking a constant value.
- Unsigned integer overflow when subtracting from the current time step to find the minimum time step allowed in the demand signal window, which caused all demand signals to be popped a the beginning of time.
- Products' `societal_living_labor_per_unit` needs to be set equal to the ideal `living_labor_per_unit`, because there are no finished plans for the PriceController to analyze, causing it to default to zero and the predicted turnaround time of the first drafted plan to be zero.
### Other notes:
- Distributors have a distribution plan, but no hiring process and thus no workers assigned to the plan. However, they tack on the `DISTRIBUTION_LABOR_PER_UNIT` labor cost to the good, but no one gets paid to make up for the cost, which is a leak in the accounting. For testing, I'm just setting this to zero for now.
- For `INEFFICIENCY_OF_WORK`, is very hard to find a constant value that works for a variety of initial conditions, which makes sense, because its dependent of the specifics of the society. See #177.
[...]

PR #175, opened 2026-04-07
Fix plan drafting and worker assignment for producer shipments

### LINKED ISSUE #174: Bug: only one distributor receives shipments
1. No producers are logging receipt of shipments, suggesting that no downstream producers are receiving any shipments.
2. Only one distributor (with id 8) receives shipments, and only for Product 3, with various (small) quantities.

The command to reproduce this issue is as follows:
`% bin/sim -n 1000 -h 8 -w 5 -o 5 -m 1 -r 5 -d 5 -p 30 -j > output.txt`

Note: @Sal47506 Please do not use any formatting settings in VSCode. It is messing up our formatting, e.g., forcing `Product * product` to become `Product *product`, which violates our coding style.

Sal47506:
- Initialize `global_living_labor_per_unit` from `living_labor_per_unit` when each `Product` is created, so early plan drafting does not depend on an uninitialized value.
- Initialize `draft_plan->local_work_hours_daily` before worker-count and turnaround predictions, so producers can draft valid plans for all products.
- When a producer pursues a plan, assign selected workers through `move_worker_off_standby()` so they are actually moved out of unemployment/standby and attached to the firm.
- Use the sorted standby-worker list in `assign_workers()` instead of iterating the unsorted set.

Resolves #174


lennyyyyyyyy (dismissed):
Please don't include the output and trace in the pull request. I'd also note that all of the changes here are changes I've already mentioned in bugfixes (see #170). 

Also, it seems like you force-pushed to the main branch? The change looks small, but why didn't you just include that in the PR?

Agent-Based Simulation/header/Logger.h
```diff
@@ -6,6 +6,7 @@
 #include <tuple>
 #include <variant>
 #include <unordered_map>
+#include <iostream>
```
Amittai-Aviram: Please move this to preserve alphabetical order in system headers. (BTW, project headers should be listed below these, also in alphabetical order, with a vertical space between the two groups.)

Amittai-Aviram (changes requested):
Thanks for these changes, Sal. Please fix the `#include` issue and also please _remove_ output.txt and trace.txt from the repo:
```
git rm output.txt
git rm trace.txt
git commit -am "Remove debug artifacts from repo"
git push
```
Thanks!

Amittai-Aviram:
I do not understand the difference between `Product::living_labor_per_unit` and `Product::global_living_labor_per_unit`.  What is the global number and why are these two distinct?

Amittai-Aviram (approved):
LGTM

PR #178, opened 2026-04-07
Bugfixes, make reordering and order rejection smarter

### LINKED ISSUE #170: Bugfixes (Devin)

### LINKED ISSUE #176: Making reorders and plan rejection smarter (Devin)
In my testing, I've found that the naiveness of the reordering system, in particular, trying to find a Producer that has enough inputs to produce a quantity equal to the Firm's entire input threshold, locks up the system when upstream Producers don't have enough inputs to do so. While this reordering system has the correct quantities and deadlines to match demand (which is the threshold quantity, and the predicted time for the threshold quantity to be depleted), this certainly isn't the only combination of reorder quantities and deadlines to match demand. 

In fact, this quantity + deadline combination is just the _maximum_ possible combination, but any scaled fraction of this quantity + deadline will match demand equivalently, just for a shorter amount of time. I propose sending out orders _starting from the maximum quantity and deadline_, then repeatedly scaling this amount down until a Producer is found with the inputs to carry out the order (up until the rounded quantity or deadline is zero, which would make no sense.)

On a similar topic, plan rejection is also too lenient. Specifically, upstream Producers that get particularly large orders accept them even though it may deplete their input inventories to values very close to zero. This triggers a reorder for a large amount with a very short deadline, which no other Producer can feasibly accept. The change mentioned above, in combination with a rejection condition as a _proportion_ of the current inventory, motivates orders to be smaller, but more frequent and more continuous. 

After implementing these changes, I've found that the demand signals are a lot more accurate (getting a large order all of a sudden causes an overestimate of demand), and that not being able to find a Producer is often no longer due to a lack of inputs, but a lack of workers.

lennyyyyyyyy:
Resolve #170 
Resolve #176


Agent-Based Simulation/header/Constants.h
```diff
@@ -11,22 +11,25 @@ const unsigned int YEAR = DAY * 365;
...
 const double INEFFICIENCY_OF_WORK = 1.5;
 
-const unsigned int STARTING_NUM_PEOPLE = 1000;
-const unsigned int STARTING_NUM_PRODUCTS = 100;
-const unsigned int STARTING_PRODUCTS_PER_MACHINE = 5;
-const unsigned int STARTING_NUM_PRODUCERS = STARTING_NUM_PRODUCTS;
-const unsigned int STARTING_NUM_DISTRIBUTORS = STARTING_NUM_PRODUCTS;
+const int STARTING_NUM_PEOPLE = 1000;
+const int STARTING_NUM_PRODUCTS = 100;
+const int STARTING_NUM_MACHINES = STARTING_NUM_PRODUCTS / 5;
```
alexbcreiner0: I think this is outdated? Currently I have products_per_machine= 5 on main. This number should be independent of the number of products now.

Agent-Based Simulation/src/Person.cpp
```diff
@@ -101,24 +99,19 @@ float Person::productivity() {
...
     for (Product * product : society->get_goods()) {
         to_consume[product] += product->mean_consumption_frequency;
-        int consumed = static_cast<int>(to_consume[product]);
-        if (consumed) {
-            inventory[product] -= consumed;
-            log_consumption(product, consumed);
-        }
+        inventory[product] -= (int) to_consume[product];
+        inventory[product] = std::max(0, inventory[product]);
         to_consume[product] -= (int) to_consume[product];
```
alexbcreiner0: I think we need to discuss this. It might be a mistake to have consumption 'debt' accumulating for everything.

alexbcreiner0:
Could you please merge main into this asap? I'd like to evaluate your bug fixes using my software, which requires the logging features that were added this weekend.

Agent-Based Simulation/src/Distributor.cpp
```diff
@@ -58,39 +58,31 @@ void Distributor::on_time_step() {
...
             << " units of " << product.product_name << " costing " 
             << cost << std::endl;
-        return false;
+        return 0;
     } 
     Plan * plan = product_to_plan[&product];
-    plan->outgoing_units_consumed += quantity;
+    plan->outgoing_units_consumed += available;
     plan->prd += cost;
-    remove_input_from_inventory(&product, quantity);
```
alexbcreiner0, line 91: Why is this removed? This is where the distributor logs the sale. Without that I can't track supply. If you have a reason to remove this, you need to put that logging call somewhere else here.

alexbcreiner0 (changes requested):
These changes are causing a LOT of crashes compared to before, from my testing so far. Very hard to get a successful run past 1000 cycles. Also sim seems to fail to run at all if the number of products, producers and distributors is 5 (10 seems to work, those are the only two settings I've tried).

Still in the middle of this, won't be finished reviewing tonight.

Issue #179, opened 2026-04-08
People are starting with WAY too many labor credits, and are shopping way too frequently.


alexbcreiner0:
Thanks to Amittai's hard work this weekend and Sal's employment bug fixes, I've got some good visuals going of how our model is doing. Here is where things look when simulating a 5 commodity economy with a 5 day 8 hour work week, 1 machine, 5 producers, 5 distributors and 30 people, simulated for 20000 hours (which is two years and some change). The simulation does not always make it to 20000 hours. At least half the time it stops at some point along that time frame for reasons that will require some investigation. The behavior I've seen so far is very rich and can vary a lot. The particular dependency graph which was randomly generated for this simulation is here.
<img width="1152" height="890" alt="Image" src="https://github.com/user-attachments/assets/c7f7f908-eb52-4c4a-9b0b-681a3285380e" />

Here we have supply, income and consumption during that time:
<img width="1277" height="903" alt="Image" src="https://github.com/user-attachments/assets/fc082232-b9f2-4110-bdb8-4c85bc5ec813" />

Note that the (1,1) 'average i possessed' plot only goes for 2000 cycles, to make the initial behavior more clearly visible. Aside from the fact that they are all consuming at the same frequencies (which another issue currently has discussion on), they are just also very clearly shopping way too often. Every 10 hours it looks like. The existence of a product that people have to restock every 10 hours isn't necessarily a problem to me (we want as many wacky possibilities as we can have), but what IS a problem is that there is no variance here. It's always a window of time like this every time I run the simulation. There needs to be a wide range of possible consumption frequencies. We should have products that need to be purchased often, and others which can last a year. Again, this is why I stressed that we should draw these periods uniformly over a very long interval, to ensure a wide variance in commodity types and consumption habits.

Also steering the weird shopping frequencies are are that the PERSON_STOCKPILE_DURATION constant is not getting interpreted properly (I see that its set to WEEK), but for some reason people are only stockpiling for 60 hours. Clearly the 'floor' where they detect a shortage and go shopping is also too low, but it might not be once we fix the ceiling.

[...]

PR #183, opened 2026-04-19
Fixed excess labor credits, price controller inflation, more varied consumption frequencies, and more.

alexbcreiner0:
Resolves #[182](github.com/BC-LTEWG/Labor-Time-Economy-Simulation/issues/182)
Resolves #[179](https://github.com/BC-LTEWG/Labor-Time-Economy-Simulation/issues/179) 
(Specifically, resolves the consumption frequency issues, the excess starting labor credits issue, and the price controller issues described there.)

Adds logging for catalogs, sectoral employment, busyness, among other things.
Changed worker skills to be determined via a lognormal distribution.
Adds optional arguments for controlling the standard deviation of worker skills and proportion of the population that gets sick per year. Revised and renamed constants associated with those systems.

I might have forgotten something.


Amittai-Aviram:
It seems that, despite .gitignore, your PR wound up including a whole lot of build artifacts, which is why the PR includes so many files (33). Could you please fix this?

header/Constants.h
```diff
@@ -32,30 +32,31 @@ const double AVERAGE_RETIREMENT_AGE = (
 
 const int PRODUCT_ABILITY_COUNT_MAX = 1;
 const int PRODUCT_ORDER_SIZE_MIN = 1;
-const int PRODUCT_ORDER_SIZE_MAX = 10;
+const int PRODUCT_ORDER_SIZE_MAX = 1000;
```
Amittai-Aviram: How does this fit into the scope of this PR?

header/Constants.h
```diff
@@ -32,30 +32,31 @@ const double AVERAGE_RETIREMENT_AGE = (
...
 const int PRODUCT_NUM_INPUTS_MIN = 1;
 const int PRODUCT_NUM_INPUTS_MAX = STARTING_NUM_PRODUCTS / 2;
-const double PRODUCT_INPUT_PER_UNIT_MIN = 0.1;
+const double PRODUCT_INPUT_PER_UNIT_MIN = 0.1; // these constraints are unnnecessessarily narrow
```
Amittai-Aviram: Could you please remove this comment and, instead, write up an issue?

header/Constants.h
```diff
@@ -32,30 +32,31 @@ const double AVERAGE_RETIREMENT_AGE = (
...
 const double PRODUCT_INPUT_PER_UNIT_MAX = 0.25;
+const int MAX_CONSUMPTION_PERIOD = 100;
 const double PRODUCT_INPUT_EPSILON = 0.0001;
 const int PRODUCT_NUM_MACHINES_MIN = 0;
 const int MAX_PROPORTION_OF_MACHINES_PER_PRODUCT = 5;
-const double PRODUCT_LABOR_PER_UNIT_MIN = 0.05;
+const double PRODUCT_LABOR_PER_UNIT_MIN = 0.05; // also way too narrow
```
Amittai-Aviram: Please see comment above.

header/Constants.h
```diff
@@ -32,30 +32,31 @@ const double AVERAGE_RETIREMENT_AGE = (
...
 const double PRODUCT_LABOR_PER_UNIT_MAX = 0.1; 
 const double DISTRIBUTION_LABOR_PER_UNIT = 0.2;
-const int PRODUCT_CONSUMPTION_GROUP_SIZE = 5;
+const int PRODUCT_CONSUMPTION_GROUP_SIZE = 1;
```
Amittai-Aviram: Do we still need this constant at all?

header/Constants.h
```diff
@@ -32,30 +32,31 @@ const double AVERAGE_RETIREMENT_AGE = (
...
 const double PRODUCT_CONSUMPTION_FALLOFF = 0.97;
 const double PRODUCT_CONSUMPTION_MULT = 0.95;
-const int PRICE_AVERAGING_WINDOW = MONTH;
+const int PRICE_AVERAGING_WINDOW = 2*MONTH;
```
Amittai-Aviram: Style: please rewrite as 2 * MONTH;
(This is a general style rule for arithmetic, logical, relational, bitwise, and assignment binary operators.)

header/Constants.h
```diff
@@ -32,30 +32,31 @@ const double AVERAGE_RETIREMENT_AGE = (
...
 const int MACHINE_LIFETIME_MIN = 1000;
 const int MACHINE_LIFETIME_MAX = MACHINE_LIFETIME_MIN * 10;
 
 const int FIRM_STOCKPILE_DURATION = 1.5 * WEEK;
 const unsigned int FIRM_DEMAND_WINDOW_MIN = WEEK;
 const unsigned int FIRM_DEMAND_WINDOW_MAX = MONTH;
-const double PRODUCER_INITIAL_INVENTORY_MULT = 10;
-const double DISTRIBUTOR_INITIAL_INVENTORY_MULT = 2;
+const double PRODUCER_INITIAL_INVENTORY_MULT = 1;
+const double DISTRIBUTOR_INITIAL_INVENTORY_MULT = 1;
```
Amittai-Aviram: Likewise — for this and PRODUCER_INITIAL_INVENTORY_MULT, do we still need this constant at all? If we shall be multiplying by 1, we should just eliminate both symbolic constants.

header/Constants.h
```diff
@@ -73,8 +74,10 @@ const int PERSON_VARIED_ABILITY_COUNT = 3;
 const double PERSON_ABILITY_STDDEV = 0.05;
 
 const double UNHEALTHY_PRODUCTIVITY = 0.5;
-const double DAILY_SICKNESS_CHANCE = 0.01;
-const double DAILY_RECOVERY_CHANCE = 0.1;
+const double ANNUAL_SICKNESS_CHANCE = 0.1;
+// const double DAILY_SICKNESS_CHANCE = 0.01;
+// const double DAILY_RECOVERY_CHANCE = 0.1;
```
Amittai-Aviram: Please do not leave commented-out code in the code base. Simply remove the statements.

header/Firm.h
```diff
@@ -145,12 +144,18 @@ class Firm : public Agent {
     void log_inventory_level(const Product * product, const int quantity);
     void log_inventory_reduction(const Product * product, const int quantity);
     void log_reorder(const Product * product, int quantity);
+    void log_initial_employment(const int worker_id, const int firm_id);
```
Amittai-Aviram, line 147: Can you not just pass the worker (`Person`) in? The worker has both `id` and `firm` fields.

header/Firm.h
```diff
@@ -145,12 +144,18 @@ class Firm : public Agent {
...
     void log_reorder_failure(const Product * product, int quantity);
+    void log_transfer_request();
     void log_product_quantity(
             const char * const label,
             const Product * product,
             const int quantity
             );
     void log_accepted_order(std::string product_name, int requested_turnaround_time);
     void log_demand(std::string product_name, double demand);
+    void log_pending_inventory(std::string product_name, double pending_inventory);
```
Amittai-Aviram: Can you identify the product by passing the `Product` in rather than using its name here? Right now, you are copying a string twice, plus you have a lot more flexibility and versatility by having the pointer to the `Product` object.

header/Person.h
```diff
@@ -57,6 +59,7 @@ class Person : public Agent {
     void update_busyness();
     void log_hours(const double hours);
     void log_purchase(const std::string& product_name, int quantity);
+    void log_shopping(double deficit);
```
Amittai-Aviram: Can the function get the deficit internally, without your having other pass it as a parameter?

src/compile_commands.json
```diff
@@ -0,0 +1 @@
+[]
```
Amittai-Aviram: Does this file belong in the code base and in your PR?

src/Distributor.cpp
```diff
@@ -40,6 +40,7 @@ Distributor::Distributor(
         product_to_plan[product] = plan;
         input_inventory[product] = quantity;
         log_inventory_level(product, input_inventory[product]);
+        log_catalog();
```
Amittai-Aviram, line 43: Do you want to log only distributors' catalogs? If you would like to log those of producers as well, maybe you should move this to the `Firm` constructor?

src/Firm.cpp
```diff
@@ -101,12 +102,15 @@ double Firm::get_busyness() {
 }
 
 std::vector<Person *> Firm::propose_transfer(int workers_wanted) {
-    if (get_busyness() >= society->get_busyness() - TRANSFER_BUSYNESS_THRESHOLD) {
+    double firm_busyness = get_busyness();
+    double societal_busyness = society->get_busyness();
+    int max_workers_to_transfer = (int) (workers.size() * (1.0 - firm_busyness / 
+            (societal_busyness - TRANSFER_BUSYNESS_THRESHOLD))); 
+    max_workers_to_transfer = std::max(max_workers_to_transfer, workers_wanted);
```
Amittai-Aviram, line 109: Why do we take the maximum? Just looking at the names of the variables, I would have guessed that we would use the minimum, so that we do not transfer workers out that we need here.

src/Firm.cpp
```diff
@@ -380,10 +387,13 @@ double Firm::get_demand(Product * product) {
 void Firm::move_worker_off_standby(Person * worker) {
     if (worker->get_firm() == nullptr) {
         society->get_unemployed_people().erase(worker);
+        log_initial_employment(worker->get_id(), id);
```
Amittai-Aviram, line 390: The name of this funtion is a little misleading. You're calling it "initial" employment because it's when a worker starts working at a firm — but the word "initial" to me means "at the beginning of time."

src/Firm.cpp
```diff
@@ -401,6 +411,44 @@ void Firm::log_inventory_reduction(const Product * product, const int quantity)
     log_product_quantity("inventory_reduction", product, quantity);
 }
 
+void Firm::log_initial_employment(const int worker_id, const int firm_id) {
+    Logger::get_instance()->log(
+        get_client_type(),
+        "newly employed",
```
Amittai-Aviram, line 417: Presumably, in most cases, this person is not newly employed, he/she has simply switched firms. Also, why did the `log_placement` function not work?

alexbcreiner0:
> It seems that, despite .gitignore, your PR wound up including a whole lot of build artifacts, which is why the PR includes so many files (33). Could you please fix this?

I think I fixed most of this by removing cached files and restoring some specific files that said they changed even though I hadn't touched them. However, I'm not sure what to do about the .gitkeeps or the producer_distributor_transactions_summary.txt. I've tried git restore --source=origin/main docs/graphs/producer_distributor_transactions_summary.txt and nothing seems to happen. I've never even opened this file so I have no idea what's going on there.

src/main.cpp
```diff
@@ -70,6 +72,29 @@ void set_params(int argc, const char ** argv, SimArgs& args) {
                         }
                     }
                     break;
+
+                case 's':
+                    ++i;
+                    dvalue = strtod(argv[i], NULL);
+                    std::cerr << "dvalue = " << dvalue << std::endl;
```
Amittai-Aviram: Is this dev code? Can you remove it?

src/main.cpp
```diff
@@ -70,6 +72,29 @@ void set_params(int argc, const char ** argv, SimArgs& args) {
...
+                    if (dvalue <= 0.0 || dvalue >= 1.0) {
+                        error = true;
+                    } else {
+                        args.sickness_chance = dvalue;
+                    }
+                    break;
+                case 'v':
+                    ++i;
+                    dvalue = strtod(argv[i], NULL);
+                    std::cerr << "dvalue = " << dvalue << std::endl;
```
Amittai-Aviram: Dev code?

src/Person.cpp
```diff
@@ -22,16 +22,15 @@ Person::Person(Society * society):
     static unsigned int unique_id = 0;
     id = unique_id++;
 
-    static std::normal_distribution<>
-        ability_dist(1.0, PERSON_ABILITY_STDDEV);
+    if (id == 1) {
+        std::cerr << "ability stddev = " << Sim::get_person_ability_stddev() << std::endl; 
+    }
```
Amittai-Aviram: This looks like a dev trace.

src/Person.cpp
```diff
@@ -106,11 +105,14 @@ void Person::purchase_good(Product * product, int quantity) {
...
             inventory[product] += quantity;
             log_purchase(product->product_name, quantity);
+            log_account();
             return;
         }
     }
 }
 
+
+// MAIN
```
Amittai-Aviram: ??

src/Person.cpp
```diff
@@ -131,6 +133,9 @@ bool Person::will_shop() {
             inventory[product] / product->mean_consumption_frequency
         );
     }
+    if (total_deficit > PERSON_DEFICIT_THRESHOLD) {
+        log_shopping(account);
+    }
     return total_deficit > PERSON_DEFICIT_THRESHOLD;
 }
```
Amittai-Aviram, line 134: Style:
```
    bool should_shop = total_deficit > PERSON_DEFITI_THRESHOLD;
    if (should_shop) {
        log_shopping(account);
   }
   return should_shop;
```

src/Person.cpp
```diff
@@ -155,6 +160,60 @@ void Person::shop() {
     }
 }
 
+// MINE
```
Amittai-Aviram: Whoa! Is this part of resolving conflicts?  Anyway, please remove these comments.

src/Person.cpp
```diff
@@ -155,6 +160,60 @@ void Person::shop() {
...
+//     for (std::pair<Product *, int> p : purchase_quantities) {
+//         int quantity = (int) (price_scalar * p.second);
+        
+//         if (quantity > 0) {
+//             purchase_good(p.first, quantity);
+//         } else {
+//             log_shopping_deficit((double) quantity); 
+//         }
+//     }
+// }
```
Amittai-Aviram: Please remove all commented-out code. If you'd like to hold onto it as a draft, please store it elsewhere.

src/Person.cpp
```diff
@@ -166,17 +225,27 @@ void Person::retire() {
 }
 
 void Person::update_health_status() {
-	static std::uniform_real_distribution<> dist(0, 1);
     bool changed = false;
-	if (health_status == HEALTHY &&
-		dist(Sim::get_random_generator()) < 1 - pow(1 - DAILY_SICKNESS_CHANCE, 1.0 / DAY)) {
-		health_status = UNHEALTHY;
+    auto& rng = Sim::get_random_generator();
```
Amittai-Aviram: Please do not use `auto` — it defeats the point of C++ explicit typing.  Either use the original type (`std::mt19937&`) or just make the call where you need it.  Also — replace the name `rng` with `rand_gen` if you do use a local variable?

src/Person.cpp
```diff
@@ -166,17 +225,27 @@ void Person::retire() {
...
+
+    double annual_prob = Sim::get_annual_sickness_chance();
+    double sickness_rate = -std::log(1.0 - annual_prob);
+
+    double p_sick_hour = 1.0 - std::exp(-sickness_rate / 8760.0);
```
Amittai-Aviram: Please replace this numeric literal with a symbolic constant (or an arithmetic expression using a symbolic constant).

src/Person.cpp
```diff
@@ -166,17 +225,27 @@ void Person::retire() {
...
+    double p_recover_hour = 1.0 - std::exp(-1.0 / (AVG_DAYS_TO_RECOVERY * 24.0));
```
Amittai-Aviram: Can we avoid the literal 24.0? Either define a symbolic constant HOURS_PER_DAY or change AVG_DAYS_TO_RECOVER to an hours figure — or define an hours figure constant in Constants.h derived from thd days figure?

src/PriceController.cpp
```diff
@@ -25,19 +26,20 @@ void PriceController::update_price(Plan * plan) {
...
-        units += plan->order->quantity;
+        units += plan->order->quantity - plan->quantity_remaining;
         hours += plan->labor_hours - plan->labor_hours_remaining;
         workers += plan->workers.size();
     }
     double price = product->living_labor_per_unit = hours / units;
     for (std::pair<Product *, double> input : product->inputs_per_unit) {
         price += input.first->price_per_unit * input.second;
     }
+    // std::cerr << "Updated living labor per unit plus cost of inputs is: " << price << std::endl;
```
Amittai-Aviram: Please remove.

src/PriceController.cpp
```diff
@@ -46,6 +48,7 @@ void PriceController::update_price(Plan * plan) {
         price += machine_cost_per_hour * machine_hours_per_unit;
     }
     product->price_per_unit = price;
+    // std::cerr << "Updating per-unit price to " << price << std::endl;
```
Amittai-Aviram: Please remove.

src/Producer.cpp
```diff
@@ -205,7 +207,7 @@ void Producer::end_plan(Plan * plan) {
     // update local labor time
     recorded_living_labor_per_unit[plan->order->product] = 
         (double) (plan->labor_hours - plan->labor_hours_remaining) 
-        / plan->order->quantity;
+        / (plan->order->quantity - plan->quantity_remaining); // RECORDED WAS CALCULATED WITHOUT USING THE QUANTITY ACTUALLY PRODUCED
```
Amittai-Aviram: Thanks — but now please remove the comment.
I still do not understand how plans could be completed without the order quanitty being produced. That seems wrong. It seems to me that the order should be completed even if it goes over budget. Maybe that amounts to the same thing, though.

compile_commands.json
```diff
@@ -0,0 +1,52 @@
...
+      "-std=c++17",
+      "-g",
+      "src/main.cpp",
+      "-o",
+      "bin/sim"
+    ],
+    "directory": "/home/alex/github/Agent-Based-Simulation-Model",
+    "output": "bin/sim"
+  }
+]
```
Amittai-Aviram: Does this file belong in the code base?

Amittai-Aviram (changes requested):
Thanks for all this great work! Please see my comments.
I understand the change to the Price Controller. I had not anticipated that we would be reporting plans as complete with some quantity having been produced that is less than the quantity of the order.  Is that the result of workers not working as fast as the plan had estimated that they would work? Is this the right thing to do, or should the plan continue until the order is complete, even if the plan therefore goes over budget?

alexbcreiner0:
> I had not anticipated that we would be reporting plans as complete with some quantity having been produced that is less than the quantity of the order

This never happens. The exact opposite is the case. To explain: the labor done each hour is discrete (ignoring modifiers due to health and ability for simplicity), but the living labor requirements are floating point numbers. Each hour, a floating point quantity is produced, and deducted from the quantity remaining (initially set to the order size). Right now, a plan is deemed finished when its quantity remaining is _less than or equal_ to zero. Because the hours worked are discrete, there will generally be some excess in the amount produced. For example, if the living labor time required to make something is 0.5, the last hour will always produce 2 per worker, and the final quantity produced will be at least one greater than the order quantity.

When this happens, the total quantity produced can be found by taking the original order quantity and subtracting the quantity remaining, which will add a little bit because it's a negative number. So these changes are necessary for taking stock of the overproduction specifically.

header/Constants.h
alexbcreiner0: I'm not sure, but that's why I left it in.

header/Constants.h
alexbcreiner0: I already changed this back to the original.

header/Constants.h
alexbcreiner0: Just because it's 1 now doesn't mean it always needs to be. I think it could be worthwhile to tweak at some point.

header/Constants.h
alexbcreiner0: Done

src/Producer.cpp
alexbcreiner0: See my response to your overall comment.

compile_commands.json
alexbcreiner0: I don't know what this file is. No matter what git commands I give, it doesn't go away. Not sure what to do.

src/Distributor.cpp
alexbcreiner0: It's already in both. I tried putting it in the Firm constructor initially, but the logging functions require a get_client_type() call, which is abstract for the Firm and only defined for the two child classes. So that was giving me compile errors. I think it is fine this way, as having the specific client type with the logged catalogs is helpful.

header/Person.h
alexbcreiner0: Done. deficit was a misnomer anyway, I only ever pass the account. Must have been in a hurry.

header/Firm.h
alexbcreiner0: I can pass the worker instead of the worker's id if you want, but the call to this function happens prior to the worker actually being assigned to the firm. If you look at where this happens (in the badly named function move_worker_off_standby), I think that's the best place for it, given the current state of this function as a whole. I guess I went with passing the ints for both because it was symmetric.

src/Firm.cpp
alexbcreiner0: This is Devin's code, not mine. All I did was replace all of the calls to get_busyness with my two doubles, so that I would have them for logging (also, it seemed redundant). This is how it was prior to this PR. Your criticism would seem to depend on what this variable is actually used for, but I didn't put much thought into it so I don't think I'm the right one to ask.

src/Firm.cpp
alexbcreiner0: It's initial in the sense of being the first time that this worker is employed by any firm, since they were unemployed before. What should I change it to

src/Firm.cpp
alexbcreiner0: No, there is a separate logging function for firm transfers. (It's called log_employment_transfer.) This logging only happens when someone is moved from unemployed to being employed. 

log_placement did not work because I wanted both the old firm and the new firm to be logged in the same packet of information. The person instance at a given moment only knows a single one of these firms.

src/main.cpp
alexbcreiner0: This is what I had to do in order to take floating point numbers as optional arguments when running the sim. Are you talking about the standard error print statement? That's the only thing here I see that can be done away with.

src/main.cpp
alexbcreiner0: See the case 's' response.

src/Person.cpp
alexbcreiner0: Yes, gone now.

src/Person.cpp
alexbcreiner0: Okay.

src/Person.cpp
alexbcreiner0: Okay.

src/Person.cpp
alexbcreiner0: Sure.

header/Firm.h
alexbcreiner0: Sure.

Amittai-Aviram:
> I think I fixed most of this by removing cached files and restoring some specific files that said they changed even though I hadn't touched them. However, I'm not sure what to do about the .gitkeeps or the producer_distributor_transactions_summary.txt. I've tried git restore --source=origin/main docs/graphs/producer_distributor_transactions_summary.txt and nothing seems to happen. I've never even opened this file so I have no idea what's going on there.

I don't know, either, but I would guess that it has to do with your IDE and how it interacts with Git.  With simple Vi and the command line, I do  not run into these issues.  Oddly, it says that the .gitignore files only have whitespace changes, but then I cannot get it to display those changes.  The .gitignore files should be _empty_, so I wonder whether your IDE actually added a space character or a newline to them. Meanwhile, however, I just cloned the repo on my local machine and fetched your branch, and I can see that the directories that are supposed to be preserved by virtue of the .gitignore  file — build and bin — are missing. And you have `compile_commands.json` in the top-level diretory, which should not be there and looks to me to be an IDE artifact.

Amittai-Aviram:
> > I had not anticipated that we would be reporting plans as complete with some quantity having been produced that is less than the quantity of the order
> 
> This never happens. The exact opposite is the case. To explain: the labor done each hour is discrete (ignoring modifiers due to health and ability for simplicity), but the living labor requirements are floating point numbers. Each hour, a floating point quantity is produced, and deducted from the quantity remaining (initially set to the order size). Right now, a plan is deemed finished when its quantity remaining is _less than or equal_ to zero. Because the hours worked are discrete, there will generally be some excess in the amount produced. For example, if the living labor time required to make something is 0.5, the last hour will always produce 2 per worker, and the final quantity produced will be at least one greater than the order quantity.
> 
> When this happens, the total quantity produced can be found by taking the original order quantity and subtracting the quantity remaining, which will add a little bit because it's a negative number. So these changes are necessary for taking stock of the overproduction specifically.

Got it! Thanks!  I have a feeling that there should be a clearer way of doing this, but I cannot think of it just now, and, meanwhile, your changes should go in.

header/Firm.h
Amittai-Aviram: OK.

src/Distributor.cpp
Amittai-Aviram: OK. Yeah, you would not be able to call it from the `Firm` constructor, but you could call it from any other `Firm` function — and you should get the correct result because of dynamic dispatch.

src/Firm.cpp
Amittai-Aviram: Yes, I saw that it was in Devin's code — I was trying to follow the "fix as you go" principle.  I'll try to remember to ask Devin.

src/Firm.cpp
Amittai-Aviram: OK, I was not sure whether the worker was coming on only from the unemployed list. In that case, it is correct. Sorry — it's hard to grasp all the context during a review.

src/Firm.cpp
Amittai-Aviram: So should you remove `log_placement`?

Amittai-Aviram (changes requested):
I just have one remaining question (about `log_placement`) — but the more important issue is restoring the `build` and `bin` directories, with their respective `.gitkeep` files, and also removing that `compile_commands.json` file.

alexbcreiner0:
> With simple Vi and the command line, I do not run into these issues

I also use Vi (well, a popular fork of it) and only do git commands from the command line. So I don't think this is it.

Amittai-Aviram (approved):
LGTM

PR #186, opened 2026-04-20
distributor can make plans

### LINKED ISSUE #185: Distributor Only Stays on One Plan
As title suggests

Sal47506:
Resolves #185


Productivity-Incentives/Makefile
```diff
@@ -0,0 +1,40 @@
+CXX = g++
```
Amittai-Aviram: Whoa! How did this file get in here? Could you please remove it from the PR? Thanks!

Productivity-Incentives/README.md
```diff
@@ -0,0 +1,14 @@
+# The Productivity Incentives Sub-Project
```
Amittai-Aviram: Please remove this file.

src/Distributor.cpp
```diff
@@ -54,6 +37,11 @@ void Distributor::on_time_step() {
         for (Person * worker : plan->workers) {
             worker->register_hours_worked(1);
         }
+        if (plan->outgoing_units_consumed == plan->order->quantity  i0) {
```
Amittai-Aviram: For safety and future-proofing, maybe make this >= rather than == ?

src/Distributor.cpp
```diff
@@ -54,6 +37,11 @@ void Distributor::on_time_step() {
...
+            start_new_plan(plan->order->product, plan->order->quantity);
```
Amittai-Aviram: I think you need to do more than start a new plan. You also need to report the completed plan to the Price Controller — right?

src/Distributor.cpp
```diff
@@ -105,6 +93,28 @@ void Distributor::check_expand_catalog() {
...
 
+void Distributor::start_new_plan(Product * product) {
+    int quantity = product->mean_consumption_frequency *
+        (FIRM_STOCKPILE_DURATION + FIRM_DEMAND_WINDOW_MIN * DISTRIBUTOR_INITIAL_INVENTORY_MULT) *
+        Sim::get_num_people();
+    Order * order = new Order(product, quantity, this, 0);
+    Plan * plan = new Plan;
+    plan->order = order;
+    plan->firm = this;
+    plan->labor_hours = DISTRIBUTION_LABOR_PER_UNIT * quantity;
```
Amittai-Aviram, line 106: This needs to change. At the beginning of time, it's appropriate, but, from now on, the hours needed should be taken from the price controller, which should register it in the `living_labor_per_unit` field. (Recall that a `ConsumerGood` is a `Product`.) I think what we should probably do is set the consumer good's living labor per unit to be that initial constant, which is then updated by the price controller.  In the code here, we should read that field instead of the constant and multiply it by the quantity to get the plan's labor hours.

Amittai-Aviram (changes requested):
Good so far, Sal!  There's some more that needs to be done, though. See especially my comment on the new `start_new_plan` function. In making the next round of changes, please check the Price Controller's logic and make sure that it will do the right thing with an order of consumer goods.

Productivity-Incentives/README.md
ghost: Removed

Productivity-Incentives/Makefile
ghost: Removed

src/Distributor.cpp
ghost: Added >=

src/Distributor.cpp
ghost: Called Price Controller's update price function to update price after plan ends

src/Distributor.cpp
ghost: Called `assign_workers` to assign workers on the plan and later called `predict_labor_hours` to assign labor hours to said plan

Amittai-Aviram: We do not need to assign a value to this field at all at this point, then. Sorry! My suggestion was unclear. That constant, `DISTRIBUTION_LABOR_PER_UNIT`, should be the initial value of the `living_labor` field when a `ConsumerGood` object is constructed. (The value can be written into the type definition itself.) It will then be used as the first `recorded_living_labor_per_unit` for this consumer good. At the end of each plan, `recorded_living_labor` is updated with the _actual_ living labor per unit of the plan coming to an end. This same value is used to predict the living labor for the next plan. Please follow the model of the producer.

Amittai-Aviram (changes requested):
One more small change, please — plus, please just make sure that the plan _does_ get reported to the Price Controller, that the current plan's actual living labor is used to update the recorded living labor per unit, etc. You will also have to resolve merge conflicts — please do so carefully.  😊

PR #189, opened 2026-04-24
Bugfixes, smaller and iterative order sizing

### LINKED ISSUE #170: Bugfixes (Devin)

### LINKED ISSUE #172: Have distributors distribute more than one product
Currently, distributors monopolize the distribution of products — and specialize to only one product each. This seems unrealistic and implies that a consumer must visit as many distributors as it has goods to buy. A reasonable approach is to tie the likelihood of a distributor to carry a good to the that good's mean consumption frequency.

### LINKED ISSUE #176: Making reorders and plan rejection smarter (Devin)

lennyyyyyyyy:
Resolve #170 
Resolve #172 
Resolve #176 
Other changes:
- A new constant `INITIAL_ACCOUNT_DURATION` gives an initial account to each person, which is an amount that will allow each Person to consume and purchase goods for that amount of time, given that they earn no extra income. Of course, People will earn income, but I think this is a good way to specify the starting amount of labor credits in circulation.
- Distributors carry all products from the start of time, which is more realistic. Logic for expanding the Distributor catalog is removed, since they already have a full catalog.
- Producers are perfectly allowed to set themselves as suppliers, which is necessary if a product's input list includes itself.
- `try_sell_goods` now returns an integer instead of a boolean, which allows People to split their shopping event among multiple Distributors.


header/Constants.h
```diff
@@ -11,7 +11,7 @@ const unsigned int YEAR = DAY * 365;
 
 const unsigned int INITIAL_WORK_HOURS_DAILY = 8;
 const unsigned int INITIAL_WORK_DAYS_WEEKLY = 5;
-const double INITIAL_ACCOUNT_MULT = 10;
+const double INITIAL_ACCOUNT_DURATION = 2 * MONTH;
```
alexbcreiner0, line 14: I think this is deprecated code

lennyyyyyyyy:
Tried running on the command line, still getting memory leaks. Found that it is in part due to Producers being able to add themselves as suppliers, but I'm not completely sure why.
Some seg faults went away after that, but some stayed:
image
image
image

I honestly don't know what some of these are.
Since the cause is definitely a result of the merge, I tried just copying the version of my branch before the merge and applied just Alex's distributed consumption frequencies to it, which was the major change from his PR, besides more realistic sickness modeling and the lognormal distribution from abilities. This works fine (you can see that the consumption frequencies are now not all equal, because the Distributor supply differs).
image
image
Logically, this suggests that the cause is the latter two changes from the merge, but I don't really know where to start with the messages I'm getting from LLDB.

In other news, I think the crazy busyness is just a logging issue. Printing from the command line shows that societal busyness never goes above 1 (for my constants, before changes to the work day, it hovers around 0.1). Printing transfers to cmd also shows that transfers work fine. My branch is consistently stable with varying consumption frequencies without relying on a good employment split from the beginning of time, so this makes sense.

src/Firm.cpp
alexbcreiner0, line 1: In the `propose_transfer` function here, could you change the logging call to `log_busyness(firm_busyness, societal_busyness, max_workers_to_transfer);`? Right now societal_busyness and max_workers_to_transfer are reversed, thats what was causing the weird reporting we were seeing.

src/Firm.cpp
```diff
@@ -62,30 +59,22 @@ void Firm::add_supplier(Producer * producer) {
 
 void Firm::receive_shipment(Plan * plan) {
     Order * order = plan->order;
-    if (order->status != Order::ORDER_FINISHED) {
-        std::cerr << "Attempted to recieve a shipment for an incomplete order."
-            << std::endl;
-        return;
-    }
-    input_inventory[order->product] += order->quantity;
+    input_inventory[order->product] += order->quantity - plan->quantity_remaining;
```
alexbcreiner0, line 62: Good catch! We were accidentally deleting surplus goods.

src/Firm.cpp
```diff
@@ -148,18 +136,30 @@ Producer * Firm::select_fastest_supplier_for_order(Order * order) {
...
+    if (!chosen_producer) {
+        for (Producer * producer : secondary_producers) {
+            int draft_plan_time = producer->draft_plan_or_reject(order);
+            if (draft_plan_time == DRAFT_ORDER_REJECTED) {
+                producer->drop_order(order);
+            } else {
+                order_time = draft_plan_time;
+                chosen_producer = producer;
+                break;
+            }
```
alexbcreiner0: A little confused by this. Secondary producers seem to just be defined as those who don't produce the commodity. Wouldn't every draft plan then just be rejected here? Why have it at all in that case?

alexbcreiner0 (changes requested):
This is fantastic work Devin. The employment issues were a logging artifact, and the memory leak issue was also a problem on my loggers part. The success of these changes is evident in the following result I was able to get - two years of relatively stable evolution! 

Image

There are still some issues with random crashes, and I think that the busyness calculations should still be done using an actual queue to model a moving time window, but I think it is imperative to get these changes merged into main as quickly as possible, so that we are finally left with a simulation which works relatively well! The only changes I absolutely need to see are those I've requested in my comment [here](https://github.com/BC-LTEWG/Labor-Time-Economy-Simulation/pull/189/changes#r3150047909). After that, you should be good to merge. 

EDIT: If this PR were merged into main in its current state, the simulation will be less stable than it was prior. However, when it doesn't crash, it gives us a very real model which, while still having problems, marks the major milestone in the project. So it can't be my decision exclusively whether this should be merged right now or if the crashes should be looked into prior. Hopefully you and Amittai will be able to make some progress there when you meet this week.

alexbcreiner0:
> I tried just copying the version of my branch before the merge and applied just Alex's distributed consumption frequencies to it, which was the major change from his PR

I think that you are right about this, but to be more specific, what is going on is that my new consumption frequencies, being more varied and prone to imbalances, are causing numerical stress to the system which your numbers couldn't. What is really needed is for our simulation to be safeguarded against overflow errors across the board. This won't likely be just a single thing which needs to be changed in a specific place. More likely there are three or four places in the code where something like a tiny double is sometimes divided that causes numbers to blow up, etcetera. Maybe I should make this its own issue.

src/Firm.cpp
lennyyyyyyyy: Done.

src/Firm.cpp
lennyyyyyyyy: Think this was an artifact of something I tried doing before the iterative order sizing, as a solution to when all available producers reject. I think you're right that they'd all reject regardless.

alexbcreiner0: Okay then this should definitely be deleted.

src/Firm.cpp
```diff
@@ -204,34 +175,30 @@ void Firm::reorder_input_product_to_threshold(
...
+    for (int i = FIRM_REORDER_ATTEMPTS; i > 0; i--) {
+        double reorder_prop = FIRM_REORDER_START * i / FIRM_REORDER_ATTEMPTS;
+        order->quantity = std::ceil(reorder_quantity * reorder_prop);
+        order->requested_turnaround_time = std::max(1.0, reorder_deadline * reorder_prop);
+        Producer * chosen_producer = send_order(order);
+        if (chosen_producer) {
+            log_reorder(product, reorder_quantity);
+            log_accepted_order(product, order->requested_turnaround_time);
+            return;
+        }
```
alexbcreiner0, line 199: Why are we making the deadline more restrictive while the quantity becomes less restrictive? I think that this is going to lead to a lot of pointlessly impossible orders. I don't think that the duration needs to be restricted at all, to be honest. As long as the firms are choosing producers according to desired turnover times and the producers are being honest in their drafting, the durations should fall on their own without it being demanded of them, right? (I might be wrong here, but you should understand that this will sometimes make it harder for a producer to satisfy orders, not easier.)

I think that we need to take repeated order attempts on the same time step like this very seriously, as it is a computationally intensive task. I am seeing many scenarios in my testing where this leads to the simulation gets stuck on certain time steps in pointless reordering loops, in particular when the number of commodities is something like 3, but the number of producers is greater than this.

10 reorder attempts is way too much for something like this, when you are taking a fourth of the quantity each time. The max reorder attempts constant doesn't need to exist at all, just have the FIRM_REORDER_START be an integer (in this case, 4), use the inverse of it (1/4 = 0.25), and then determine the max of attempts as the ceiling of log base 4 (or log base whatever FIRM_REORDER_START is) of the initial order size. This will make it so that a firm always stops after the order size becomes 1. A fourth is a strange choice as well. Why not half? 4 is a very steep drop-off, though I suppose I can't complain given that these repeated reorders are computationally taxing.

At present though, if we have an initial quantity of 100, a duration of 10, then the firm will attempt to place an order for 1 of something with a duration of 1 hour 8 consecutive times before giving up. That is not acceptable when it requires all other producers to draft a new plan every attempt. But if you did it the way I suggested above, the firm would only try 3 additional times - one order for 25 with a duration of 3, one order for 6 with a duration 1, and one order for 1 with a duration 1.

alexbcreiner0 (changes requested):

header/Constants.h
```diff
@@ -37,37 +37,30 @@ const int PRODUCT_NUM_INPUTS_MIN = 1;
...
 const int MACHINE_LIFETIME_MIN = 1000;
 const int MACHINE_LIFETIME_MAX = MACHINE_LIFETIME_MIN * 10;
 
 const int FIRM_STOCKPILE_DURATION = 1.5 * WEEK;
-const unsigned int FIRM_DEMAND_WINDOW_MIN = WEEK;
-const unsigned int FIRM_DEMAND_WINDOW_MAX = MONTH;
-const double PRODUCER_INITIAL_INVENTORY_MULT = 10;
-const double DISTRIBUTOR_INITIAL_INVENTORY_MULT = 2;
+const int FIRM_DEMAND_WINDOW_MIN = WEEK;
+const int FIRM_DEMAND_WINDOW_MAX = MONTH;
```
alexbcreiner0, line 56: So the fact that these were unsigned ints before was I guess truncating the demand window, which was the source of firms waiting too long to place orders? I would never have found this bug. Great job.

alexbcreiner0 (approved):
This is a major milestone in the project. We almost have a fully working system. Just some efficiency issues to resolve and also the busyness problem, but those will be left to future PRs. Great work Devin.

PR #195, opened 2026-05-03
Revise smaller ordering to be more efficient, fix issues relating to falling wealth

### LINKED ISSUE #194: Avoid slowdown from reordering loops
Right now, with Devin's latest PR merged, producers will try a certain number of times to place orders of smaller amounts when they don't find a satisfactory draft plan returned to them. This process is very computationally inefficient and can slow the simulation down to a crawl. A better strategy that we worked out together would be to avoid having multiple 'rounds' of drafting at all, and instead have the producer firm, rather than rejecting when they can't meet the requirements of an order, draft a plan for a smaller quantity than what was requested, and offer both that quantity as well as the expected turnover time to the customer. This way, the customer doesn't have to ask any further questions, and can accept the feasible offer as 'better than nothing' at their own discretion.

### LINKED ISSUE #196: Improve memory safety of `move_plans_forward_one_step`
Use a temporary vector to ensure no illegal memory is touched in Producer's `move_plans_forward_one_step`

### LINKED ISSUE #197: Falling wealth - PriceController approximation, incorrect input consumption
Falling wealth seems to be a result of two issues:

First, the PriceController adds an _ideal_ contribution of the inputs to the price of the output product, by simply multiplying the documented price of the inputs by the inputs_per_unit measure. However, products are made with an integral amount of input (which is always rounded up, more realistic than rounding down), which distorts the true value of the output product. Fortunately, plans keep track of the actual raw material value of the inputs used, so the PriceController can use that instead. However, these input prices are the prices _at the time of plan initiation_, not plan completion, but I don't think this really matters.

The other issue is that a little while ago we decided to keep the excess of Plan production, since plan runtime is quantized in hours. However, the actual inputs consumed, as well as their cost, is calculated with the _planned_ production. Since we can never know how much we may produce in excess, it's not reliable to remove the inputs for the excess when we make it (we may subtract from an empty inventory). 
From this point we have two solutions:
 - Pretend we don't make the excess, and also don't consume the raw materials required to make the excess.
 - Keep as much of the excess as we can, and consume the corresponding inputs as we can at the time of plan completion to make up for this.

Image

lennyyyyyyyy:
Resolve #194 
Resolve #196
Resolve #197 
I also got rid of the constant `DEADLINE_SAFETY_MULT`, which was sort of the previous solution to huge risky orders, but is unnecessary now.


Amittai-Aviram (changes requested):
These changes mostly look good, especially the bug fix I had requested — but I don't understand the basic idea here regarding order size. The order size is now the minimum of (a) a quarter of the producer's reorder threshold for that product and (b) the maximum over all producers of the required input product of how much that producer has on hand of any secondary input needed for that product _per unit_ of the required product. Why? What is the rationale for this formula?

src/Firm.cpp
```diff
@@ -212,9 +169,41 @@ void Firm::check_and_reorder_input(Product * product) {
...
+        pending_inventory *
+        FIRM_STOCKPILE_DURATION / threshold * 
+        reorder_quantity / threshold;
+    return new Order(
+            product,
+            std::ceil(reorder_quantity),
+            this,
+            std::max(1.0, reorder_deadline)
+            );
 }
```
alexbcreiner0: So the way you've set this up, if I've understood the code correctly, producers are now looping over and querying all suppliers twice - once to figure out what the max reorder size is that they can get away with, and then again after that with the old requests for draft plans based on that. What I remember us talking about is that the suppliers, upon being queried in the usual `draft_plan_or_reject` method, draft a plan for a smaller amount rather than just rejecting when they don't have the necessary inputs. They would then return that quantity of the reduced order along with the usual `predicted_turnaround_time`. 

I am not sure why you are doing this clunky preemptive round of communication instead of that. Aside from being redundant, this way seems very faulty in the sense that if the firm is picking the max reorder quantity that they think they can get from any producer, and then it turns out that the (likely) only producer capable of meeting that demand is too busy to take the order, then the firm placing the order will just be screwed. I take it your `FIRM_MAX_REORDER_PROP` constant is here to muffle the firm's ambitions and reduce the chance of this happening, but if you had just done it the way we originally talked about, that band-aid measure wouldn't be necessary at all, nor would this entire extra round of communication between customer and supplier. 

I'm going to need some explanation as to why it's a bad idea to do it the way we originally talked about, or better yet a redo which is more in line with the original idea we discussed last week. Why can't we just edit the `draft_plan_or_reject` method to return a reduced order quantity, and have the firm pick from those options?

Amittai-Aviram (dismissed):
Please write an issue to revise the names of max_self_reorder_quantity, get_max_reorder_quantity, and max_producer_reorder quantity in order to make the rationale more obvious.

alexbcreiner0:
I don't approve of this being merged into main yet.

alexbcreiner0 (changes requested):

src/Firm.cpp
alexbcreiner0: I forgot to click "Submit Review" here, so nobody saw this. Ugh.

lennyyyyyyyy: > I take it your FIRM_MAX_REORDER_PROP constant is here to muffle the firm's ambitions and reduce the chance of this happening...

This constant used to be `FIRM_REORDER_START`. Just an upper bound to make order sizes smaller and ensure stability. I just renamed it.

Other than that, you're right that the likely single Producer that is asked to complete the order may end up rejecting from a lack of workers. My bad, I thought this solution was pretty much equivalent (having the Firm figure out the maximum possible order size first, and then doing the drafting as usual).

I've changed it so that Producers send back an updated **order** (not a draft plan, like you said), because the draft plan includes a lot of unnecessary information, when all the downstream Firm needs is the quantity and how long it's expected to take. I had to change some things about how draft plans are organized and kept within each Producer, since the values inside each order are modified by each Producer that gets sent one (specifically, you can't look for a draft plan by order anymore, but by the requesting customer).

Now, since Producers can approve orders of various different quantities, based on what they're capable of, downstream Firms decide which one to move forward with by the highest rate of production. Let me know what you think.

alexbcreiner0 (approved):
Great job!

PR #199, opened 2026-05-13
Don't produce extra quantity, make inventories real

### LINKED ISSUE #198: Make the surplus production in final hour have real cost
Right now we are allowing firms to work through the final hour of production even if that means an amount is produced which exceeds the order amount. I think this is a good system, but since inputs are being set aside before production begins, this extra amount produced is effectively produced for free. This is causing prices to drift below their values, and consequently an upward drift in income over time. 

To fix this, I propose we implement a special method which is called in the final hour of production, which computes the minimum of these two quantities:
- The extra product which would be produced during this final hour
- The maximum product which _can_ be produced given the firm's current inventory. 

and which then allocates these additional resources to the plan. The final amount produced in this last hour is also given by this calculation. Workers can still be paid for the final hour of production, but at that point we will be able to fairly account for the production of these extra goods, which should stabilize prices a good amount.

lennyyyyyyyy:
Resolve #198


header/Firm.h
```diff
@@ -44,19 +44,19 @@ struct Order {
     Product * product;
     int quantity;
```
Amittai-Aviram, line 45: I'm a little worried about this — if we shall later compare it to a double value to determine whether the order is complete, or if we use it in arithmetic in the Price Controller.

header/Firm.h
```diff
@@ -70,16 +70,16 @@ class Firm : public Agent {
     virtual Logger::Client get_client_type() = 0;
     virtual void on_time_step() override;
     double get_avg_productivity();
-    virtual int get_inventory(Product * product);
+    virtual double get_inventory(Product * product);
```
Amittai-Aviram: Could we change this name to `get_inventory_level`?

header/Firm.h
```diff
@@ -90,20 +90,20 @@ class Firm : public Agent {
...
+    std::unordered_map<Product *, double> total_demands;
     std::unordered_map<Product *, std::unordered_set<Order *>> product_to_outbound_orders;
     std::unordered_map<Product *, double> recorded_living_labor_per_unit;
     std::vector<Plan *> plans_in_progress;
 
     Producer * send_order(Order * order);
-    bool remove_input_from_inventory(Product * product, int quantity);
-    void add_input_inventory(Product * product, int quantity);
+    bool remove_input_from_inventory(Product * product, double quantity);
+    void add_input_inventory(Product * product, double quantity);
```
Amittai-Aviram: Change this name to `add_input_to_inventory`?

header/Firm.h
```diff
@@ -90,20 +90,20 @@ class Firm : public Agent {
...
     double get_reorder_threshold(Product * product);
-    int get_pending_input_inventory(Product * product);
+    double get_pending_input_inventory(Product * product);
```
Amittai-Aviram: `get_pending_inventory_level` — and I wonder whether that name is clear enough. The function returns the _anticipated_ inventory level, which includes the _pending_ orders. Maybe `get_anticipated_inventory_level` might be clearer. What do you think?

src/Distributor.cpp
```diff
@@ -67,7 +67,7 @@ int Distributor::try_sell_goods(Product& product, int quantity, Person * person)
     add_demand_signal(&product, quantity);
     check_and_reorder_input(&product);
     if (!catalog.count(&product)) return 0;
-    int available = std::min(get_inventory(&product), quantity);
+    int available = std::min(static_cast<int>(get_inventory(&product)), quantity);
```
Amittai-Aviram: Why are `available` and `quantity` ints?

src/Distributor.cpp
```diff
@@ -67,7 +67,7 @@ int Distributor::try_sell_goods(Product& product, int quantity, Person * person)
...
     if (available < quantity) {
```
Amittai-Aviram, line 71: If we change `available` and `quantity` to be doubles, then this comparison could be slightly tricky. Basically, you have to define some tolerance `epsilon` below which a difference between the two values is considered trivial.

lennyyyyyyyy:
Wealth is stable while prices hug values:
image
image

src/Producer.cpp
Amittai-Aviram: Please change `0` to `0.0`.

Amittai-Aviram (changes requested):
Looks great! Please make hte changes I suggested. Thanks!

Amittai-Aviram (approved):
LGTM!

PR #201, opened 2026-05-15
Implement string argument flags and deterministic RNG option

### LINKED ISSUE #184: Make argument flags strings, not characters
Right now it seems like argument flags are limited to being a single character, e.g. I can add a flag for -j but not a flag for --sick_chance. This seems a bit overly restrictive. We will run out of letters quickly I think.

### LINKED ISSUE #190: Provide a deterministic pseudo-random generator option
Currently, all our (pseudo-)random generation uses a Mersenne-Twister-based generator (`std::mt19937`) that is a field of the singleton class `Sim`, which, in turn, is seeded by an `std::random_device` object.  This ensures that the generated values will not be repeatable &mdash; but we actually need a repeatable sequence for development and debugging. Please provide a deterministic option, together with a command-line flag to enable it.

nathanjlim:
Resolves #184, Resolves #190


src/main.cpp
```diff
@@ -12,104 +14,81 @@ void print_usage() {
...
+        
+        if(arg == "-n" || arg == "--time-steps" ||
+           arg == "-p" || arg == "--people" || 
+           arg == "-h" || arg == "--work-hours" || 
+           arg == "-w" || arg == "--work-days" || 
+           arg == "-o" || arg == "--products" || 
+           arg == "-m" || arg == "--products-per-machine" || 
+           arg == "-r" || arg == "--producers" || 
+           arg == "-d" || arg == "--distributors" ||
+           arg == "-e" || arg == "--seed") {
```
Amittai-Aviram: Can you please refactor this long compound condition as one of two neater solutions? —
1. Call a helper function. In the helper function, iterate over an array or vector of these string literals. If you hit a match, return true; otherwise, false.
2. Put all the allowable strings into a std::set and check for membership.

src/main.cpp
```diff
@@ -12,104 +14,81 @@ void print_usage() {
...
-                case 'j':
-                    args.json = true;
-                    break;
-                default:
-                    error = true;
-                    break;
+            long negative_check = strtol(argv[++i], NULL, 10);
 
+            bool seed_exception = (arg == "-e" || arg == "--seed");
+            if(negative_check < 0 || (negative_check == 0 && !seed_exception)) error = true; //prevents "0" seed from sending error
```
Amittai-Aviram: Style: please always enclose the condition or loop body in braces, with the opening brace on the same line as the conditional expression or loop header, and each statement on a line by itself. Please see the C++ style guide on the wiki.
Also, please limit comments to the absolutely necessary (i.e., cases where the code would otherwise be too baffling &mdash; but try hard not to write baffling code!) &mdash; and always put the comments on a line (or block) by itself above the item to which they refer, never to the right of it.

src/main.cpp
```diff
@@ -12,104 +14,81 @@ void print_usage() {
...
+            else {
+                value =
+                    static_cast<unsigned int>(negative_check);
+                    if (arg == "-n" || arg == "--time-steps") args.time_steps = value;
```
Amittai-Aviram: Please fix indentation from here down.
Please use a `switch` statement rather than a ton of `else if`s after an `if`.  (This is a general style principle: avoid long `if`...`else if`... constructs.)

src/Sim.cpp
```diff
@@ -59,6 +59,10 @@ bool Sim::does_json() {
 
 void Sim::set_params(SimArgs& args) {
     this->args = args;
+    
+    if(this->args.fixed_seed) {
+        this->gen.seed(this->args.seed);
```
Amittai-Aviram, line 64: Wait. What if the user does not specify a seed. Then it should be set by the random device, right? But you've initialized the seed in the `Sim` class definition to a default value of 0. So where do you get to initialize it with the random device?

Amittai-Aviram (changes requested):
Good so far. I'm confused about one thing: do you still ensure that the default behavior of the random number generator is to use the random device rather than a seed of 0? I'm afraid it does not look like it to me.

src/main.cpp
nathanjlim: I originally looked into using a std::set as you suggested, but realized that while it handles the validation cleanly, I'd still be stuck with a long if-statement chain when trying to route all of the logic together.

So, I implemented a std::map that links the string arguments to custom enum values. This way, the argument parser uses a clean valid_args.count(arg) to verify the input, and then immediately passes the mapped enum into a switch statement.

src/Sim.cpp
nathanjlim: When the simulation starts, the generator is immediately initialized with a random number right in the Sim constructor: "Sim::Sim() : gen{rd()} {}".

Later, when the program checks the arguments, it looks at if(this->args.fixed_seed). 
If the user doesn't provide the -e flag, fixed_seed remains false. Then, the program skips that if statement entirely and never touches the default value of 0. It just keeps the true random seed it started with. 

(I also verified this locally: comparing consecutive runs without a seed yields completely different outputs, confirming it doesn't stall on 0)

src/main.cpp
nathanjlim: Fixed. The indentation has been corrected and the logic is now handled by the new switch block.

src/main.cpp
nathanjlim: I've updated the formatting to follow the style guide and moved the inline comments to their own lines.

src/main.cpp
Amittai-Aviram: That's a reaonable solution. But you should not need any long if-statement. You initialize the set the same way as you now do your map, and then you just check whether the argument is _in_ the set, which means that it's valid.  (Or maybe I do not understand the problem to which you refer &mdash; ?) One tiny advantage your solution has &mdash; that I can see &mdash; is that a switch statement is more efficient if the item on hwich you switch is an enumerator (i.e., an integer) than if it's a string.

But I'm going to ask you to replace the `map` with an `unordered_map`.

Amittai-Aviram (approved):
Excellent work!
Please replace `std::map` with `std::unordered_map` in `main.cpp`. The latter is faster, and you do not need to iterate over the args, so you do not need them to be stored in a particular order. Otherwise, LGTM.

src/main.cpp
nathanjlim: I've swapped it out for a std::unordered_map and pushed the update. 
Thanks for the catch!

PR #208, opened 2026-05-19
Simplify logging API and implementation

Amittai-Aviram:
This PR represents the following changes to the logger:
- Replace the proliferation of overloads and API options with a single variadic function template.
- Replace the `Logger` singleton class with a simple `static` function, where the class name serves merely as a namespace designator.
- Replace the output format with a uniform JSON listing containing the time step, client name, client instance ID, label, and 0 or more key-value pairs, in which the key is always a string and the value may be either a number or a string.
- Update all calls accordingly (and remove obsolete calls.)
- Remove unnecssary header includes and otherwise clean up.

Note that most calls could rely on the compiler's type inference for routing, but every template specialization had to be invoked by at least one call with explicit type parameters, in order to trigger the compiler to generate the necessary function specialization.  Which call does this is arbitrary.


Amittai-Aviram:
Illustrative examples from running `bin/sim -j`:
```
{"t":0,"client":"Society","id":0,"label":"A","dim":120}
{"t":0,"client":"Society","id":0,"label":"l","dim":120}
{"t":0,"client":"Society","id":0,"label":"mean_consumption_frequency","product_id":0,"value":0.00155889}
{"t":0,"client":"Society","id":0,"label":"price","product_id":0,"price_per_unit":1.37692}
{"t":0,"client":"Producer","id":0,"label":"inventory_level","product_id":1,"amount":134.619}
{"t":0,"client":"Producer","id":0,"label":"catalog","product_id":0}
{"t":0,"client":"Distributor","id":100,"label":"catalog","product_id":49}
{"t":0,"client":"Distributor","id":100,"label":"inventory_level","product_id":90,"amount":6}
{"t":0,"client":"Person","id":0,"label":"ability","ability":2,"value":1.02153}
{"t":0,"client":"Person","id":999,"label":"inventory","product_id":99,"amount":0}
{"t":0,"client":"Person","id":999,"label":"account","value":205}
{"t":2,"client":"Person","id":952,"label":"shopping_deficit","deficit":0}
{"t":2,"client":"Person","id":953,"label":"is_shopping","account":205}
{"t":2,"client":"Person","id":953,"label":"shopping_deficit","deficit":0}
```

src/Society.cpp
```diff
@@ -319,53 +323,48 @@ void Society::retire_person(Person *person) {
...
-                        "A",
                         id,
-                        std::make_pair(i, j),
-                        A(i, j)
+                        "A",
+                        LogPair(
+                            std::string("(") + std::to_string(i) +"," +
+                            std::to_string(j) + ")",
+                            A(i, j)
+                            )
```
alexbcreiner0, line 341: The logs this produces are of the form `{"t":0,"client":"Society","id":0,"label":"A","(0,1)":0.185964}`. This is a problem - the keys need to be constants. Please separate coordinates and values as two different fields, i.e.
`{"t":0,"client":"Society","id":0,"label":"A","coords": "(0,1)", "value":0.185964}`

src/Society.cpp
```diff
@@ -319,53 +323,48 @@ void Society::retire_person(Person *person) {
...
-            Logger::get_instance()->log(Logger::SOCIETY, "l", id, i, l(i));
+            Logger::log(
+                    Logger::SOCIETY,
+                    id,
+                    "l",
+                    LogPair(std::to_string(i), l(i))
+                    );
         }
     }
 }
```
alexbcreiner0, line 360: Same issue as above for A. Please have it return something like

`{"t":0,"client":"Society","id":0,"label":"l", "prod_id": "5", "value": 0.0532255}`

instead of `{"t":0,"client":"Society","id":0,"label":"l","5":0.0532255}`

src/Firm.cpp
```diff
@@ -409,77 +414,51 @@ void Firm::log_product_quantity(
...
 
 void Firm::log_catalog() {
-    std::vector<int> product_ids;
-    product_ids.reserve(catalog.size());
-
-    for (Product* p : catalog) {
-        product_ids.push_back(p->id);
+    for (Product * product : catalog) {
+        Logger::log(get_client_type(), id, "catalog", LogPair("product_id", product->id));
     }
```
alexbcreiner0, line 486: It doesn't really matter right now, but catalog is a list of product ids. So I think it should be a single key called "product_ids", with the value being a string encoding of a list of integers. Either that, or multiple logged messaged for each new product added or removed from the catalog.

src/Firm.cpp
```diff
@@ -358,26 +358,31 @@ void Firm::log_inventory_reduction(const Product * product, const double quantit
...
-void Firm::log_initial_employment(const int worker_id, const int firm_id) {
-    Logger::get_instance()->log(
+void Firm::log_initial_employment(
+        const unsigned int worker_id,
+        const unsigned int workplace_id
+        ) {
+    Logger::log(
         get_client_type(),
+        id,
         "newly employed",
```
alexbcreiner0: These might be hangovers from my own bad coding, but this should probably be `newly_employed` for consistency with the others (it was like this before though). Also workplace id isn't necessary since its just the same as id.

src/Firm.cpp
```diff
@@ -409,77 +414,51 @@ void Firm::log_product_quantity(
...
             get_client_type(),
-            "accepted_order",
             id,
-            product->product_name,
-            requested_turnaround_time
+            "accepted_order",
+            LogPair("product_id", product->id),
+            LogPair("requested_turnaround_time", requested_turnaround_time)
             );
 }
```
alexbcreiner0: Can this also log the quantity of the order? My Overseer code is acting like that was always what it paired with the product id, which might have been a bug on my end but that would still be the fix.

src/Firm.cpp
```diff
@@ -358,26 +358,31 @@ void Firm::log_inventory_reduction(const Product * product, const double quantit
...
+    Logger::log(
         get_client_type(),
+        id,
         "transfer",
-        worker_id,
-        old_employer_id,
-        new_employer_id
+        LogPair("worker_id", worker_id),
+        LogPair("old_worksplace_id", old_workplace_id),
+        LogPair("new_worksplace_id", new_workplace_id)
```
alexbcreiner0: Typo (although admittedly worksplace is a pretty fun word).

src/Society.cpp
```diff
@@ -319,53 +323,48 @@ void Society::retire_person(Person *person) {
...
 
 void Society::log_io_matrix(Eigen::MatrixXd& A, size_t dim) {
-    Logger::get_instance()->log(Logger::SOCIETY, "A_dim", id, static_cast<int>(dim));
+    Logger::log(Logger::SOCIETY, id, "A", LogPair("dim", static_cast<int>(dim)));
```
alexbcreiner0: This logging of the dimension isn't necessary and causes problems because it uses the same label as the logging of the individual values. Dimension can be inferred in a bunch of ways. Can we get rid of this?

(l has the same issue, it logs dimension with the same event label when it doesn't need to and absolutely shouldn't be using the same event label)

alexbcreiner0 (changes requested):
This is great! Exactly what we needed. Very nice new system. There are some little things I ran into while I was refactoring the visualization layer, none of them should take more than a moment.

src/Society.cpp
Amittai-Aviram: Done:
`{"t":0,"client":"Society","id":0,"label":"A","coords":"(0,6)","value":0.0387628}`

src/Society.cpp
Amittai-Aviram: Done.
`{"t":0,"client":"Society","id":0,"label":"l","prod_id":0,"value":0.0504833}
`

src/Firm.cpp
Amittai-Aviram: Done.
`{"t":0,"client":"Distributor","id":199,"label":"catalog","product_ids":"2,47,75,92,11,18,53,4,1,48,10,34,6,55,36,98,15,79,31,16,72,24,13,17,28,62,41,5,60,82,20,9,8,57,12,21,84,64,88,23,25,90,44,29,30,37,39,19,33,35,45,32,96,3,14,27,40,68,22,42,26,69,65,49,63,80,97,50,89,51,59,46,70,54,56,58,61,38,67,71,77,7,73,43,83,74,78,81,85,76,86,87,93,94,0,91,52,95,66,99"}`

src/Firm.cpp
Amittai-Aviram: Done.
`{"t":502,"client":"Producer","id":81,"label":"newly_employed","worker_id":997}`

src/Firm.cpp
Amittai-Aviram: Done. Note that it now logs both the initial order quantity (as simply "quantity") and the quantity _offered_ by the chosen producer (as "offered_quantity"), because, as far as I can tell, those two numbers may be different. From `Producer::draft_plan_and_return_order`:
```
    int return_order_quantity = std::min(order->quantity, get_max_order_quantity(order->product));
```
Also, I have changed the name of the "requested_turnaround_time" field to "offered_turnaround_time" to reflect its meaning. The ordering firm requests; the upstream producer offers. Here is an example:
```
{"t":489,"client":"Distributor","id":189,"label":"accepted_order","product_id":50,"quantity":1,"offered_turnaround_time":0.327546,"offered_quantity":1}
```

src/Firm.cpp
Amittai-Aviram: Oops!  Done.
`{"t":169,"client":"Producer","id":3,"label":"transfer","worker_id":2,"old_workplace_id":1,"new_workplace_id":3}`

src/Firm.cpp
alexbcreiner0: This is great, thanks! More info is always better.

alexbcreiner0 (approved):

Issue #205, opened 2026-05-16
Make Distributors execute a distribution plan


lennyyyyyyyy:
This includes:
- Making use of `Firm`'s hiring process to hire workers most suitable for distribution work
   - We could consider generating a new set of abilities corresponding for the distribution of each good, but I think it's better to share the same set of abilities for now
- Executing a plan converts non-consumer goods in the input inventory to consumer goods in the output inventory, and represents something like organizing and stocking shelves in the store, working the register, etc.
- A consumer good's price should be updated as the normal good's price with distribution labor added on (this should result from correct plan reporting anyway)
- A Distributor's "pending inventory" should combine the quantities in both its input and output inventories
- For the arbitrary length / quantity of a distribution plan, we can try to make them last around a day, so that the price of ConsumerGoods update at around that frequency.

PR #211, opened 2026-05-24
Reorganize Product hierarchy, generalize Plan incrementation to Firm, etc.

### LINKED ISSUE #202: Recast Product class hierarchy
The classes `Product`, `Machine`, and `ConsumerProduct` need to be revised. Their inheritance relations should be reorganized as follows:
```
                             Product
          _______________________|________________________________________
         |                       |                    |                   |
       Good                 ConsumerGood            Machine            (Service)
```

- A product has
  - A unique ID
  - A type (to avoid the need for RTTI)
  - A map of inputs (input products) per unit
  - A list of machines needed to produce it
  - A list of skills (abilities) needed to produce it
  - Average socially necessary living labor per unit
  - A price per unit.
- A good has, in addition, 
  - A natural manufacturing batch size (?)
- A consumer good is a product with one and only one input product, which must be a good. It is consumed only one at a time, so a firm cannot order a consumer good &mdash; only a consumer can buy a good, one at a time. A consumer good also has
  - A mean consumption period
  - A mean consumption frequency.
- A machine is a product that has
  - A lifetime
  - A reorder threshold (?) &dash; the estimated number of hours that it takes to build the machine. (This should be tracked and updated by the Price Controller.)  The firm orders a replacement for the machine when the number of hours remaining equals the number of hours required to get a replacement, so that work will  not get interrupted.
- A service is a product of which only one at a time is consumed. Unlike a consumer good, a service can be consumed by any person or firm.

### LINKED ISSUE #206: Restrict shopping to hours outside of a Plan
People can currently shop at any time, which is unrealistic. At the very least, they cannot go shopping at the same time they are assigned and working on a Plan, because that would imply that the 1 labor credit per hour they are given is not accurate.

### LINKED ISSUE #213: Simplify Order rejection and dropping
Currently, rejected orders are sent through `draft_plan_and_return_order` with the order status as `Order::REJECTED`. However, the draft plan associated with the rejected order is kept, only to be dropped later on when the downstream Firm realizes it is a rejected order through `drop_order`. 

It's simpler to not store draft plans associated with rejected orders, and also skip the call to `drop_order` later on.

lennyyyyyyyy:
Resolve #202 
Resolve #206 
Resolve #213 
 - Moved over Plan incrementation functions from Producer to Firm, which should facilitate Distributors executing distribution plan as described in #205 
 - Also moved the Society constructor to an initialization function, so that Products, Persons, Firms, etc. can call `Society::get_instance()` in their constructors
 - Allow a variable amount of abilities 
 - Moved `underlying_living_labor_per_unit` to be stored as a field of the Society to avoid confusion


src/ConsumerGood.cpp
```diff
@@ -1,12 +1,34 @@
...
-    price_per_unit = product->price_per_unit + DISTRIBUTION_LABOR_PER_UNIT;
-    mean_consumption_frequency = product->mean_consumption_frequency;
+    Society::get_instance()->
+        set_underlying_living_labor_per_unit(this, living_labor_per_unit);
+    required_abilities = { DISTRIBUTION_ABILITY };
+    static std::uniform_real_distribution<>
+        consumption_freq_dist(0.0, 1.0);
+    mean_consumption_frequency = consumption_freq_dist(Sim::get_random_generator());
+    corresponding_good = good;
+    inputs_per_unit[corresponding_good] = 1.0;
```
Amittai-Aviram: Isn't this redundant with `ConsumerGood::set_inputs` below?

header/ConsumerGood.h
```diff
@@ -2,7 +2,15 @@
...
+struct Good;
+
 struct ConsumerGood : public Product {
-    ConsumerGood(Product * product);
+    ConsumerGood(Good * good);
+    void set_inputs() override; 
+    void set_machines() override;
+    void log_mean_consumption_frequency();
+
+    Good * corresponding_good = nullptr;
```
Amittai-Aviram, line 13: Isn't this redundant with `inputs_per_unit`?  (DId you just feel that readability is worth the duplication?)

src/Distributor.cpp
```diff
@@ -19,28 +20,13 @@ Distributor::Distributor(
...
     Firm(society, initial_catalog)
 {
-    for (Product * product : get_products_to_reorder()) {
-        society->add_consumer_good(product);
-        int quantity =
-            product->mean_consumption_frequency *
+    for (Product * product : catalog) {
+        Good * good = static_cast<ConsumerGood *>(product)->corresponding_good;
+        input_inventory[good] =
+            good->corresponding_consumer_good->mean_consumption_frequency *
```
Amittai-Aviram, line 27: The handling of types here seems hard to follow, but I can see why you are doing things this way. Every product in the catalog is a consumer good, and the good corresponding to each consumer good must be stocked in the input inventory. One thing that could make it clearer would be this, I think:
```
input_inventory[good] = consumer_good->mean_consumption_frequency *
    (FIRM_STOCKPILE_DURATION + FIRM_DEMAND_WINDOW_MIN) *
    (Sim::get_num_people() / Sim::get_num_distributors());
```
Would you agree?

src/Distributor.cpp
```diff
@@ -50,47 +36,25 @@ Logger::Client Distributor::get_client_type() {
...
-        return 0;
-    }
-    add_demand_signal(&product, quantity);
-    check_and_reorder_input(&product);
-    if (!catalog.count(&product)) return 0;
-    int available = std::min(static_cast<int>(get_inventory_level(&product)), quantity);
+int Distributor::try_sell_goods(ConsumerGood * consumer_good, int quantity, Person * person) {
+    add_demand_signal(consumer_good->corresponding_good, quantity);
+    check_and_reorder_input(consumer_good->corresponding_good);
+    if (!catalog.count(consumer_good)) return 0;
```
Amittai-Aviram: How could we have gotten this far without the consumer good being in this distributor's catalog? We have just checked the inventory for its corresponding good and possibly stocked up on it. Is that appropriate if the consumer good is not in the catalog?

As a style point, please give all condition and loop bodies a complete block, even if the body contains only a single statement:
```
if !catalog.count(consumer_good)) {
    return 0;
}
```
(Please do this everywhere.)

src/Distributor.cpp
```diff
@@ -50,47 +36,25 @@ Logger::Client Distributor::get_client_type() {
...
+    int available = std::min(static_cast<int>(
+                get_inventory_level(consumer_good->corresponding_good)), quantity);
     if (available < quantity) {
-        log_shortfall(product.product_name, quantity - available);
+         log_shortfall("", quantity - available);
```
Amittai-Aviram: Yeah, this needs to be updated according to the latest changes to the Logger. You'll see those when you resolve merge conflicts.

header/Firm.h
```diff
@@ -11,12 +11,12 @@
 #include "Logger.h"
 #include "Person.h"
 
-class Firm;
 struct Machine;
 struct Order;
+struct Product;
+class Firm;
```
Amittai-Aviram, line 18: Do you need this to compile `Plan`?

header/Firm.h
```diff
@@ -118,14 +125,18 @@ class Firm : public Agent {
         Plan * draft_plan
     );
     double calculate_machinery_cost_for_plan(Plan * draft_plan);
-	void assign_plan_dependent_fields(Plan * draft_plan, std::vector<Person::Ability>& required_abilities);
+	void assign_plan_dependent_fields(Plan * draft_plan,
+            std::vector<int>& required_abilities);
```
Amittai-Aviram: Our preferred style is as follows:
```
void assign_plan_dependent_fields(
    Plan * draft_plan,
    std::vector<int>& required_abilities
);
```

header/Firm.h
```diff
@@ -118,14 +125,18 @@ class Firm : public Agent {
...
     void add_demand_signal(Product * product, double quantity);
-    Plan * draft_plan_with_required_abilities(Order * order, std::vector<Person::Ability>& required_abilities); 
+    Plan * draft_plan_with_required_abilities(Order * order,
+            std::vector<int>& required_abilities);
```
Amittai-Aviram: See above.
```
Plan * draft_plan_with_required_abilities(
    Order * order,
    std::vector<int>& required_abilities
);
```

src/Society.cpp
```diff
@@ -217,49 +211,51 @@ void Society::set_product_prices_production_consumption() {
...
-        product->mean_consumption_frequency *= consumption_scalar;
-    }
-    for (Product *product : products) {
-        product->mean_consumption_period = static_cast<int>(std::ceil(1 / product->mean_consumption_frequency));
+    consumption_scalar = PRODUCT_CONSUMPTION_MULT
+        * initial_work_week
+        / WEEK
+        / consumption_scalar;
+    for (ConsumerGood * consumer_good : consumer_goods) {
+        consumer_good->mean_consumption_frequency *= consumption_scalar;
```
Amittai-Aviram:
```
double value_consumed_per_hour = 0.0;
for (ConsumerGood * consumer_good : consumer_goods) {
    value_consumed_per_hour += consumer_good->price_per_unit *
        consumer_good->mean_consumption_frequency;
}
double worked_portion_of_week = initial_work_week / WEEK;
double consumption_scalar =
    (worked_portion_of_week * PRODUCT_CONSUMPTION_MULT) /
    value_consumed_per_hour;
for (ConsumerGood * consumer_good : consumer_goods) {
    consumer_good->mean_consumption_frequency *= consumption_scaling_factor;
}
```
Put this code into a _separate function_ — `static void adjust_consumption_frequencies`.
NB: I have changed my mind about _scalar_ &mdash; you use it in one other place that I could see, so I guess it acquires the weight of convention &mdash; so I am OK with it. If you do change this to `consumption_scaling_factor`, please also change the other instance of _scalar_ to _scaling\_factor_.

header/Good.h
```diff
@@ -0,0 +1,12 @@
+#pragma once
+
+#include "Product.h"
+
+struct ConsumerGood;
+
+struct Good : public Product {
+    Good();
+    void set_corresponding_consumer_good(ConsumerGood * consumer_good);
```
Amittai-Aviram: Since this is a structure type, maybe we should  not bother with this? It could make the client code a bit tider by just using an assignment.

src/Person.cpp
```diff
@@ -35,9 +39,9 @@ Person::Person(Society * society):
             ranked_distributors.end(),
             Sim::get_random_generator()
             );
-    for (Product * product : society->get_goods()) {
-        inventory[product] = 
-            (int) (PERSON_STOCKPILE_DURATION * product->mean_consumption_frequency);
+    for (ConsumerGood * consumer_good : society->get_consumer_goods()) {
+        inventory[consumer_good] = 
+            (int) (PERSON_STOCKPILE_DURATION * consumer_good->mean_consumption_frequency);
```
Amittai-Aviram: Could you please change this to use `std::static_cast<int>`?

src/Person.cpp
```diff
@@ -48,15 +52,15 @@ unsigned int Person::get_id() {
     return id;
 }
 
-std::unordered_map<Person::Ability, double>& Person::get_abilities() {
+std::unordered_map<Ability *, double>& Person::get_abilities() {
     return this->abilities;
```
Amittai-Aviram: Just wondering: will this not work without the `this` explicit pointer? If it will, could you please remove it?

src/Person.cpp
```diff
@@ -48,15 +52,15 @@ unsigned int Person::get_id() {
...
 }
 
 double Person::get_busyness() {
     return busyness;
 }
 
-void Person::train(std::unordered_map<Person::Ability, double> target_abilities) {
+void Person::train(std::unordered_map<Ability *, double>& target_abilities) {
     // can introduce < 100% effectiveness on training later
     for (auto &pair : target_abilities) {
```
Amittai-Aviram: Please replace `auto` with the actual type.  (Then again, I think Nathan will replace it soon anyway.) But also please change the name of the variable from `pair` to `ability_level`.

src/Person.cpp
```diff
@@ -93,47 +97,38 @@ float Person::productivity() {
...
+        to_consume[consumer_good] += consumer_good->mean_consumption_frequency;
+        int consumed = static_cast<int>(to_consume[consumer_good]);
         if (consumed) {
-            inventory[product] -= consumed;
-            log_consumption(product, consumed);
+            inventory[consumer_good] -= consumed;
+            log_consumption(consumer_good, consumed);
         }
-        to_consume[product] -= (int) to_consume[product];
+        to_consume[consumer_good] -= (int) to_consume[consumer_good];
```
Amittai-Aviram, line 121: I'm sorry — I know we went over this before, but — why do we subtract the floor of this value from the value itself regardless of whether we have consumed anything? I would have expected this statement to go within the if-condition body: for each consumer good, first add to its `to_consume` value its mean consumption frequency. If that adds up to an integer value, consume that much — and take that much away from the `to_consume` value. (Otherwise, just leave that increased value for the next time around.)

src/Person.cpp
```diff
@@ -93,47 +97,38 @@ float Person::productivity() {
...
-//         int period = product->mean_consumption_period;
-//         if (time % period == 0) {
-//             inventory[product] -= 1;
-//             log_consumption(product, 1);
-//         }
-//     }
-// }
-
 bool Person::will_shop() {
+    if (busyness_this_time_step) return false;
```
Amittai-Aviram: Please revise for style.

src/Person.cpp
```diff
@@ -209,9 +203,10 @@ void Person::update_busyness() {
 void Person::on_time_step() {
 	++age;
     consume();
-	if (will_shop()) { shop(); }
 	if (will_retire()) { retire(); }
```
Amittai-Aviram: Style.

src/Person.cpp
```diff
@@ -209,9 +203,10 @@ void Person::update_busyness() {
...
 	update_health_status();
+
+	if (will_shop()) { shop(); }
```
Amittai-Aviram: Style.

header/Producer.h
```diff
@@ -26,23 +26,12 @@ class Producer : public Firm {
...
-    bool is_within_work_schedule() const;
-
-	void start_plan(Plan * plan);
-	void move_plan_forward_one_step(Plan * plan);
-	void end_plan(Plan * plan);
-	void move_plans_forward_one_step();
-    std::unordered_set<Product *> get_products_to_reorder() override;
-    void log_plans();
     void log_draft_plan(const Plan * draft_plan);
     void log_dropped_order(const Order * order);
```
Amittai-Aviram, line 36: Could the logging functions also be taken out?

src/Product.cpp
```diff
@@ -63,7 +54,8 @@ void Product::set_inputs(std::vector<Product *>& goods) {
     }
 }
 
-void Product::set_machines(std::vector<Machine*> machines) {
+void Product::set_machines() {
+    std::vector<Machine *> machines = Society::get_instance()->get_machines();
     if (!machines.size()) return;
```
Amittai-Aviram: Style.

Amittai-Aviram (changes requested):
This is a great set of changes! The code base will benefit enormously from it. I am asking for just a few changes.

src/ConsumerGood.cpp
lennyyyyyyyy: Yeah, but set_inputs requires an override, since it's a pure virtual function, and this seemed like the best thing to do.

header/ConsumerGood.h
lennyyyyyyyy: Yup. Avoids `(*inputs_per_unit.begin()).first`, which is a lot less readable.

src/Distributor.cpp
lennyyyyyyyy: Done.

src/Distributor.cpp
lennyyyyyyyy: The check probably isn't necessary, but I guess it's just preventative. Fixed the style.

src/Distributor.cpp
lennyyyyyyyy: Resolved.

header/Firm.h
lennyyyyyyyy: Yeah. Should be unnecessary once we move `Plan` and `Order` to their own files.

header/Firm.h
lennyyyyyyyy: Fixed.

header/Firm.h
lennyyyyyyyy: Fixed.

header/Good.h
lennyyyyyyyy: Done.

src/Person.cpp
lennyyyyyyyy: Done.

src/Person.cpp
lennyyyyyyyy: Done. Don't know how that happened.

src/Person.cpp
lennyyyyyyyy: Done.

src/Person.cpp
lennyyyyyyyy: Done.

src/Person.cpp
lennyyyyyyyy: Done.

header/Producer.h
lennyyyyyyyy: These logs are specific to Producer. The other log removed is now moved to Firm.

src/Product.cpp
lennyyyyyyyy: Done.

src/Person.cpp
lennyyyyyyyy: Done

src/Person.cpp
lennyyyyyyyy: I think it was just a decision for consumption not to carry over if they weren't able to in the current time step. I could change it, but this affects the simulation more when it's crashing than when it's stable.

src/Society.cpp
lennyyyyyyyy: Done.

src/ConsumerGood.cpp
Amittai-Aviram: You could just take it out of the constructor and keep it in `set_inputs`.  I think that is probably the clearest and cleanest solution.  You set `corresponding_good` in the constructor, and then `set_inputs` sets a unit of that as the only input. No?

header/ConsumerGood.h
Amittai-Aviram: Yeah, it's good, especially, I think, if you take that `inputs_per_unit` initialization out of the constructor. The effect is that `corresponding_good` is the "ground" datum, from which the `inputs_per_unit` entry is _derived_.

src/Distributor.cpp
Amittai-Aviram: My point re the check is that either it should happen before _anything_ that would touch the consumer good on the assumption that it is in the catalog or it should  not happen at all. You should not check on a condition that you would have needed for previous steps to succeed. Do you see what I mean?

header/Good.h
Amittai-Aviram: (I meant "tidier." I seem to have slightly sticky keys.)

src/Person.cpp
Amittai-Aviram: I think you can actually use a reference here:
`for (const std::pair<Ability * const, double>& ability : target_abilities) {`

src/Person.cpp
Amittai-Aviram: I think I need to discuss this with you for a couple of minutes to make sure I understand it before I sign off on it. I will try to catch you at a free moment.

Amittai-Aviram (changes requested):
Looking great! I asked for just a few more changes — and I would like to discuss the consumption thing before signing off.

src/Distributor.cpp
lennyyyyyyyy: Oh, actually I think it is necessary. Person calls `try_sell_goods` sequentially down its list of ranked distributors, which may or may not have the requested consumer good in its catalog. I'll just move this to the beginning of the method so it doesn't cause any unnecessary reorder checks.

src/ConsumerGood.cpp
lennyyyyyyyy: Sure, we'll just have to make sure we always call `set_inputs` on the consumer goods after their construction then.

Amittai-Aviram: I think that's a reasonable assumption.

src/Distributor.cpp
Amittai-Aviram: Yeah, that's exactly the solution I was looking for. Thanks.

Amittai-Aviram (approved):
LGTM
Looks great now!
I do not understand why we need to log the zeroes in the labor vector, but perhaps you can explain that to me separately.

PR #212, opened 2026-05-27
Fix null keys, segfaults, and division-by-zero cases

### LINKED ISSUE #191: Conduct a project-wide safety audit to guard against overflow errors and the like
Since my more varied consumption frequencies were merged into main and especially with Devin's most recent PR, the simulation is coming under more numerical stress than before, and I think this is one of the main reasons why it appears less stable than it did previously. I don't have any concrete examples to offer, but I think we are just generally due for someone to carefully comb through the code and make sure that, at the very least, there are no division operations that run any risk of causing blow ups or overflow errors. This is something that has caused a ton of problems for me over the course of building my own models. 

More specifically, one or more minimum 'safeguard sizes' $\sigma$ should be introduced as a constant(s) in `constants.h`, something like `0.005` , and then whenever we find ourselves dividing a number $c$ by any _variable_ $x$ within the simulation, we should instead divide $c$ by the _maximum_ of $x$ and $\sigma$. More generally, any potential _vertical asymptotes_ needs to be guarded against. 

This seems like a great job for Nathan to do, since it requires taking a broad but attentive overview over the code which he is already doing. So I'm going to assign it to him. That said, everyone should be chipping in a bit here, and vigilant for reasons why the sim is randomly crashing.

### LINKED ISSUE #216: Make the `ArgType` scoped enumeration type in `main.cpp` conform to Google C++ style standards
According to [the latest Google standards](https://google.github.io/styleguide/cppguide.html#Enumerator_Names):
- The type name itself should be in **UpperCaseCamelCase**: `enum class ArgType`.
- Each enumerator should be in camel case (as now), but prefixed with a lower-case **k**: `kPeople`, `ArgType::kPeople`; `kWorkHours`, `ArgType::kWorkHours`; etc.

nathanjlim:
Resolves #191
Resolves #216 

- Explicit statement checkers on maps: Distributor.cpp & Firm.cpp
- Division-by-zero guards: Firm.cpp & PriceController.cpp
- Replaced auto with explicit std::pair in Person.cpp


src/Firm.cpp
```diff
@@ -94,13 +94,14 @@ double Firm::get_busyness() {
 std::vector<Person *> Firm::propose_transfer(int workers_wanted) {
     double firm_busyness = get_busyness();
     double societal_busyness = society->get_busyness();
-    int max_workers_to_transfer = (int) (workers.size() * (1.0 - firm_busyness / 
-            (societal_busyness - TRANSFER_BUSYNESS_THRESHOLD))); 
-    max_workers_to_transfer = std::max(max_workers_to_transfer, workers_wanted);
-    log_busyness(firm_busyness, societal_busyness, max_workers_to_transfer);
-    if (firm_busyness >= societal_busyness - TRANSFER_BUSYNESS_THRESHOLD) {
+    double gap = societal_busyness - TRANSFER_BUSYNESS_THRESHOLD;
```
Amittai-Aviram: This is good, but let's change the new variable's name. It represents an adjustment on the society's overall busyness. The idea is that the maximum number of idle workers that a firm can transfer _out_ to meet the needs of some other firm is a little smaller, by a fudge factor, than the number left after you take out the portion of busy workers in all of society that is working specifically at this firm. (At least I think that's the idea.) By making that exemption from transfer a little bigger than the actual number of workers who are busy, we give the firm a little elbow room from plan to plan, to avoid too many transfers out and then back in. So I think I'd propose `adjusted_societal_busyness` instead of `gap`.

src/Firm.cpp
```diff
@@ -153,13 +154,19 @@ double Firm::get_reorder_threshold(Product * product) {
 }
 
 double Firm::get_pending_inventory_level(Product * product) {
-    double pending_inventory = input_inventory[product];
-    for (Order * order : product_to_outbound_orders[product]) {
-        pending_inventory += order->quantity;
+    double pending_inventory = 0.0;
+    if(input_inventory.count(product)) {
```
Amittai-Aviram: Style: `if (input_infentory.count(product)) {`

src/Firm.cpp
```diff
@@ -153,13 +154,19 @@ double Firm::get_reorder_threshold(Product * product) {
...
+        pending_inventory = input_inventory[product];
+    }
+    if(product_to_outbound_orders.count(product)) {
```
Amittai-Aviram: Style. (See above.)

src/Firm.cpp
```diff
@@ -153,13 +154,19 @@ double Firm::get_reorder_threshold(Product * product) {
...
+        for (Order * order : product_to_outbound_orders[product]) {
+            pending_inventory += order->quantity;
+        }
     }
     return pending_inventory;
 }
 
+
```
Amittai-Aviram: Please remove. Only one vertical space between every pair of functions.

src/Firm.cpp
```diff
@@ -229,6 +236,11 @@ void Firm::assign_workers(
 }
 
 double Firm::predict_turnaround_time(Plan * plan, std::vector<Person *>& workers) {
+
```
Amittai-Aviram: Remove space.

src/Firm.cpp
```diff
@@ -229,6 +236,11 @@ void Firm::assign_workers(
...
+    double count = safe_handle_zero(static_cast<double>(workers.size()), "predict_turnaround_time workers", 0.0);
+    if (count == 0) {
+        return 0;
```
Amittai-Aviram: If you look at the larger context, I think the point should be that, if the firm has no workers, then its predicted turnaround time for an order should be impossibly long, i.e., infinite. So you can return the special `double` value `std::numeric_limits<double>infinity(). Now, as it happens, the standard result of division by `0.0` in the `double` type is, indeed, infinity. (Please verify this.) Moreover, it may not quanlify as a system error if the firm cannot allocate any workers for the plan.

The (indirect) caller, `Producer::draft_plan_and_return_order`, rejects the order if it cannot assign any workers to the plan for it. However, that happens _after_ it attempts to draft the plan, through the call to `Firm::draft_plan_with_required_abilities`, which calls `Firm::assign_workers` and then `Firm::assign_dependent_fields`, the latter of which then calls `Firm::predict_turnaround_time`. So you can avoid this whole problem by checking whether `assign_workers` has assigned any workers. Then, if not, perhaps return `nullptr` from `Firm::draft_plan_with_required_abilities`. Finally, back in `Producer::draft_plan_and_return_order`, instead of checking whether the plan has any workers, just check whether the plan exists at all:
```
if (!draft_plan) {
    return_order->status = Order::ORDER_REJECTED;
}
```
But that cannot be the end of the story, because, as things stand now, we then use the plan's predicted turnaround time, and then store the plan. I don't know why we do either of those things if the order is rejected — I have just asked Devin. If returning `nullptr` turns out not to work, for whatever reason we learn from Devin (or figure out with further study), then setting predicted turnaround time to infinity should be a reasonable option.

You _do_ have to worry about integer division by 0. But, if it involves doubles, returning the infinity explicitly is not a bad way of documenting the behavior.

On a side note, the style that I prefer is `if (!count)` or `if (!plan)` rather than, respectively, `if (count == 0)` or `if (plan == nullptr)`.

src/Firm.cpp
```diff
@@ -238,6 +250,11 @@ double Firm::predict_turnaround_time(Plan * plan, std::vector<Person *>& workers
 }
 
 double Firm::predict_labor_hours(Order * order, std::vector<Person *>& workers) {
+
```
Amittai-Aviram: Delete.

src/Firm.cpp
```diff
@@ -238,6 +250,11 @@ double Firm::predict_turnaround_time(Plan * plan, std::vector<Person *>& workers
...
+    double count = safe_handle_zero(static_cast<double>(workers.size()), "predict_labor_hours workers", 0.0);
+    if (count == 0) {
+        return 0;
```
Amittai-Aviram: There's something similar going on here. Basically, it would be better to change the logic so that the 0 never happens in the first place, and then, perhaps, to throw an exception if it leaks thorugh, as a way of documenting, in effect, that the 0 should never occur.

src/Firm.cpp
```diff
@@ -270,6 +287,10 @@ double Firm::calculate_machinery_cost_for_plan(Plan * draft_plan) {
     for (Machine * machine : machines) {
         machinery_cost_per_hour += machine->price_per_unit / machine->lifetime;
     }
+    double count = safe_handle_zero(static_cast<double>(draft_plan->workers.size()), "machinery_cost workers", 0.0);
+    if (count == 0) {
```
Amittai-Aviram: See comments above.

src/Firm.cpp
```diff
@@ -327,12 +348,16 @@ void Firm::apply_demand_window() {
...
     int window_start = Sim::get_current_time_step();
-    if (!demand_signals[product].empty()) {
+    if (demand_signals.count(product) && !demand_signals[product].empty()) {
         window_start = demand_signals[product].front().timestep;
     }
     int window_length = std::max(FIRM_DEMAND_WINDOW_MIN, 
         Sim::get_current_time_step() - window_start);
-    return total_demands[product] / window_length;
+    double total_d = 0;
+    if(total_demands.count(product)) {
```
Amittai-Aviram: Style.

src/Firm.cpp
```diff
@@ -327,12 +348,16 @@ void Firm::apply_demand_window() {
...
+        total_d = total_demands[product];
+    }
+    return total_d / safe_handle_zero(static_cast<double>(window_length), "get_demand window", 1.0);
```
Amittai-Aviram: I think the upstream logic here should ensure that window_length is at least 1. Please check. If that's the case, then just throw an exception if it's below 1.

src/PriceController.cpp
```diff
@@ -34,6 +34,16 @@ void PriceController::update_price(Plan * plan) {
         hours += plan->raw_materials - plan->raw_materials_remaining;
         workers += plan->workers.size();
     }
+    if(units <= 0) {
```
Amittai-Aviram: Style.
Just throw an exception. The units in a plan should never be 0 or less.

src/PriceController.cpp
```diff
@@ -34,6 +34,16 @@ void PriceController::update_price(Plan * plan) {
...
+        std::cerr << "PriceController units ran 0 or negative for product: " 
+                  << product->product_name << ". Using safeguard: 1\n";
+        units = 1;
+    }
+    if(workers <= 0) {
```
Amittai-Aviram: STyle: `if (...)`
Same issue here — just throw an exception. A plan cannot be completed without workers. (if we ever come up with a notion of a plan with no workers, we shall run into that exception, which will ensure that we think through all the appropriate changes.)

Amittai-Aviram (changes requested):
Great start!
There's one point where we need some information from Devin — or, if you would, please study the code to see whether you can figure it out yourself. In general, I think I'd prefer to prevent division by 0 with _upstream_ changes, so that, then, when you get to the division, you can just throw an exception if you encounter a 0 at that point, since we have every reason to expect that it will not occur.

Amittai-Aviram (approved):
LGTM

PR #215, opened 2026-05-29
Create flags for abilities and ability-stddev

### LINKED ISSUE #214: Add num_abilities and ability-stddev flags for data aggregation layer
Overseer has a control for ability standard deviation, yet the C++ doesn't have a flag for this yet. Additionally, since changing the number of abilities to be variable, we have not yet added corresponding flags and controls in the LTE data aggregation layer yet. I'll create a corresponding issue there, as well as make it compatible with the most recent version of main.

lennyyyyyyyy:
Resolve #214


src/main.cpp
```diff
@@ -15,6 +15,7 @@ void print_usage() {
     std::cout << "\t-m N: Set the initial number of products per machine to N." << std::endl;
     std::cout << "\t-r N: Set the initial number of producers to N." << std::endl;
     std::cout << "\t-d N: Set the initial number of distributors to N." << std::endl;
+    std::cout << "\t-a N: Set the number of abilities to N." << std::endl;
```
Amittai-Aviram: I've always been troubled by the likely opacity of the word _abilities_ for anybody not already intimately familiar with our project. We should (soon) explain it and other details in the README, but, for now, how about this? —
`Set the number of modeled abilities (work skills) to N.`

src/Sim.cpp
```diff
@@ -45,6 +45,10 @@ unsigned int Sim::get_num_distributors() {
     return get_instance().args.num_distributors;
 }
 
+unsigned int Sim::get_num_abilities() {
+    return get_instance().args.num_abilities;
+}
+
```
Amittai-Aviram, line 51: Could you please go ahead and implement a `get_abilities_stddev`? It should be safe, because it's currently set to the default in the `SImArgs` definition, right? That way, you have everything already set up except the CLI for when you have added that parameter to Overseer's data aggregation.

Amittai-Aviram (changes requested):
Looks great! I am asking for a couple of smallchanges.
In addition, we need to change the style of the type name and the enumerator names in the `argType`, but I should probably ask Nathan to do that. (Unfortunately, I did  not catch that style issue in reviewing his code. I think I was in a big rush at the time.)

src/main.cpp
lennyyyyyyyy: Done.

src/Sim.cpp
lennyyyyyyyy: Done.

src/main.cpp
```diff
@@ -49,103 +51,116 @@ void set_params(int argc, const char ** argv, SimArgs& args) {
...
-                        args.ability_stddev = value;
-                        break;
-                    }
-                    case argType::Seed: {
-                        args.seed = value;
-                        args.fixed_seed = true;
-                        break;
-                    }
+            case argType::People: {
+                args.num_people = value;
```
Amittai-Aviram: Could you please do the check and error setting before the parameter setting, consistently in each case, and set the parameter only if the check passes (with `else`)? —
```
case argType::People: {
    if (value <= 0) {
        error = true;
    } else {
        args.num_people = value;
    }
}
````
So the `args` object will hold the default value. It will not affect behavior, since we do not run the simulation if the argument value is invalid, but this might avoid some surprising overflow effects, such as a huge value for `num_people`, where the user had entered a negative value.  (The developer would then have to check the actual value of `value`, which is what should happen.)  It also just seems more logically consistent this way, I think.

Amittai-Aviram (changes requested):
Almost there.

src/main.cpp
lennyyyyyyyy: Done.

Amittai-Aviram (approved):
LGTM

PR #220, opened 2026-05-31
Replace firm-level supplier lists with a society-wide map

### LINKED ISSUE #217: Create a global map of suppliers to products
Currently, each firm has its own vector of suppliers (producers). In `Firm::send_order`, the ordering firm searches through its suppliers for the one that (a) has the required product in its catalog and (b) offers the best deal (with the best ratio of output to turnaround time). But the `suppliers` vector is populated by the society, after all the firms are created, as simply all producers &mdash; so every firm's list of suppliers is a copy of the list of all producers. This is a wasteful O(n^2) operation with a lot of data duplication. It is also wasteful, in the ordering sequence, to iterate over all producers, including those that do not produce the required product.

Instead, `Society` should have a map from products to the producers that produce them.

Amittai-Aviram:
Replace the vector of producers that every firm has, over which it iterates in looking for a supplier to which to send an order for an input product, with a global map from product to a vector of firms that produce that product. The product in question may be a machine or a good, but  not a consumer good.

Request to the reviewer: please try building and running the code. If you have Overseer set up, please also check the results in Overseer to make sure that they are comparable to those obtained without these changes. Thanks!

Resolves #217


src/Society.cpp
```diff
@@ -38,22 +38,25 @@ Society::Society() :
 void Society::initialize() {
     set_abilities();
     set_initial_products();
+    unsigned int num_products = Sim::get_num_products();
     for (unsigned int i = 0; i < Sim::get_num_producers(); i++) {
-        Producer * producer = new Producer(this, {goods[i % Sim::get_num_products()]});
+        Product * product = products[i % num_products];
```
lennyyyyyyyy: I think this has the same behavior, since goods are stored from indices 0 to num_products - 1, but you should change products to goods just to be safe (since consumer goods are also in products, but they're handled by Distributors). 

This is my bad, but the naming of the constants is also a little confusing - num_products is actually the number of goods, and the total number of products is double this plus the number of machines. If you could do a naming change from num_products to num_goods, that'd also be great.

lennyyyyyyyy (changes requested):
One small thing.

src/Society.cpp
Amittai-Aviram: I've changed this code so that it uses `products.size()`, and I have changed the constant names to refer to GOOD instead of PRODUCT. However, this led me to change the corresponding symbols in `Sim` and `main`, including the flag for the  number of goods (formerly "products") from **-r** to **-g**. Does anything need to change now on the Overseer side because of this CLI change?

src/main.cpp
```diff
@@ -13,7 +13,7 @@ void print_usage() {
     std::cout << "\t-w N: Set the work week to N days." << std::endl;
     std::cout << "\t-o N: Set the initial number of products to N." << std::endl;
     std::cout << "\t-m N: Set the initial number of products per machine to N." << std::endl;
-    std::cout << "\t-r N: Set the initial number of producers to N." << std::endl;
+    std::cout << "\t-g N: Set the initial number of producers to N." << std::endl;
```
lennyyyyyyyy: I think you changed the wrong line here. There should be a change on line 14 from -o to -g.

src/Society.cpp
lennyyyyyyyy: I think you mean from -o to -g. It's a small change on the Overseer side, so you should just let Alex know.

I think line 43 should really be `Product * product = goods[i % get_num_goods()];`. Wouldn't the current version create Producers for consumer goods?

lennyyyyyyyy (changes requested):

src/Society.cpp
Amittai-Aviram: Yeah, sorry — from -o to -g. Re goods — ugh! — yeah, thanks. Because we currently store the consumer goods right after the goods, we've got a problem. I want some producers to produce machines. I'll make some changes accordingly and get back to you.

src/main.cpp
Amittai-Aviram: Sorry. Fixed.

src/Society.cpp
Amittai-Aviram: I've made several more changes in order to support the assignment of both goods and machines to producers' catalogs. I've also changed the initialization code in Society to accommodate cases where there are more of such upstream products (as I am calling them) than producers, or more producers than products.  In this set of changes, I've moved the initialization of the catalog out of the `Producer` constructor, and made it optional for the `Firm` constructor. For producers, you insert products into the catalog with `add_to_catalog` after construction. The optional `initial_catalog` parameter to the `Firm` constructor is still there to accommodate the construction of distributors as we currently have it.

src/Firm.cpp
```diff
@@ -32,11 +32,10 @@ Order::Order(
...
 Firm::Firm(
-        Society * society,
-        const std::unordered_set<Product *>& initial_catalog
-        ) :
-    society{society},
-    catalog(initial_catalog)
+       Society * society,
+       const std::unordered_set<Product *>& initial_catalog
+       ) :
+    society{society}, catalog{initial_catalog}
```
lennyyyyyyyy: Instead of putting the initial_catalog in the initialization list, could you call `add_to_catalog` on each element. This keeps  initial inventories working in the Producer and Distributor overrides.

src/Firm.cpp
lennyyyyyyyy, line 607: Can you change this to `log_catalog_addition(Product *)`, since we now add to the catalog individually. Call this method in `add_to_catalog`.

src/Producer.cpp
```diff
@@ -12,11 +12,17 @@
 #include "Sim.h"
 #include "Society.h"
 
-Producer::Producer(
-        Society * society,
-        const std::unordered_set<Product *>& initial_catalog
-        ) :
-    Firm(society, initial_catalog) {
+Producer::Producer(Society * society) : Firm(society) {}
```
lennyyyyyyyy: Why don't we just the same default as Firm (initial_catalog = {}) and pass both into the base constructor?

lennyyyyyyyy (changes requested):
Producer initial inventories aren't being set, and some logs are now missing.

src/Firm.cpp
Amittai-Aviram: The initial inventories were not getting set because I had not updated the code for `add_to_catalog`, which was my bad: I had just transferred the code from the old Producer constructor without reading it carefully. (Sorry!) Do you still want me to change how the Distributor adds to its catalog? My idea was that, in the case of the Distributor, you get the whole catalog all at once.

src/Producer.cpp
Amittai-Aviram: Because I think it's unlikely that we would go back to constructing a producer with an already-set catalog. I thik the case for distributors is rather different, so the constructor signatures are different. We can change this back later if necessary, but I wanted the Firm constructor to accommodate these two different derived types and thus two different approaches to construction.

src/Firm.cpp
Amittai-Aviram: I have actually _added_ a _new_ `log_catalog_addition`, which `add_to_catalog` calls at the end. This way, the Distirbutor constructor (and perhaps other functions) can still log the whole catalog at once.

Amittai-Aviram:
The failure of the inventories to get initialized resulted from bad coding in `Producer::add_to_catalog`. I think this is fixed now — the JSON trace is showing the inventories. I can also see the individual additions to the catalog for producers (and not distributors).

src/Firm.cpp
lennyyyyyyyy: This is fine, but I was thinking of a more general system where you can pass an initial_catalog or call add_to_catalog on Producer or Distributor. Since you already have a virtual add_to_catalog, in the Firm constructor all you'd need is:
```
for (Product * product : initial_catalog) {
   add_to_catalog(product);
}
```
and in add_to_catalog(Product * product):
```
catalog.insert(product);
log_catalog_addition(product);
```
It just seems to me like having a `virtual void add_to_catalog` in the Firm implies that both Producers and Distributors should be able to call it and not have anything go wrong, however Producers are able to call it, but not set an initial inventory, and Distributors can only set an initial inventory, but not call add_to_catalog.

This change would also remove the need for log_catalog.

src/Producer.cpp
lennyyyyyyyy: Sort of the same thing, why make `add_to_catalog` a virtual void method in Firm then? If its specific to Producer, it should be in Producer only.

src/Firm.cpp
Amittai-Aviram: OK, I think I have more or less followed your suggestions here. Both `Producer` and `Distributor` can accept both an initial catalog in the constructor and individual additions to the catalog. I could not define `Firm::add_to_catalog` as a pure virtual function, because that puts `Firm`'s construction into question. (You get a compiler warning) &mdash; but, in principle, I suppose, nobody should ever call `Firm::add_to_catalog` &mdash; that call should always be routed to the derived class's specific behavior. So it might be reasonable to throw an exception in that function's body instead of inserting the product into the catalog.  (But I have done the latter, at least for now.)

I have left the definition of `Firm::log_catalog` in place for now, though nobody is calling it (so it is dead). I want to bring it up with Alex to see whether Overseer needs it. We may have ot put those calls back in. If not, I can remove the definition in another PR.

Amittai-Aviram: BTW, could you please mark conversations as resolved if they are indeed resolved in your view? Thanks!

src/Producer.cpp
Amittai-Aviram: Addressed now, I think.

lennyyyyyyyy (approved):
Results look good!
Image

PR #225, opened 2026-05-31
added new initial setting, tweaked IO matrix generation, fixed flaw in pricecontroller

### LINKED ISSUE #221: Add option to start sim off with equilibrium prices instead of values as initial prices (also Scenarios more generally)
# Actual Issue
Right now, the initial prices are simply set at their values. This does not adequately test our model's ability to assert equilibrium and manage the tumult of post-revolution. A much better alternative would be to set them at their _equilibrium prices of production_. 

In the standard Sraffian/Leontief models of capitalism, we have an input/output matrix $A$, a living labor vector $\mathbf{l}$, and a fixed hourly real wage vector $\mathbf{b}$. $A$ and $\mathbf{l}$ exactly the same as what is in our model. Workers are all assumed to purchase goods in an identical manner using their wages, and $\mathbf{b}$ is the bundle of goods that a worker purchases each hour using an hour's wages. The equilibrium profit rate (EPR) prices are the unit prices such that if capitalists charged these prices for their goods, they would all enjoy identical profit rates. Let $\pi$ denote the EPR (so it is a variable, not the number 3.14...), and $\mathbf{p}$ denote the EPR price vector. In order to produce a unit of commodity $i$, we must purchase the necessary inputs $a_{1i}, a_{2i}, \ldots$, and pay for each input amount:

$$ p_1a_{1i} + p_2a_{2i} + \ldots + p_n a_{ni} = (A^T\mathbf{p})_i $$

Thus for any unit price vector, the matrix product $A^T\mathbf{p}$ represents the vector of costs for the materials needed to produce a unit of each commodity type. Additionally, we must pay for the labor costs. With the real wage bundle fixed, this is equivalent to purchasing a second bundle of inputs, albeit indirectly through paying the worker. If $l_i$ hours are required to produce a unit of commodity $i$, and the worker earns $b_1$ units of commodity 1 for each hour worked, $b_2$ units of commodity 2 for each hour worked, and so on, then the total labor cost required to produce a unit of $i$ is

$$ l_ib_1p_1 + l_ib_2p_2 + \ldots + l_ib_np_n $$

It's not as easy to see this, but the reader can check that this is the $i^{th}$ entry of the matrix product $(\mathbf{l}\mathbf{b}^T)\mathbf{p}$. $\mathbf{l}\mathbf{b}^T$ is called the outer product of $\mathbf{l}$ with $\mathbf{b}$. Thus the vector of unit costs of production for each commodity type is given:

$$ (A^T + \mathbf{l}\mathbf{b}^T)\mathbf{p} $$

Typically, the matrix $A + \mathbf{l}\mathbf{b}$ is denoted $M$ and called the *augmented requirements matrix*. The matrix in the expression above is the transpose of this matrix, so the expression can be rewritten more compactly as $M^T\mathbf{p}$.  If $\mathbf{p}$ is the EPR price vector, then all capitalists earn an equal return on their investment, $\pi$. Thus prices are given by scaling up all costs equally by this amount. This gives us the EPR price equation:

$$ \mathbf{p} = (1+\pi)M^T\mathbf{p} $$

What this implies is that $\mathbf{p}$ is actually an eigenvector of $M^T$, with associated eigenvalue $\frac{1}{1+\pi}$. There are a couple of other requirements which are implicit about $\pi$ and $\mathbf{p}$:
1. $\mathbf{p}$ is strictly positive. So the eigenvalue $r = \frac{1}{1+\pi}$ needs to have at least one vector which is positive in all entries. 
2. 
 $$r = \frac{1}{1+\pi} \Leftrightarrow \pi = \frac{1}{r}-1 \Leftrightarrow r < 1$$

Now, I apologize in advance for introducing this, but the *spectral radius* of a matrix $B$, denoted $\rho(B)$, is the absolute value of the largest eigenvalue of $B$. Since the eigenvalues of a matrix can be real or complex, it need not be the case that a matrix has an eigenvalue which actually equals it spectral radius. However, it is a [theorem](https://en.wikipedia.org/wiki/Perron%E2%80%93Frobenius_theorem) (rather, a whole theory, what you can easily get lost in for months) that for any matrix which is strictly non-negative in its entries, there is always a real eigenvalue witnessing its spectral radius, call it $\hat{r}$. Furthermore, it says that *any* strictly positive eigenvector $\mathbf{p}$ must have this number $\hat{r}$ as its associated eigenvalue. Thus the number $r$ above which determines the EPR is *always* the maximal eigenvalue of $M^T$ (and also $M$).  

Why does this matter? Because $r$ is now the maximal eigenvalue of $M^T$, we have imposed a new condition on the matrix, which is that the spectral radius of the matrix must be less than $1$: $\rho(M^T) = \rho(M) < 1$. Equivalently, we have that **for a profitable capitalist economy, the matrix $M$ must be productive.** 

Now we are already normalizing our matrix $A$ to be productive. This is to ensure the existence of gross bundles, i.e. it creates an economy where it is always possible to produce more from less. But this alone is not sufficient to ensure *profitability* of the economy. For that, conditions must be imposed on the second half of the matrix $M$, the $\mathbf{l}\mathbf{b}^T$ part. Simply put, if the hourly real wage bundle is too big, then profitability will be impossible. 

This is the capitalist economy's version of our economy's condition limiting overconsumption: that $\mathbf{\lambda}\cdot \mathbf{c} < 1$. Indeed, the consumption frequency vector in our model is effectively our economy' version of the real wage. However, it stands to reason that our condition is a fair bit more forgiving than the profitability condition. I would wager, though I have not proven, that, it will often be the case that the $\mathbf{c}$ vectors generated do not always meet the profitability conditions for a capitalist economy. However, that does not mean we can't always obtain a realistic price vector from our consumption frequency vector. A simple algorithm should do nicely:

Given the requirements matrix $A$, the living labor vector $\mathbf{l}$, and a consumption frequency vector $\mathbf{c}$, do the following:
Let $\mathbf{b} := \mathbf{c}$, $s < 1$ be some arbitrary constant close to 1, say $0.97$.
While $\rho(A+\mathbf{l}\mathbf{b}^T) \geq 1$, do: $\mathbf{b} \mapsto s\mathbf{b}$

This loop must eventually terminate, since $\rho(A) < 1$ and the sequence of matrices here converges to $A$. After that, we will have a *pre-revolution* set of consumption frequencies, i.e. a real wage. The idea is that exploitative conditions limited people's consumption. Once a suitable $\mathbf{b}$ vector is found, there is guaranteed to be a strictly positive equilibrium price vector which can be seen as a reasonable proxy for real world pre-revolution prices, which can be easily found using the same eigen library we are already using.

If we set initial prices equal to these, rather than values, we will have a much more interesting scenario and a more rigorous test of our model's ability to assert stability for itself in the wake of a revolution. 

# Scenarios in General
I think that this should be implemented as an optional argument. Sometime like `--initial_prices`, with different strings giving different initial conditions. I think the two options for now should be 'values' and 'epr', with the default being the former if the user gives a bad argument.  I think that this is how scenarios should be implemented across the board from here on.

### LINKED ISSUE #222: Add optional arguments for product input epsilon and product consumption mult
I've already done this on my branch, I'm just making it an issue for my PR to resolve. I needed the latter to properly keep track of the minimal busyness levels, and I figured it would be better to simply make it a controllable parameter from Overseer rather than logging it.

### LINKED ISSUE #224: Always adjut I/O matrix according to product_input_epsilon, also widen range of input costs
Currently we are only adjusting the I/O matrix when the max eigenvalue is at least 1:

```
    if (max_eigenvalue >= 1.0) {
        adjust_io_matrix(A, max_eigenvalue);
    }
```

This means that when the matrix is already productive, the `PRODUCT_INPUT_EPSILON` constant has no influence on the simulation. Worse yet, this nearly always is the case because of our extremely conservative choices for `PRODUCT_INPUT_PER_UNIT_MIN` and `MAX`:

```
const double PRODUCT_INPUT_PER_UNIT_MIN = 0.1;
```

This window is unnecessarily narrow. We should make this window as possible and let our normalization handle the extreme, to maximize the variety of our simulated economies. Furthermore, we should make the call to `adjust_io_matrix` unconditional. This will give us more direct fine-grained control over economies via the `PRODUCT_INPUT_EPSILON` constant.

### LINKED ISSUE #230: Recompute prices based on up-to-date data
When a firm completes a plan to produce a product, it sends the completed plan to the Price Controller to update that product's price, based upon the firm's actual experience, especially its expenditure of living labor.

Currently, in `PriceController::update_price`, the Price Controller
- Iterates over all a limited historical window of plans to produce that same product.
- Sums over all plans
  - The total hours expended in living labor and raw materials
  - The total number of workers over all plans
  - The total number of units produced over the whole window.
- Adds to the total hours the expenditure on machinery, which is computed on the assumption that each machine is used all the time, but by one worker at a time.
- Divides the total hours by the total units to get the new price.

There are a couple of problems with this algorithm:
- The computation of raw materials costs takes, for each plan, the budget for raw materials minus _raw materials remaining_&mdash; but this latter field in `Plan` is apparently obsolete and should be eliminated. As we model production now, when a firm starts executing a plan, it moves the _whole_ budget item for raw materials from the budget to the pooled input value account and sets `raw_materials_remaining` to 0. Certain other computations dependeng on `raw_materials_remaining` are therefore nonsensical, such as its update in `Firm::move_plan_forward_one_step`.
- The computation is thus indirectly based on the costs of inputs at the time when the plan was first drafted. This dependency imposes a delay between any updates to the prices of the inputs and the current update of the product's price. That delay has been tolerable because, when a firm creates a new draft plan, it draws the costs of the inputs from their current prices. However, this means that the current price of the product is out of step with the current prices of the inputs to make that product.

Instead, the updated price of a product should be based upon the current average living labor and the current prices of the inputs and machinery. The inputs and machinery may have already had their prices updated at some other time prior to the current plan being completed, and the new price should reflect those possible updates.

The proposed new algorithm would work as follows:
- Get the total living labor hours and total units over the window.
- Divide living labor hours by units to get the current average living labor per unit.
- Add to this
   - For each machine needed to make a unit of the product, the cost per unit of that machine
   - For each input (raw material) needed, the cost per unit of that input product.
- Update the product's price per unit accordingly.

alexbcreiner0:
Resolves #222
Resolves #224 (Partially. I left the constants untouched because they seem to have a pretty extreme effect on the simulation stability. Requires group discussion.)
Resolves #221
Resolves #230 

Also replaced the `PRODUCT_INPUT_EPSILON` constant with a new one called `DEGREE_OF_PRODUCTIVITY`. A small calculation is performed to calculate the epsilon from that which is required to set the matrix productivity to the desired number. This means that the spectral radius of the number is exactly what you set the parameter to - higher numbers give more strict requirements with less room for error, and lower numbers are more forgiving. This also has the advantage of being parallel with how the `PRODUCT_CONSUMPTION_MULT` controls consumption, with lower numbers being more forgiving. 

- Fixed a bug with consumer goods in which the matrix was being normalized against them, i.e. if 1 unit of a good was required to make the consumer good, then this was getting reduced to something below 1, which doesn't make sense. So now we only normalize the submatrix which doesn't include those.

The big thing is the addition of an option for starting the simulation off with equilibrium prices. The calculations I do are pretty hairy because we don't guarantee irreducibility of the matrix $A$. Because of this there is no guarantee of a one dimensional eigenspace, requiring a lot more work to find a suitable price vector. I ended up using a very odd recursive algorithm that guesses different combinations of the eigenvectors trying to find something non-negative. 

It's fun watching the prices snap to values. You can mess with the settings yourself by pulling the latest version of the data aggregation layer. 
image


src/main.cpp
```diff
@@ -49,130 +46,117 @@ void set_params(int argc, const char ** argv, SimArgs& args) {
...
-            }
-            case argType::ProductsPerMachine: {
-                if (value <= 0) {
-                    error = true;
-                } else {
-                    args.products_per_machine = value;
-                }
-                break;
+        
+        if(valid_args.count(arg)) {
```
lennyyyyyyyy: The flag logic here is a bit outdated, so I'm just leaving this as a note to merge and resolve the conflicts that'll pop up here.

src/main.cpp
```diff
@@ -15,11 +15,10 @@ void print_usage() {
     std::cout << "\t-m N: Set the initial number of products per machine to N." << std::endl;
     std::cout << "\t-r N: Set the initial number of producers to N." << std::endl;
     std::cout << "\t-d N: Set the initial number of distributors to N." << std::endl;
-    std::cout << "\t-a N: Set the number of modeled abilities (work skills) to N." << std::endl;
-    std::cout << "\t-v N: Set the standard deviation of abilities to N." << std::endl;
     std::cout << "\t-e N: Set the random seed to N." << std::endl;
     std::cout << "\t-s N: Set the annual chance of an agent getting sick." << std::endl;
     std::cout << "\t-j: Write JSON log traces to stdout." << std::endl;
+    std::cout << "\t--init_prices S: Options are 'labor_values' and 'equilibrium_prices'." << std::endl;
```
lennyyyyyyyy: We should have an error check for the entered string being valid at least. Either that, or just make an option --epr, where labor values are assumed by default.

src/Society.cpp
```diff
@@ -191,7 +193,9 @@ void Society::log_total_employment() {
 void Society::adjust_io_matrix(
     Eigen::MatrixXd& io_matrix,
     double max_eigenvalue) {
-    io_matrix /= (max_eigenvalue + PRODUCT_INPUT_EPSILON);
+    double epsilon = Sim::get_productivity();
+    double scalar = ((1-epsilon)*max_eigenvalue) / epsilon;
+    io_matrix /= (max_eigenvalue + scalar);
```
lennyyyyyyyy: Think this is equivalent to `io_matrix *= epsilon / max_eigenvalue`. But also, I think a productivity of 0 should give a spectral radius of 1 (exactly reproducing), not 0 (gives infinite surplus). I also think using epsilon for this variable is a little confusing.

`io_matrix *= (1 - Sim::get_productivity()) / max_eigenvalue`

header/Sim.h
```diff
@@ -16,8 +16,10 @@ struct SimArgs {
     unsigned int num_producers = STARTING_NUM_PRODUCERS;
     unsigned int num_distributors = STARTING_NUM_DISTRIBUTORS;
     double sickness_chance = ANNUAL_SICKNESS_CHANCE;
-    unsigned int num_abilities = NUM_ABILITIES;
-    double ability_stddev = ABILITY_STDDEV;
+    double ability_stddev = PERSON_ABILITY_STDDEV;
+    double degree_of_productivity = DEGREE_OF_PRODUCTIVITY;
+    double consumption_epsilon = PRODUCT_CONSUMPTION_MULT;
```
lennyyyyyyyy: I know I made the name for PRODUCT_CONSUMPTION_MULT, but I'm thinking CONSUMPTION_PRODUCTION_RATIO is more descriptive.

src/Society.cpp
```diff
@@ -258,6 +261,237 @@ void Society::set_product_prices_production_consumption() {
...
+
+            while (spectral_radius >= 1.0) {
+                b *= CONSUMP_REDUCTION_FACTOR;
+
+                augmented_matrix = A+l*b.transpose();
+                spectral_radius = get_max_eigenvalue(augmented_matrix);
+                
+                ++reductions;
+
+                if (reductions > MAX_CONSUMPTION_REDUCTIONS) {
```
lennyyyyyyyy, line 458: I think if we correctly check that the productivity is greater than 0 and less than 1, it should always be possible to find a b. Binary search would also make this better, since the spectral radius is monotonic while scaling b.

src/main.cpp
lennyyyyyyyy: Also, it looks like we're scaling b down until M has a spectral radius barely under 1. However, since the EPR prices are a function of the capitalist return on investment, I think it makes sense to define an exploitation parameter
$e = \frac{1 - r}{p}$ that ranges from 0 to 1. p is the productivity that ranges from 0 to 1, that sets the largest eigenvalue of A to 1 - p.

Then we can do like `--epr 0.5`

src/Society.cpp
```diff
@@ -258,6 +261,237 @@ void Society::set_product_prices_production_consumption() {
     }
 }
 
+static bool try_active_constraint_combination(
```
lennyyyyyyyy: After some research, this iterative algorithm may not be necessary. If our augmented requirements matrix is all non-negative, even if it is reducible, the leading eigenvector will also have non-negative entries. Setting initial prices to 0 shouldn't cause any problems either.

src/main.cpp
alexbcreiner0: This is a good idea, I'll do that.

src/Society.cpp
alexbcreiner0: I agree the naming is off, but I think what's warranted here might be a different variable name rather than a reversal of the meaning. Setting the spectral radius seemed the most direct, and is parallel with how product_consumption_mult already behaves. So I'll probably just change the names.

header/Sim.h
alexbcreiner0: I agree, I'll change both names to something better.

src/Society.cpp
alexbcreiner0: I agree, I wasn't sure how necessary this was, it's probably an unnecessary guardrail but I wanted to be 100% certain that something this would never make the sim hang.

src/Society.cpp
alexbcreiner0: The leading eigenvector CAN have non-negative entries, but doesn't have to, because the real problem is that there isn't always just one leading eigenvector. That's what you get in the case of a reducible matrix: the possibility of a multi-dimensional eigenspace. If that's the case, the PF theorem only guarantees that a non-negative vector exists in this span. But there's no guarantee that any of the linearly independent eigenvectors returned by the Eigen library will be strictly non-negative. It might require a linear combination of them to actually get that. So zero prices aren't the issue, negative prices are. (In fact, I don't think zero prices can occur at all actually as long as the living labor vector is strictly positive.)

src/Sim.cpp
```diff
@@ -57,6 +57,18 @@ double Sim::get_annual_sickness_chance() {
     return get_instance().args.sickness_chance;
 }
 
+double Sim::get_productivity() {
```
lennyyyyyyyy: Can you change this to `get_difficulty_of_production`?

src/Society.cpp
```diff
@@ -235,8 +237,41 @@ void Society::log_total_employment() {
...
+            productive_matrix(i, j) =
+                io_matrix(kept_indices[i], kept_indices[j]);
+        }
+    }
+
+    double productive_max_eigenvalue = get_max_eigenvalue(productive_matrix);
+
+    double epsilon = Sim::get_productivity();
+    double scalar = ((1 - epsilon) * productive_max_eigenvalue) / epsilon;
+    double divisor = productive_max_eigenvalue + scalar;
```
lennyyyyyyyy: Again, I think this math is equivalent to setting divisor to `productive_max_eigenvalue / Sim::get_productivity()`

src/Society.cpp
```diff
@@ -302,6 +336,237 @@ void Society::set_product_prices_production_consumption() {
...
+            }
+
+            log_vector(b, "b", b.size());
+            Eigen::VectorXd prices = get_epr_prices(
+                augmented_matrix,
+                spectral_radius
+            );
+
+            for (std::size_t i = 0; i < dim; ++i) {
+                if (prices(i) <= 0.0) {
```
lennyyyyyyyy: If you look in get_epr_prices, you set prices equal to 0.0 if they're just below the tolerance. You should probably just change this condition to < 0.0. 

Even better, get_epr_prices should throw an error by itself, and should never return prices with negative values anyway.

src/Society.cpp
```diff
@@ -302,6 +336,237 @@ void Society::set_product_prices_production_consumption() {
...
+                "initial_price_fallback",
+                LogPairS("requested_mode", price_mode),
+                LogPairS("reason", e.what())
+            );
+            use_labor_values = true;
+        }
+    }
+    else {
+        use_labor_values = true;
+        if (price_mode != "labor_values") {
```
lennyyyyyyyy, line 515: This should be checked in main.cpp, when the program first reads the flags entered on the command line.

src/Society.cpp
```diff
@@ -302,6 +336,237 @@ void Society::set_product_prices_production_consumption() {
...
+            }
+        }
+
+        double norm = p.norm();
+
+        if (norm <= tol) {
+            return false;
+        }
+
+        p /= norm;
```
lennyyyyyyyy: The normalization is arbitrary, since prices are an eigenvector. This makes sense for capitalism. We should figure out a way to justify a normalization when we set initial prices in labor hours though.

src/Society.cpp
```diff
@@ -302,6 +336,237 @@ void Society::set_product_prices_production_consumption() {
...
+            throw std::runtime_error("Perron root is not non-negative!");
+        }
+
+        for (Eigen::Index i = 0; i < n; ++i) {
+            if (p(i) < 0.0) {
+                p(i) = 0.0;
+            }
+        }
+
+        p /= p.norm();
```
lennyyyyyyyy, line 380: Same thing about normalization. Might need more thinking.

lennyyyyyyyy (changes requested):

src/Society.cpp
alexbcreiner0: This normalization is arbitrary, which is precisely why it is standard practice within all of these models to normalize p to being a unit vector. Nobody would question this.

src/Sim.cpp
alexbcreiner0: done.

src/Society.cpp
alexbcreiner0: done.

src/Society.cpp
alexbcreiner0: Note sure what you mean by this. This function still needs to decide how to set prices, and behave differently based on the price mode. The idea is if the price mode is equilibrium prices, it tries to compute those, but falls back to labor values if something goes wrong or if the user gives a garbage input. What are you suggesting I change here?

src/main.cpp
alexbcreiner0: Done I think.

src/main.cpp
alexbcreiner0: I want to leave the door open for other initial pricing options besides values and EPR prices, so I want to stick with a string for init_prices. I'll do an error check though.

alexbcreiner0: Done.

src/Society.cpp
lennyyyyyyyy: It falls back if the flag --init_prices is given anything other than "labor_values" or "equilibrium_prices". I'm just saying that if the user passes anything else, it shouldn't even get here. It should throw an error while reading the flags in main.cpp.

In main.cpp::90, you can just put something like 
```
if (args.init_price_mode != "labor_values" && args.init_price_mode != "equilibrium_prices") {
   error = true;
   break;
}
```

header/Constants.h
```diff
@@ -34,14 +34,13 @@ const int PRODUCT_ABILITY_COUNT_MAX = 1;
...
-const double PRODUCT_INPUT_EPSILON = 0.25;
+const double PRODUCT_INPUT_PER_UNIT_MAX = 1.0;
+const double DEFAULT_PRODUCTION_DIFFICULTY = 0.25;
 const int PRODUCT_NUM_MACHINES_MIN = 0;
 const int MAX_PROPORTION_OF_MACHINES_PER_PRODUCT = 5;
-const double PRODUCT_LABOR_PER_UNIT_MIN = 0.05;
-const double PRODUCT_LABOR_PER_UNIT_MAX = 0.1; 
-const double DISTRIBUTION_LABOR_PER_UNIT = 0.05;
-const double PRODUCT_CONSUMPTION_MULT = 0.6;
+const double PRODUCT_LABOR_PER_UNIT_MIN = 0.05; const double PRODUCT_LABOR_PER_UNIT_MAX = 0.1;
```
lennyyyyyyyy, line 41: Little style thing here. Also it'd be cool if this was just a MIN MAX ratio constant. Doesn't have to be included in this PR though.

src/Society.cpp
lennyyyyyyyy: Nevermind. You already fixed it lol

src/Society.cpp
```diff
@@ -302,6 +339,186 @@ void Society::set_product_prices_production_consumption() {
...
+            }
+
+            try {
+                price_candidate = get_epr_prices(
+                    augmented_matrix,
+                    spectral_radius
+                );
+            }
+            catch (const MultidimensionalPerronEigenspace& e) {
+                price_candidate = retry_with_perturbation(A, l, b);
```
lennyyyyyyyy: Weird that you only try once and then give up. But at least its x^2 as unlikely now.

src/Society.cpp
```diff
@@ -302,6 +339,186 @@ void Society::set_product_prices_production_consumption() {
...
+                b(consumer_good->id) = consumer_good->mean_consumption_frequency;
+            }
+
+            Eigen::MatrixXd augmented_matrix = A+l*b.transpose();
+            double spectral_radius = get_max_eigenvalue(augmented_matrix);
+
+            int reductions = 0;
+            constexpr int MAX_CONSUMPTION_REDUCTIONS = 10000;
+
+            while (spectral_radius >= DESIRED_INITIAL_PROFITABILITY) {
```
lennyyyyyyyy: The way you've set up DESIRED_INITIAL_PROFITABILITY, it's required to be greater than DEFAULT_PRODUCTION_DIFFICULTY. Which is fine, but when I first suggested this, it was more like an interpolation factor:
profitability = 0 -> spectral_radius = 1
profitability = 1 -> spectral_radius = production_difficulty

This gives you an easy to understand range between 0-1 for profitability.

The naming here is opposite again too. When this constant is high, it's less profitable, and when it's low, it's more profitable. It'd be nice if you could either use my suggestion or name it like "unprofitability" (but this one doesn't sound that great).

src/Society.cpp
```diff
@@ -235,8 +244,37 @@ void Society::log_total_employment() {
 void Society::adjust_io_matrix(
     Eigen::MatrixXd& io_matrix,
     double max_eigenvalue) {
-    io_matrix /= (max_eigenvalue + PRODUCT_INPUT_EPSILON);
+    (void) max_eigenvalue;
```
lennyyyyyyyy: If you don't use the argument, you should probably get rid of it instead of void casting to get rid of the warning.

lennyyyyyyyy (approved):
I'm approving because 99% of it is here, and I don't think it's necessary to do another review, but please don't merge without looking at my final comments.

src/Society.cpp
alexbcreiner0: Done.

src/Society.cpp
alexbcreiner0: Okay, I like this idea. Finishing that up now. I am making the modification though that the profitability of zero corresponds to the minimum of 1 and the spectral radius of whatever M started as, so that there is never a situation where b needs to be scaled up.

src/Society.cpp
alexbcreiner0: Yeah I was too tired to be programming yesterday. Fixing this.

lennyyyyyyyy (approved):
LGTM

src/PriceController.cpp
```diff
@@ -43,6 +43,10 @@ void PriceController::update_price(Plan * plan) {
     double price = product->living_labor_per_unit = hours / units;
     double machine_use_hours = hours / workers;
     double machine_hours_per_unit = machine_use_hours / units;
+    for (std::pair<Good * const, double>& input_pair : product->inputs_per_unit) {
```
Amittai-Aviram, line 46: Could you please change this name? It's not the input cost (labor time) per unit, it's the input _quantity_ per unit. It's how much (in natural units of the respective input product) of the input is needed to make one unit of the output. That's why you would multiply it by the price per unit of that input.

Amittai-Aviram (changes requested):
I just reviewed the Price Controller changes, which look good. I am requesting one name change for clarity.

src/PriceController.cpp
alexbcreiner0: Done.

Amittai-Aviram (approved):
LGTM

PR #226, opened 2026-06-01
Distributor plan execution, demand signal exponential averages

lennyyyyyyyy:
Resolves #186 
Additionally:
 - demand signals are averaged exponentially, similar to busyness, which gets rid of the need for a queue, and the two constants `DEMAND_SIGNAL_WINDOW_MIN` and `DEMAND_SIGNAL_WINDOW_MAX`.
 - Simplifies the code for calculating a demand signal and also smoothens the signal - using a queue causes discontinuities when large demand signals exit the back end.
 - Allows us to set an initial demand, although currently predicted demand is still set to 0, with the Producers and Distributors using their surplus inventories to "learn" what the demand is before hitting the reorder threshold

image
As before, Distributors hit their threshold first (timestep ~800), and start reordering from Producers, whose supplies drop. They also start executing distribution plans, which can be seen by the emerging dotted peaks, which represent the incoming shipment, followed by the consumption and production of the respective consumer good. Producers then hit their threshold after a similar amount of time (timestep ~1800). Because demand signals are much more continuous than before, it's a lot easier to see when they hit the threshold.
image
It's a little confusing what the respective goods and consumer goods are here, but they're spaced by a distribution labor of 0.05 hours per unit, for example, the light purple and yellow, or the orange and dark blue. These prices are from plan reporting only.


alexbcreiner0 (approved):
It all looks good to me.

PR #228, opened 2026-06-03
Implement consumption by wealth

### LINKED ISSUE #227: Reduce wealth divergence through consumption
Wealth divergence results from the inability to always maintain an ideal division of labor; therefore, some firms will be busier or less busy than average, and their workers will respectively make more or less than the amount they consume. Implementing a public sector that redistributes wealth to the workers that happened to end up at less busy Firms obscures the connection between actual hours worked and labor hour credits gained by a worker. 

To fix this, we should just make wealthier members consume more and less wealthy citizens consume less to offset their diverging income.

lennyyyyyyyy:
Resolves #227

### DIFF
--- a/src/Person.cpp
+++ b/src/Person.cpp
@@ -60,6 +60,10 @@ double Person::get_busyness() {
     return busyness;
 }
 
+double Person::get_account() {
+    return account;
+}
+
 void Person::train(std::unordered_map<Ability *, double>& target_abilities) {
     // can introduce < 100% effectiveness on training later
     for (const std::pair<Ability * const, double>& ability : target_abilities) {
@@ -112,7 +116,12 @@ void Person::purchase_good(ConsumerGood * consumer_good, int quantity) {
 
 void Person::consume() {
     for (ConsumerGood * consumer_good : society->get_consumer_goods()) {
-        to_consume[consumer_good] += consumer_good->mean_consumption_frequency;
+        double consumption_from_wealth = 
+            (account / society->get_average_account() - 1)
+            * CONSUMPTION_FROM_WEALTH_MULT;
+        double consumption = consumer_good->mean_consumption_frequency 
+            * (1.0 + consumption_from_wealth);
+        to_consume[consumer_good] += consumption;
         int consumed = static_cast<int>(to_consume[consumer_good]);
         if (consumed) {
             inventory[consumer_good] -= consumed;
--- a/header/Constants.h
+++ b/header/Constants.h
@@ -12,9 +12,6 @@ const unsigned int YEAR = DAY * 365;
 const unsigned int INITIAL_WORK_HOURS_DAILY = 8;
 const unsigned int INITIAL_WORK_DAYS_WEEKLY = 5;
 const double INITIAL_ACCOUNT_DURATION = YEAR;
-const int WORK_HOURS_UPDATE_START = YEAR;
-const int WORK_HOURS_UPDATE_PERIOD = MONTH;
-const double INEFFICIENCY_OF_WORK = 1.5;
 
 const unsigned int STARTING_NUM_PEOPLE = 1000;
 const unsigned int STARTING_NUM_GOODS = 100;
@@ -57,6 +54,7 @@ const double PERSON_STOCKPILE_DURATION = WEEK;
 const double PERSON_DEFICIT_THRESHOLD = WEEK / 2.0;
 const unsigned int NUM_ABILITIES = 5;
 const double ABILITY_STDDEV = 0.05;
+const double CONSUMPTION_FROM_WEALTH_MULT = 0.5;
 
 const double UNHEALTHY_PRODUCTIVITY = 0.5;
 const double ANNUAL_SICKNESS_CHANCE = 0.1;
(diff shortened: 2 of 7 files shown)


src/Person.cpp
```diff
@@ -112,7 +116,12 @@ void Person::purchase_good(ConsumerGood * consumer_good, int quantity) {
 
 void Person::consume() {
     for (ConsumerGood * consumer_good : society->get_consumer_goods()) {
-        to_consume[consumer_good] += consumer_good->mean_consumption_frequency;
+        double consumption_from_wealth = 
+            (account / society->get_average_account() - 1)
+            * CONSUMPTION_FROM_WEALTH_MULT;
+        double consumption = consumer_good->mean_consumption_frequency 
+            * (1.0 + consumption_from_wealth);
+        to_consume[consumer_good] += consumption;
```
alexbcreiner0, line 124: I don't think I like this. There is no need to violate the principle of agents operating off of local knowledge just for something like this. Instead, they should be monitoring the growth rate of their own personal accounts. People should be seen as individual regulators in charge of stabilizing their accounts independently of each other. If they see their account is steadily falling, they should slow down their consumption. If they see their account steadily rising, they should increase their consumption. I recommend you implement monitoring of the relative rate of change of their accounts over time and have their consumption decisions be based on that.

alexbcreiner0 (changes requested):

src/Person.cpp
lennyyyyyyyy: Whether it's based off the rate of change or the absolute amount of labor credits in their account, operating only off of local knowledge creates feedback loops (I thought we talked about this?), if the average busyness or wealth across society isn't stable.

If we based it off local rate of change, any small fluctuation upwards in average societal busyness will cause those people to consume more, while the rest consume the same. Demand increases, people get busier, and then everyone consumes more. Likewise for downwards.

If we did it off local wealth only, then the same thing happens when the price controller is slightly off. Money flows into or out of the system, and if everyone sees an increase or decrease in their local wealth, they'll reinforce that with their consumption.

Admittedly, this is the first time that the average wealth has been publicly available, but societal average busyness has already been available to Firms, so maybe I can just use that.

alexbcreiner0: So your point about a runaway positive feedback loop in my suggestion is correct, but not because of over-reliance on local knowledge. The issue is that I didn't specify a signal to stabilize. What we need is a local _target_ level for their account to stay at.  How about this instead: people have a target lower level of account, which is determined by the credits they would need in order to consume at ideal levels for a certain amount of time (maybe something like a month). If they begin to fall below this desired level, then they slow down their consumption. However, the correction only happens when accounts fall below the target. Maybe we hold off on having excess account lead to higher consumption for now, or implement a cap to how much it can increase.  

The bottom line is that if I can regulate my local purchases in real life to maintain a consistent level of available money in without knowing how much money the average person around me has, then the agents in our model can do it too. 

A more general note about something you said before that:

> Whether it's based off the rate of change or the absolute amount of labor credits in their account, operating only off of local knowledge creates feedback loops (I thought we talked about this?), if the average busyness or wealth across society isn't stable.

This is not the right lesson to takeaway from that issue. In fact, what we encountered there with the price controller was specifically due to _not_ operating sufficiently off of local knowledge. There, we were over-relying on global signals, using the global average labor time to determine both estimated completion times as well as local outcomes. It was this double dipping which created a runaway _positive_ feedback loop, and it was precisely the _introduction_ of local estimates which resolved the issue. 

Feedback loops can be good or bad. Positive feedback loops are self-reinforcing behaviors that spin out of control. They tend to be very dangerous and unintended, and they happen when the deviation of a signal triggers further deviations in that signal in the future _in the same direction._ Negative feedback loops, on the other hand, are _incredibly useful_ for regulating system. The only difference is that future signal deviations are always in the opposite direction. This makes them ideal for stabilizing a signal at a desired value. This is what we need here.

What you're doing here is using a global measure of average wealth as the target for stabilizing account levels using a negative feedback loop. I think we just need to replace that target with something locally determined. There is no danger in this.

Amittai-Aviram: The question at issue here is how to model an individual person's behavior, given, as an input, how much in labor credits that person has available.  This question lies in the realm of guessing what seems like reasonable behavior.  If the divergence in wealth (acount holdings) had a feedback effect on the society as a whole, then we would have to move the question from that realm to the realm of system design &mdash; it would belong in Systems II and III. As it is, as far as I can tell, it has o such effect, so it belongs to System I, and we're trying to model a reasonable person's behavior.  In answering that question, we then have to ask what the system provides (through System II) to enable person's to make reasonable decisions. 

We could look to two such possible inputs provided by the society as a whole:
- The average individual wealth (or income), which is what you have chosen here.
- The current price of a reasonable basket of consumption.
_Both_ of these items are available to us in the real world, so it's reasonable to expect that our modeled society will also provide them.

In the real world, what enables you to feel that you can afford to splurge, or what compels you to tighten up on expenses?  On the first criterion — I find that I have more in my bank account than I would even need for a somewhat bad emergency (such as a car breakdown or being evicted or losing my job) &mdash; so I can afford to go out to a nice restaurant once a week or so. Also, I find that, even with staple prices as high as they are, my partner and I can manage our budget and still have something left over, so at least some of that can be spent on nice-to-haves (such as a trip to visit my son in New York City). My point is that either one of those criteria, or both together, provide a common basis for a reasonable decision about spending. Neither input datum about the society as a whole is extraordinary, in my (very) humble (current) opinion.

PR #229, opened 2026-06-04
Implement machine life cycle

### LINKED ISSUE #203: Implement the replacement cycle of machines

Amittai-Aviram:
- When a firm adds a product to its catalog, it acquires the necessary machines, but does not record information about its life cycle. (This avoids a circular dependency: in order to estimate when to re-order a machine nearing the end of its useful life, the firm needs an estimate of its required turnaround time, which requires that a supplier have that machine as a product, which, at the beginning of time, requires that `Society` finish assigning products to producers.)
- When a firm starts a new plan
  - It "initializes" every machine needed for the plan, i.e., it starts tracking the machine's remaining useful life and also gets an estimate of when to re-order the machine.
  - It moves all the value budgeted in the plan for machinery to the general input value pool.
- On every time step in executing the plan
  - We check whether it is time to re-order the machine
  - We re-order the machine if it is indeed time to do so. (The value to pay for the new machine comes out of the
  - Meanwhile, if a machine has actually reached the end of its lifetime, we just stall until the new machine is available.
  - Every machine involved in the plan has its remaining lifetime reduced by one hour
input value pool, as it does in re-ordering input goods.)
- Re-ordering a machine is like re-ordering other inputs, but we order only one machine at a time.
- When the new machine comes in (`Firm::receive_shipment`) , we add its full lifetime to the lifetime remaining for that type of machine.
- When a producer receives an order, it checks the remaining lifetimes of all of its machines that are needed for the order, and. if it will need to replace a machine during execution of the draft plan, it adds that machine's replacement lead time to the draft plan's total turnaround time.

These changes also include
- Some cases of avoiding memory leaks
- Use of `double` instead of `int` for the machine's lifetime (which is an "internal" accounting value)
- A couple of small stylistic corrections.

Resolves #203


alexbcreiner0:
> * We re-order the machine if it is indeed time to do so. (The value to pay for the new machine comes out of the

Out of the what?

header/Firm.h
```diff
@@ -9,10 +9,11 @@
 
 #include "Agent.h"
 #include "Logger.h"
+#include "Machine.h"
 #include "Person.h"
 
 struct Ability;
-struct Machine;
+// struct Machine;
```
alexbcreiner0: Aghast at this blatant hypocrisy!

src/Firm.cpp
```diff
@@ -51,13 +51,36 @@ void Firm::on_time_step() {
...
 
 void Firm::receive_shipment(Plan * plan) {
     Order * order = plan->order;
-    input_inventory[order->product] += order->quantity;
+    Product * product = order->product;
+    if (product->product_type == Product::ProductType::TYPE_GOOD) {
+        input_inventory[product] += order->quantity;
+    } else if (product->product_type == Product::ProductType::TYPE_MACHINE) {
+        Machine * machine = static_cast<Machine *>(product);
+        machine_lifetime_remaining[machine] += machine->lifetime;
```
alexbcreiner0: Am I right in the implication here that firms will only have a single machine per machine type? Is this a good assumption to make?

src/Firm.cpp
```diff
@@ -174,19 +196,60 @@ void Firm::check_and_reorder_input(Product * product) {
...
-    if (!send_order(order)) {
-        log_reorder_failure(product, order->quantity);
+    if (!send_order(&order)) {
+        log_reorder_failure(product, order.quantity);
+    }
+}
+
+void Firm::check_and_reorder_machine(Machine * machine) {
+    if (
+            machine_lifetime_remaining[machine] > machine_reorder_time[machine] ||
```
alexbcreiner0: This assumes rather optimistically that machines will be used every hour regardless of plans. Maybe the LHS here should be modified to take the actual hourly usage of the machines into account? 

For example, if the machine is only being used one hour of every two, then we would want to double the lifetime remaining as our timetable for when to have a replacement by.

alexbcreiner0 (changes requested):
If I've understood the code correctly, we're currently making the following assumptions about our machines:
1. Firms will only elect to possess a single instance of any needed machine type at a time.
2. Only one machine will ever be needed at a time for a plan, regardless of the number of workers assigned to it.
3. Regardless of the number of workers assigned to use a machine, the machine will only ever be deducted a single hour of usage. 

I don't really hate these assumptions as a first step, but I'm not satisfied with them either: the system still seems extremely simplistic. So I don't really mind approving this as a first step but I will still offer some things I would have liked to see here, and you can choose to add some or none of them. I don't believe any of the things I'm about to list are new. They are all based on my (maybe foggy) recollection of discussions we've had over the months: 

- I think that machines should have properties designating a min and max number of workers who can use it at a time (perhaps with an option for no maximum at all, which would be appropriate for something like an overall workspace). 
- Firms do their best with 1 machine (or some reasonable number which could be determined as part of the initialization process, or some unreasonable random number) at first, but then order more if the turnover times this allows for are inadequate. If they have too many machines, this just means there will be less machine demand until those excess machines run out.
- Firms can possess more than one of a machine, and they decide how many copies of each to possess dynamically based on their demand, number of employees, and desired turnover times.
- If, say, 5 workers are assigned to a plan using 3 machines and each machine facilitates a maximum of 2 people, then each hour of the plan's execution, two of the machines will have 2 hours reduced from their lifespan, while the third machine has one hour reduced from its lifespan. 

This all makes the decision for when to order replacement machines much more interesting - it requires that firms keep track of how much use their machines are actually getting hour by hour.

src/Firm.cpp
alexbcreiner0, line 384: Shouldn't this happen before the budget is initialized? It also looks like in the initialize plan budget function, we aren't adding the machine cost anywhere.

alexbcreiner0 (changes requested):

header/Firm.h
Amittai-Aviram: Oops! Sorry! (Removed — as you might guess, I was testing whether I could do without it, and then forgot to delete it.)

src/Firm.cpp
Amittai-Aviram: Yes, that is the current assumption, along with the assumptions —
- That each machine is used at all times that the workers are working
- That only one worker works with a machine at a time.
Both of these are simplifying assumptions that it will take some work to relax, so we have deferred that challenge for later.

src/Firm.cpp
Amittai-Aviram: It a machine is used only every other hour, then reordering it when the remaining lifetime is less than or equal to the replacement lead time would err on the safe side. You would want to reorder earlier if you're using the machine faster than one hour per hour of team work.

src/Firm.cpp
Amittai-Aviram: Moved to `Firm::initialize_plan`. (I'm afraid this was a Sal thing that I did not catch on review or re-examine. I may  not have reviewed it in the first place.)

src/Firm.cpp
alexbcreiner0: That only addresses my admittedly backward use of the word optimistic. What I meant by that word is exactly what you're saying - it's excessively safe. I think that there should really be some monitoring of _actual_ hourly usage of the machines in this determination. That can be in a future PR though.

src/Firm.cpp
alexbcreiner0, line 360: This should have machinery cost added to it. Likewise, in move_plan_forward_one_step, machinery cost should be properly decremented from it.

alexbcreiner0, line 226: This is missing the appropriate updates and deductions for machinery_cost accounting.

alexbcreiner0 (changes requested):
See my comments in `move_plans_forward_one_step` and `initialize_plan_budget` - there is still some accounting that is missing.

src/Firm.cpp
Amittai-Aviram: Your first point is correct, and I have changed the code accordingly. However, the machinery line in the plan's initial budget is moved _altogether_, in `Firm::start_plan`, from the budget to the pooled input account. This is also the case for all the input products. In neither case should the budget line be decremented on every time step of the plan. This decrementing, and the associated fields and variables, are relics from an earlier conception of the process, one that Sal never brought up to date when he implemented the pooled account. I have written a separate issue (Issue #231) to correct this and remove the obsolete code, but it is beyond the scope of the current PR.

Amittai-Aviram: Please see above. The only budget line that should be decremented on every step of the plan is living labor, because we need it at the end to find out how much living labor we have actually spent. We do not do that for either inputs or machinery, so the code that looks as if it does that needs to be deleted (in another PR). We may eventually wish to model variations in the use of inputs nad perhaps wear and tear on machinery, but each of those would involve a bigger change than might first appear. To start with, we would have to decide how we wish to model such variations, as we have now modeled, to some extent, variations in work speeds among workers. Accordingly, those design choices belong to different tasks. Right now, we just need to account for machines at all, which we have not done hitherto.

alexbcreiner0:
I have no further comments, but I am now testing the simulation and am encountering many segfaults. In general, the simulation never makes it to 1000 hours in Overseer. Example run `./bin/sim -n 1000 -p 100 -r 5 -d 5 -g 5`

Please find out what is causing the crashes and fix it and then I will approve the merge.

Amittai-Aviram:
I believe the memory management issues have now been corrected. I had been testing on only 500 time steps, which was not enough to reveal them. There were a few different issues having to do with the lifetimes of orders and plans. I can now reliably run on -n 2000 without a crash.  Command: 
`bin/sim -e 1 -n 2000 -g 100 -p 30 -r 5 -d 5`

alexbcreiner0:
It no longer segfaults but it is causing serious problems for the overall stability of the simulation. Under these changes the simulation has lost all ability to assert a reasonable production flow, even under ideal conditions (low difficulty of production, low variance in abilities, low sickness rate). I think you need to take a look at the effects that this is causing and investigate before we push it to main.

image

I would offer more help but I need to pack today and tomorrow. I can't dedicate any more time to this right now, probably for mot of the week. Maybe Devin could take a look?

PR #235, opened 2026-06-08
FIC and Distributor demand flow

### LINKED ISSUE #218: Implement a public sector (Detailed outline provided)
Here is a proposed outline on how we can go about implementing a 'public sector'. I believe my plan can easily be working by the end of the summer if everyone is on board with doing things the way I'm proposing here. 

- First, in terms of scope, let us limit the public sector to the realm of 'free goods'. This is really more of an abstraction than a scope limiter - we are simply choosing to see all public services as goods like any other which are produced. These could be unique goods produced entirely within the public sector, or just free instances of goods which are also produced under normal circumstances.
- Second, let us make our public sector 'ethereal' - that is to say, no firms will be deemed explicitly as 'public sector' or not. This is not to say that some firms aren't public sector firms, I am simply proposing that we refrain from worrying about whether this or that firm is a 'public sector firm', or that this or that worker is a 'public servant'.  

Under these two assumptions, *the task of implementing a public sector becomes equivalent to the task of designating a portion of the net product as free goods*, which can be taken from distributors without any expenditure of labor credits. The key to doing this is the FIC, or *factor of individual consumption*, as the GIC describes it. 

To frame this properly, let $\mathbf{n}$ denote the weekly net product. Our implementation of a public sector would be to take this net product and split it into two components: a free component $\mathbf{n}_f$, and a paid component $\mathbf{n}_p$.:
 
$$ \mathbf{n} = \mathbf{n}_f + \mathbf{n}_p $$

This could be done in a variety of ways, but perhaps the simplest way, and the way I am proposing, is that we take a specific commodity type (perhaps the first or last commodity of the ones initialized), and designate it to be free as a rule. Let producers do their normal thing, and simply instruct distributors to give the goods out for free without the need to exchange labor credits. Keep consumption the same as before. This is my preferred route, because it requires no anticipation of what the net bundle will actually be. It is worth noting though that this framing of the public sector as a partition of the net demand bundle is very general. 

In a [sister discussion post to this one](https://github.com/BC-LTEWG/Labor-Time-Economy-Simulation/discussions/219), I derive an equation for the FIC, which I'll denote $\alpha$, within our model:

$$ \alpha = \frac{\mathbf{\lambda}\cdot \mathbf{n}_p}{NK} $$ 

Here, $\mathbf{\lambda}$ is the vector of unit values, i.e. the prices currently reported by the price controller, and $\mathbf{n}_p$ is the net weekly bundle of goods which are *not* free, so that $\mathbf{\lambda} \cdot \mathbf{n}_p$ is just the total price of all goods which workers are exchanging labor credits for. Finally, $NK$ is simply the total number of hours which workers spend working on plans each week. All of these numbers are easily accessible within our model and already being kept track of. This equation looks quite a bit different from that of the GIC, but in the sister post I prove that the two formulas are equivalent.

 What we need to do with this number is simply multiply it by any labor credits which are given to workers as they work on a plan. Since we are only giving hours out one at a time, we would simply give $\alpha$ instead. 

So my proposal for implementing a public sector is very simple. Let the system stabilize (or don't). Pick a good to designate free, and have distributors stop charging labor credits for it. At the same time, modify the labor credits paid to workers according to the FIC as I've defined it here. We could do some interesting things, like have a new commodity be made free each year until the FIC is 0. In this way, we could easily witness the transition to 'full communism'!

### LINKED ISSUE #234: Reorganize Distributor demand flow
Currently, the demand created by people buying consumer goods is translated directly into demand to reorder the producer good. Distribution plans are carried out whenever there are any producer goods left. Instead, purchases should translate to distribution plan execution, which should translate into demand for the producer good as it's consumed as an input.

lennyyyyyyyy:
Resolve #218  - Every `PUBLIC_SECTOR_EXPANSION_PERIOD`, a new consumer good is funded by the public sector, and the FIC is adjusted accordingly.

Resolve #234


header/PriceController.h
```diff
@@ -3,15 +3,24 @@
...
 
-class PriceController {
+class PriceController : public Agent {
     public:
         static PriceController * get_instance();
+        void on_time_step() override;
+        unsigned int get_id() override; 
         void update_price(Plan *);
+        void report_distribution(ConsumerGood *, int quantity);
+        double get_fic();
```
alexbcreiner0, line 18: I don't know about making these functions the job of the price controller. Why not give them to Society? I think the PriceController should be a very small, simple thing which does a very specific job. I'd rather have a similar FIC controller than this, but my first choice would be giving these over to Society.

src/Society.cpp
```diff
@@ -119,12 +120,28 @@ unsigned int Society::get_id() {
...
+    busyness /= people.size();
+    average_account /= people.size();
+
     for (Person * person : people) {
         person->on_time_step();
     }
     for (Firm * firm : firms) {
         firm->on_time_step();
     }
+    PriceController::get_instance()->on_time_step();
```
alexbcreiner0, line 138: Again I think it should really be Society doing these things for now. If we're worried about the file getting too big, we should maybe move some of the initialization details to a separate file imo. I'm hesitant to allow the PriceController to be given full agent status within our system. Curious what your thoughts are on this @Amittai-Aviram

src/Society.cpp
```diff
@@ -119,12 +120,28 @@ unsigned int Society::get_id() {
...
+
+    // public sector expansion
+    if (Sim::get_current_time_step() % PUBLIC_SECTOR_EXPANSION_PERIOD == 0
```
alexbcreiner0: Does this mean we have a free good at time = 0? I think we should include in this PR some optional arguments to allow for toying around with the FIC in ways such as this.

alexbcreiner0 (changes requested):
Can we include some logging of which goods are free and which are paid? Additionally, the FIC should be logged each time it is updated, or something like every week. It would also be nice if society logged the numerator and denominator of the FIC calculation separately, allowing us to keep track of the ratio of free to paid goods in value terms ourselves.

In the meantime, our main indicators that the FIC is working properly would be a drop in the average account each time that a new good is made free, with the average inventory of each person remaining unchanged and production flow now showing any visible changes either. Unfortunately I can't really get to that point right now. There seem to be some stability issues, presumably with the other side of this PR regarding consumer goods because I don't think the FIC side is doing anything until `6*MONTH` elapses (assuming that no good is free initially? I can't tell if it is or not), because I can't actually get to the point in time when a new good is made free. Things always seem to collapse for me at around 3-4k cycles. Can you look into whatever is causing this?

header/ConsumerGood.h
```diff
@@ -12,5 +12,6 @@ struct ConsumerGood : public Product {
 
     Good * corresponding_good = nullptr;
 	double mean_consumption_frequency;
+    bool paid = true;
```
Amittai-Aviram: Would you consider flipping the sense to `bool free = false;` or, even better, `bool public_sector = false;`?   Somebody just starting to look at our code might misunderstand what a "paid consumer good" means. (It could mean that the consumer has already paid for it or something.)  In Alex's discussions, he has referred to public items as "free" as a shorthand, but, of course, they're not free — _all_ goods are paid for.  The difference is whether the individual pays for the product out of his or her account or the collective pays for it out of its collective account, which pools the value of individual workers.

src/Distributor.cpp
```diff
@@ -30,29 +31,24 @@ Logger::Client Distributor::get_client_type() {
...
-        renew_distribution_plan(consumer_good);
-    }
 }
 
 void Distributor::add_to_catalog(Product * product) {
     catalog.insert(product);
     ConsumerGood * consumer_good = static_cast<ConsumerGood *>(product);
     Good * good = consumer_good->corresponding_good;
+    input_inventory[good] = 
     input_inventory[consumer_good] =
```
Amittai-Aviram: Reformat — or did you mean to delete this? The consumer good is no longer an input, right? The corresponding good is the (sole) input for the consumer good.

src/Firm.cpp
```diff
@@ -323,7 +323,7 @@ void Firm::assign_workers(Plan * draft_plan) {
 
 double Firm::predict_turnaround_time(Plan * plan, std::vector<Person *>& workers) {
     if (workers.empty()) {
-        return std::numeric_limits<double>::infinity();
+        throw std::runtime_error("Cannot predict turnaround time with 0 workers: " + plan->order->product->product_name);
```
Amittai-Aviram: Could you please rewrite this to use the product ID?
`" [...] workers: Product " + std::to_string(plan->order->product->id);`

src/Firm.cpp
```diff
@@ -405,7 +405,7 @@ void Firm::add_demand_signal(Product * product, double quantity) {
 void Firm::update_demands() {
     for (std::pair<Product * const, double>& demand : demands) {
         demand.second *=
-            (DEMAND_AVERAGING_WINDOW - 1.0) / DEMAND_AVERAGING_WINDOW;
+            1.0 - 1.0 / DEMAND_AVERAGING_WINDOW;
```
Amittai-Aviram: Isn't this the same as
`demand.second -= demand.second / DEMAND_AVERAGING_WINDOW;`
Maybe even
```
double current_average_demand = demand.second;
demand.second -= current_average_demand / DEMAND_AVERAGING_WINDOW;
```
(It's up to you — just trying to make sure the logic is clear to everybody.)

src/Person.cpp
```diff
@@ -70,8 +75,8 @@ void Person::train(std::unordered_map<Ability *, double>& target_abilities) {
 
 void Person::register_hours_worked(double hours_worked) {
     log_hours_worked(hours_worked);
-    account += hours_worked;
     busyness_this_time_step += hours_worked;
+    account += PriceController::get_instance()->get_fic() * hours_worked;
```
Amittai-Aviram: Yeah. The thing is that the remainder, `hours_worked * (1.0 - PriceController::get_instance()->get_fic()`, has to go someplace. In particular, it needs to go to a public fund, from which the cost of "free" goods needs to be drawn. Yes, the mathematical model shows that it shouldn't matter, but the modeling program needs to confirm that the books are balanced, I think.

src/Distributor.cpp
```diff
@@ -61,36 +57,42 @@ int Distributor::try_sell_goods(ConsumerGood * consumer_good, int quantity, Pers
     if (available < quantity) {
         log_shortfall(consumer_good->id, quantity - available);
     }
-    double cost = available * consumer_good->price_per_unit;
+    double cost = 0.0;
+    if (consumer_good->paid) {
+        cost = available * consumer_good->price_per_unit;
+    }
```
Amittai-Aviram, line 69: Right around here is where you'd charge the public fund instead of the individual's account. Please see my comment on `Person::register_hours_worked`.

src/PriceController.cpp
```diff
@@ -9,12 +9,26 @@
...
+
 PriceController * PriceController::get_instance() {
     static PriceController * instance = new PriceController;
     return instance;
 }
 
-PriceController::PriceController() {}
+void PriceController::on_time_step() {
+    double decay = 1.0 - 1.0 / FIC_AVERAGING_WINDOW;
+    paid_consumer_goods_value *= decay;
```
Amittai-Aviram: Similar to what I said about demands decaying. I'd prefer this:
```
double decay = goods_value / FIC_AVERAGING_WINDOW;
paid_consumer_goods_value -= decay;
```
But, as always, I am not going to throw a fit if you insist on your way.

src/PriceController.cpp
```diff
@@ -9,12 +9,26 @@
...
+    all_consumer_goods_value *= decay;
+    fic = 1.0;
+    if (all_consumer_goods_value) {
+        fic = paid_consumer_goods_value / all_consumer_goods_value;
```
Amittai-Aviram: Alex's (and your) postings have
$$\mathrm{fic} = \frac{\lambda \mathbf{n}_p}{NK}$$
But is `all_consumer_goods_value` equal to `NK`? These should be approximately equal, or should constantly converge, but is it safe to treat them as identical here? And would it be safe for a society to build an automatic infrastructure based on that assumption?

Amittai-Aviram (commented):
This looks good to me so far, except that I (at least personally) would prefer to have an explicit accounting of the public fund. I don' tthink it should take a whole lot of work to implement that. I presume that the public fund (or account) should belong to `Society`. If the product is "free," its price is charged to the public fund.

Other than that, could you please just respond to my comments and questions? Thanks!

src/Society.cpp
Amittai-Aviram: @alexbcreiner0 — With utmost respect, for  now, I'm afraid that I beg to differ. First, Society does not model a mechanism, it models the general state of the whole society, especially its initial state.  Computing the FIC requires a specific mechanism. And, to me, it makes sense for that mechanism to belong to the same agent to which firms report their expenditures. In other words, the Price Controller is the accounting agency for the society, so it makes sense to compute the FIC there, based on current prices. But I am not generally a fighter, so — I'm OK with whatever.

alexbcreiner0: I think the place we differ is that I don't really see the price controller as an accounting agency at all. Agencies make decisions, but our PriceController is not doing that. It's just a simple algorithm - in my mind something that could be implemented with some kind of distributed blockchain technology allowing for maximal trust and understanding in the calculations. Let's discuss at the coming meeting.

src/PriceController.cpp
alexbcreiner0: @Amittai-Aviram Yes, the two quantities should approximately equal assuming that our price controller works properly (which it does). As soon as Devin brought this up during our meeting when I presented it, I realized this could be a very clean and potentially useful way to express it. This should be a sound implementation.

header/ConsumerGood.h
lennyyyyyyyy: Done.

src/Distributor.cpp
lennyyyyyyyy: This corresponds with the new separate demand signals in the Distributor for the good and consumer good. Essentially, I need to start both with a surplus.

src/Firm.cpp
lennyyyyyyyy: Sure.

src/Person.cpp
lennyyyyyyyy: Done.

src/PriceController.cpp
lennyyyyyyyy: Done.

src/Distributor.cpp
lennyyyyyyyy: Done.

src/Society.cpp
lennyyyyyyyy: For now, I'll have no public sector goods at time = 0, because the system needs a bit of time to figure out the FIC. It's actually not too much of an issue since people also don't work for some time in the beginning, but I think I'll include this feature in the option to start at equilibrium.

header/PriceController.h
```diff
@@ -17,10 +17,14 @@ class PriceController : public Agent {
         void report_distribution(ConsumerGood *, int quantity);
         double get_fic();
     private:
+        double fic = 1.0;
```
Amittai-Aviram: Let's please group all the data fields together and, likewise, all the private function signatures together — with the constructor first among the latter. I think I'd be in favor of putting the function declarations above the data fields.

Amittai-Aviram (changes requested):
These changes look great! I'm asking for one more tiny change, and let's also go over the whole thing when we meet.  Great work!

header/PriceController.h
lennyyyyyyyy: Done.

src/Distributor.cpp
Amittai-Aviram: OK, I see.

Amittai-Aviram (approved):
LGTM

lennyyyyyyyy:
The change in the flow of distributor demand did cause some more instability, because there was a new lag between the demand of the producer good versus the demand of the consumer good (which we forced to be equal before). Increasing the surplus seemed to help, but I decided that starting at a predicted demand at 0 is sort of arbitrary anyway. A more realistic solution is to start Firms with a predicted demand normally distributed with some variance around the equilibrium demand. I set this variance right now as `DEMAND_PREDICTION_VARIANCE = 0.05`
Results:
image

src/Society.cpp
alexbcreiner0: I agree with this. No goods should be free at first. Was just making sure.

alexbcreiner0 (changes requested):
This is fantastic work Devin. My only requested changes, aside from the discussion we'll have tomorrow about the PriceController, is that we log deductions from the public fund separately from additions to it. This way, we can separately track public revenue and public expenses. I still like the log_public_fund, so don't think it should be deleted. I'll certainly make use of it occasionally for debugging purposes. I really like the decision to make demand normally distributed with a controllable standard deviation.

src/PriceController.cpp
```diff
@@ -62,3 +64,63 @@ void PriceController::update_price(Plan * plan) {
...
+}
+
+void PriceController::update_fic() {
+    double public_sector_distribution_value = 0.0;
+    double all_distribution_value = 0.0;
+    for (std::pair<ConsumerGood * const, double>& consumer_good : total_consumer_good_distribution_value) {
+        double decay = consumer_good.second / FIC_AVERAGING_WINDOW;
+        consumer_good.second -= decay;
+        all_distribution_value += consumer_good.second; 
+        if (consumer_good.first->public_sector) {
```
alexbcreiner0, line 86: Did this change? Why? I don't think this is working properly. It looks like you are double dividing by the `FIC_AVERAGING_WINDOW`. I tested the simulation for 30k cycles and the FIC never seems to reach 1, and I think this might be why.  Also this isn't really a big deal, but why did we shift from tracking the value of the paid goods to the value of the public sector goods? 

EDIT: It seems like the double dividing isn't creating issues, at the very least:
image

The FIC goes to zero, and all value production eventually shifts into the public sector. However, I find it strange that despite this, workers are still receiving and spending labor credits:
image

I would expect a complete flatlining of accounts at this point. Do we have an explanation for this? Seems like there must be a bug to me somewhere.

Also, I think that it shouldn't be referred to as consumer goods value. It's the value of the net product. These are the same now, but could diverge in the future (for example, a bridge is part of the net product, but it isn't a consumer good).

alexbcreiner0 (changes requested):

src/PriceController.cpp
lennyyyyyyyy: Hopefully the naming is better now. Looking at the raw logs, it seems like the FIC does go exactly to 0, and accounts don't move over long periods of time:
image
image
image

So I'm tempted to say it's a graphing issue.

alexbcreiner0: Hmm. I wouldn't necessarily assume from these logs that the account of person 91 is never changing. Especially compared to my own graph here it seems to always be settling back down to the same number, and so you could easily have caught it at two 'quiet' moments. Still though, we can go ahead and merge despite this I think, and worry about it later.

alexbcreiner0 (approved):

PR #237, opened 2026-06-09
Implement machine life cycle

### LINKED ISSUE #203: Implement the replacement cycle of machines

### LINKED ISSUE #231: Remove `raw_materials_remaining`
See Issue #230 : our logic does not use `raw_materials_remaining`, so we should eliminate it everywhere. It is a confusing relic from obsolete code.

Amittai-Aviram:
# Assumptions
- A producer acquires all the machines it needs to make its products without having to pay for the initial machines. (It does have to pay to replace the machine when the machine wears out.)
- The useful lifetime of each machine is a known property of that machine.
- A firm has (and needs) only one of any kind of machine.
- A machine is used during every hour of execution of a plan to produce something. I.e., it is in use throughout all the time that workers are working.
- Only one person can use a machine at a time.

# Sequence of Events
- When a producer adds a product to its catalog (`Producer::add_to_catalog`)
  - It makes sure that it has all the machines it needs to make that product.  If it does not already have it
    - It gets the machine for free.
    - It starts tracking the age (remaining lifetime) of the machine (in `Firm::machine_lifetime_remaining`).
    - It records that the machine is not on order (yet).
- When a firm drafts a plan
  - It includes the cost of needed machinery in the plan's budget. This is the cost of the machine per hour of its lifetime times the number of hours that the workforce as a whole will be working (with one person at a time using the machine).
- When a firm starts a plan, it moves the labor time credits in the plan's `machinery_cost` budget line into the pooled input account.
- When a firm moves a plan forward one step
  - Each machine involved in the plan ages by one hour.
  - The firm checks whether it is time to re-order the machine &mdash; i.e., whether the machine's remaining usable life is less than or equal to some arbitrary portion of its lifetime to be considered the "last" portion, set as the constant `MACHINE_REPLACEMENT_TIME` in `headers/Constants.h`).
- If necessary, the firm re-orders the machine.
  - The order is for one machine at a time.
  - The machine is marked _on order_. The machine would not be re-ordered if it is already on order.
  - The requested turnaround time is the same as the time remaining in the macine's lifetime at which the machine should be re-ordered.
  - When the machine comes in, its (new) lifetime is simply added to the remaining lifetime of that machine, with the idea that the firm continues to use the old machine until it no longer works, and then swaps the new machine in, instantly.
  - The firm pays for the new machine out of its pooled input account.

# Other Changes
This change set also includes a few corrections and improvements.
- Deletion of some objects no longer in use so as to avoid some memory leaks.
- Correction of the `Plan` type's fields and how they are handled.
  - There is no need for `raw_materials_remaining`, `total_hours`, or `total_hours_remaining`. These are relics from previous (and possibly incorrect) implementations.
  - Nothing should be transferred into the pooled input account except the machinery and raw materials budget lines in `Firm::start_plan`.

Resolves #203 
Resolves #231


src/Firm.cpp
```diff
@@ -57,18 +57,24 @@ double Firm::get_inventory_level(Product * product) {
...
     pooled_input_value -= transaction_amount;
     plan->firm->receive_payment(plan, transaction_amount);
     log_shipment_received(order->product, order->quantity);
-    log_inventory_level(order->product, input_inventory[order->product]);
 }
 
 void Firm::receive_payment(Plan * plan, double transaction_amount) {
     plan->prd += transaction_amount;
-    pooled_input_value += transaction_amount;
+    // pooled_input_value += transaction_amount;
```
lennyyyyyyyy: Remove the commented line.

src/Producer.cpp
```diff
@@ -99,6 +105,11 @@ Order * Producer::draft_plan_and_return_order(const Order * order) {
 
 void Producer::drop_order(Firm * customer) {
     log_dropped_order(customer_to_draft_plan[customer]->order);
+    /*
+    Plan * dropped_plan = customer_to_draft_plan[customer];
+    delete dropped_plan->order;
+    delete dropped_plan;
+    */
```
lennyyyyyyyy: Remove comments.

header/Firm.h
```diff
@@ -29,13 +29,10 @@ struct Plan {
 	// dependent/output fields	
 	double predicted_turnaround_time;
     double machinery_cost;
-    double labor_hours;
     double raw_materials;
-    double total_hours;
+    double labor_hours;
     double prd;
     double labor_hours_remaining;
```
lennyyyyyyyy: I believe there should be `machinery_cost_remaining` that is considered in accounting. The machinery cost is similar to the labor hours in that it's not entirely certain at the start of the plan. It will vary depending on the workers' productivity.

lennyyyyyyyy (commented):
General comments about the design of the machine life cycle:

- It doesn't seem like a good idea to decrement the lifetime of machines as the plan executes, for the same reason we don't decrement raw materials from the inventory as the plan executes. Other plans can't rely on an inventory of raw materials (or machine lifetime) that is ticking down at a rate they don't know about. That's why we take away the raw materials all at once. (As a bonus, we only need to call check_and_reorder at the start of a plan, for both raw materials and machines.)
- I know you talked about this somewhat, but could you clarify why the reordering process for machines looks so different than for inputs? It seems to me that the thing we want to model with machines is that it's a fixed cost - as the amount of workers and the size of the plan scales up, the machine cost does not. However, there shouldn't be anything special about reorders. Orders that use the machine should generate demand signals for the machine, which determines its reorder threshold. You may have multiple orders out for a machine, or you may not. It seems to me that `machine_lifetime_remaining` provides the same function as `input_inventory`, and `machine_on_order` the same as `product_to_outbound_orders`, but more restrictive.

src/Firm.cpp
Amittai-Aviram: Ugh! Done. Thanks. (Sorry!)

src/Producer.cpp
Amittai-Aviram: Done — thanks!

header/Firm.h
Amittai-Aviram: I've restored the accounting of raw materials budgeted value used on every time step and added the same for machinery.  At the end of the plan, we add those values back to the pooled input account. But note that the Price Controller still only averages labor hours and does not account for discrepancies between actual input or machinery usage and the respective planned expenditures. I think that that adjustment would belong to a separate issue.

Amittai-Aviram:
> General comments about the design of the machine life cycle:
> 
> * It doesn't seem like a good idea to decrement the lifetime of machines as the plan executes, for the same reason we don't decrement raw materials from the inventory as the plan executes. Other plans can't rely on an inventory of raw materials (or machine lifetime) that is ticking down at a rate they don't know about. That's why we take away the raw materials all at once. (As a bonus, we only need to call check_and_reorder at the start of a plan, for both raw materials and machines.)

Sorry — could we perhaps discuss this?

Just for the moment, I will try to list the conceptual differences between fixed capital ("machines") and liquid capital ("raw materials"):
- IRL, fixed capital items are (a lot) more expensive than input items.
- IRL, fixed capital items last much longer than either input items or plan cycles.

Because machines are more expensive than input products, one tends to limit how may of them one has that one is not actually using.

Because machines outlast plans, each plan has to budget an amortized cost for the machine.

Our model is not accurate with respect to these considerations, because we have not taken care ot price machines higher than input goods, but their lifetimes are generally scaled at longer than plan cycles.

Amittai-Aviram:
Closing this PR in favor of #238 .

PR #238, opened 2026-06-12
Implement machine life cycle

### LINKED ISSUE #203: Implement the replacement cycle of machines

Amittai-Aviram:
Resolves #203 
Supersedes earlier PRs #237 and #229 .

# Assumptions
- A producer acquires all the machines it needs to make its products without having to pay for the initial machines. (It does have to pay to replace the machine when the machine wears out.)
- The useful lifetime of each machine is a known property of that machine.
- A firm has (and needs) only one of any kind of machine at a time.
- A machine is used during every hour of plan execution. I.e., it is in use throughout all the time that workers are working.
- Only one worker can use a machine at a time.

# Sequence of Events
- When a producer adds a product to its catalog (`Producer::add_to_catalog`)
  - It makes sure that it has all the machines it needs to make that product.  If it does not already have any needed machine, it gets the machine for free.
  - It starts tracking the remaining useful lifetime of the machine as the "quantity" of the machine in the firm's input inventory.
  - It adds a demand signal for that machine (as an "input").
- When a firm drafts a plan, it includes in the plan's budget the predicted cost of needed machinery. This is the sum over machines needed of the cost of the machine per hour of its lifetime times the number of hours that the workforce as a whole will be working (with one person at a time using the machine).
- When a firm starts a plan
  - It moves the labor-time credits in the plan's `machinery_budget`, together with its `raw_materials_budget`, into the pooled input account.
  - It checks and possibly reorders any inputs running low, including machines whose remaining lifetimes are low. The formula for deciding whether the value is low enough to require reordering is the same for machines as for other inputs, except that, if the machine's remaining lifetime is below the threshold, we order at least one new machine. Any new machine's lifetime is simply added to the current lifetime of that machine in the input inventory. (The firm will continue using the old machine until it wears out and will then switch instantly to the  new machine.)
- When a firm moves a plan forward one step, we track how much of the machinery, raw materials, and labor budgets were actually used during that time step.
- At the end of the plan,
  - Any value left in the budget for machinery and raw materials that was not actually used is returned to the pooled input account.
  - The Price Controller uses the actual labor value consumed in recomputing the average.

# Other Changes
This change set also includes a few corrections and improvements.
- Deletion (`delete`) of some heap-allocated objects no longer in use so as to avoid some memory leaks.
- Renaming of the `Plan` type's fields and changes to how they are used. E.g., rather than tracking `raw_materials_remaining`, etc., we track the values actually used in the plan and then transfer those values appropriately.
  - Nothing should be transferred into the pooled input account except the machinery and raw materials budget lines in `Firm::start_plan` and `Firm::end_plan`.
  - I have removed several obsolete fields from `Plan`, including `total_hours` and `total_hours_remaining`.
- We no longer use `Product::product_name`, only the product's `id`.


Amittai-Aviram:
Screenshot 2026-06-12 at 12 49 45 AM

src/Firm.cpp
```diff
@@ -174,7 +173,13 @@ void Firm::check_and_reorder_input(Product * product) {
...
+    if (pending_inventory >= threshold || !threshold) {
+        return;
+    }
+    double reorder_quantity = threshold * FIRM_REORDER_MAX_PROP;
+    if (product->product_type == Product::ProductType::TYPE_MACHINE) {
+        reorder_quantity = std::max(std::ceil(reorder_quantity), 1.0);
+    }
     Order * order = new Order(
             product,
             threshold * FIRM_REORDER_MAX_PROP,
```
lennyyyyyyyy, line 181: The `reorder_quantity` variable you defined should go here. Also, it'd be good to define a variable dependent on this called `reorder_deadline`, and define this as `pending_inventory / demands[product] * std::max(reorder_quantity / threshold, 1.0)`. The logic is that this is the same computation in the case for normal goods, where the reorder amount is always `threshold * FIRM_REORDER_MAX_PROP`. For machines, whose reorder may be beyond that, or even beyond the threshold, a more lenient reorder_deadline should be given, but not beyond the expected amount of time for the current machines to run dry, which is `pending_inventory / demands[product]`

src/Firm.cpp
```diff
@@ -191,49 +196,53 @@ void Firm::start_plan(Plan * plan) {
...
+	for (std::pair<Good * const, double>& input : product->inputs_per_unit) {
         double required_input = input.second * plan->order->quantity;
         remove_input_from_inventory(input.first, required_input);
         check_and_reorder_input(input.first);
 	}
-    pooled_input_value += plan->raw_materials;
-    plan->raw_materials_remaining = 0;
+    double expected_plan_duration = plan->labor_budget / plan->workers.size();
+    for (Machine * machine : plan->order->product->machines_needed) {
+        input_inventory[machine] -= expected_plan_duration;
```
lennyyyyyyyy: You should be able to use the general `remove_input_from_inventory` here. After merging with my branch, this method will also invoke the corresponding demand signal.

src/Firm.cpp
```diff
@@ -191,49 +196,53 @@ void Firm::start_plan(Plan * plan) {
...
+    plan->machinery_budget = plan->raw_materials_budget = 0.0;
     plan->order->status = Order::ORDER_IN_PROGRESS;
 }
 
 void Firm::move_plan_forward_one_step(Plan * plan) {
-    double ideal_quantity_produced = calculate_quantity_produced_from_worker_suitability(plan);
-    double quantity_produced = std::min(ideal_quantity_produced, plan->quantity_remaining);
+    double expected_quantity_produced =
+        calculate_quantity_produced_from_worker_suitability(plan);
+    double quantity_produced =
```
lennyyyyyyyy: There should be another check here to make sure that the labor hours used don't cause `machinery_value_used` to exceed `machinery_budget`.

header/Firm.h
```diff
@@ -20,24 +20,19 @@ class Producer;
...
+    unsigned int local_work_hours_daily = 0;
+	double predicted_turnaround_time = 0.0;
+    double machinery_budget = 0.0;
+    double raw_materials_budget = 0.0;
+    double labor_budget = 0.0;
+    double debt = 0.0;
+    double machinery_value_used = 0.0;
+    double raw_materials_value_used = 0.0;
+    double labor_value_used = 0.0;
+    double quantity_remaining = 0.0;
```
lennyyyyyyyy, line 36: Any reason you restructured most of the fields to tick up, while you kept quantity_remaining ticking down? No need to change it, just curious.

src/Firm.cpp
lennyyyyyyyy: I think there should be a check on both the quantity and the machine budget. You shouldn't be allowed to use more than your machine budget, even if you still have quantity that remains.

src/Firm.cpp
```diff
@@ -191,49 +196,53 @@ void Firm::start_plan(Plan * plan) {
...
     }
-
-    double raw_materials_used =
-        plan->raw_materials_remaining *
-        quantity_produced /
-        plan->order->quantity;
-	//pay workers
-    double labor_hours_per_worker = quantity_produced / ideal_quantity_produced;
+    double labor_per_worker = quantity_produced / expected_quantity_produced;
+    plan->machinery_value_used +=
```
lennyyyyyyyy, line 232: If I'm not mistaken, machinery value isn't in units of lifetime, but in units of labor value. This should then mirror the raw material incrementation like `(plan->machinery_budget / plan->order->quantity) * quantity_produced`

src/Firm.cpp
lennyyyyyyyy: This is also a bit tricky, but I think we should store the machines in the input_inventory in units of whole machines, and not lifetime. This will make demands, thresholds, and order sizes all have the same unit. 

Currently, inventories are in lifetime, which causes demand signals to be in lifetime, which causes the threshold to be in lifetime. However, your reordering quantity is in units of machine, which makes sense, but when you get the shipment back, it'll add the machine as if it is _1 hour of lifetime_.

In short, maybe this should be `input_inventory[machine] -= expected_plan_duration / machine->lifetime;`

src/Producer.cpp
```diff
@@ -35,6 +36,8 @@ void Producer::add_to_catalog(Product * product) {
     catalog.insert(product);
     for (Machine * machine : product->machines_needed) {
         machines.insert(machine);
+        input_inventory[machine] = machine->lifetime;
```
lennyyyyyyyy: Same discussion about the way to store machines in the inventory. Units of lifetime are complicated. I would suggest `input_inventory[machine] = 1;` Also, what's with the random demand signal during intialization?

src/Producer.cpp
lennyyyyyyyy, line 87: As part of the reordering process, to make Producers adaptable to an order size they may not have the resources to fill, they can send a return order, which the ordering Firm considers. 

For machines, we need to take care that the scaled down return order still has an integer machine quantity, and if we can't achieve that, we have to reject. Let me know if you need help from me here.

lennyyyyyyyy (changes requested):

src/Firm.cpp
Amittai-Aviram: First point — oops! — thanks. 
But I have reverted the whole function back to its original version (in `main`), without any special consideration for machines, in view of changing the accounting from lifetimes to units and the addition of a call to `Firm::add_demand_signal` inside `Firm::remove_input_from_inventory`.

src/Firm.cpp
Amittai-Aviram: Done.

src/Firm.cpp
Amittai-Aviram: This raises the question of what to do in that situation?  It means that the actual expenditures have been greater than what we have budgeted. One way to handle this is to set the debt for the plan to reflect the actual expenditures — instead of testing on every step whether we have exceeded the budget. The budget is an estimate, but the debt reflects real expenditures.  I think the next step would be to extend the Price Controller to recompute machinery and raw materials expenditures. (Right now, raw materials would be unaffected by the state of the workforce, but, as you have pointed out, the machinery cost is indeed affected. But we could model process and productivity changes to involve changes in the use of raw materials.) With this change, you can expect that `debt` will not generally arrive at 0.0 after payment.

header/Firm.h
Amittai-Aviram: Yes. It's a matter of clarity. The budget is what you apportion for an expense, and then you take the cost out of that budget. In our case, we take the expected machinery and input costs immediately out of the budget and spend the value on the pooled input account. Then, as we execute the plan, we need to track our actual expenses, which _accrue_ from 0 upwards. I got to this point by first trying to find better names for the fields that were called "raw_materials_remaining," etc. — that name does not describe what it actually means — it does not represent remaining raw materials or even the value thereof, or even the value of the budget remaining, since we've already spent that budget item.  I have that problem even with "quantity_remaining," but it's not so bad — it's the quantity remaining _to be produced_.

src/Firm.cpp
Amittai-Aviram: See above. How would you handle this situation?  The budget originally comes out of debt, so to speak, so adding to the debt would be a way to account for the cost overrun. Now, the downstream consumer will always pay the current price for the goods, so, in the case of a cost overrun, the debt will remain on the producer's books forever. (Likewise, if the producer was more efficient than average, the debt value will be made positive after the goods are sold, so the company will have a surplus on the books forever.) We can then think about control mechanisms that use the debt — for instance, in the aggregate, to adjust expectations and budgets, and, in the particular case, to exert some sort of pressure on a firm to produce more efficiently.

src/Firm.cpp
Amittai-Aviram: What I _was_ doing there was incrementing the amount (in lifetime hours) of the machine that was used on this time step. But I've redone the whole sequence to account for machines in units rather than lifetimes.

src/Producer.cpp
Amittai-Aviram: Done.

src/Producer.cpp
Amittai-Aviram: I'm not sure I understand what you mean. Do you mean the case of orders for machines as products, that the proposed order quantity must be an integer? Or do you mean that the proposed maximum quantity must be an integer? In either case, `Producer::get_max_order_quantity` returns an `int` value, and, in the new code therein that bounds the quantity by machine capacity, likewise, we compute an integer value (the floor).

src/Firm.cpp
```diff
@@ -60,24 +60,26 @@ void Firm::receive_shipment(Plan * plan) {
...
 bool Firm::remove_input_from_inventory(Product * product, double quantity) {
     if (!input_inventory.count(product) || input_inventory[product] < quantity) {
+        check_and_reorder_input(product);
         return false;
     }
     input_inventory[product] -= quantity;
+    add_demand_signal(product, quantity);
     log_inventory_reduction(product, quantity);
     log_inventory_level(product, input_inventory[product]);
+    check_and_reorder_input(product);
```
lennyyyyyyyy, line 82: You can simplify the logic by putting this at the top of the function.

src/Distributor.cpp
lennyyyyyyyy: You have to remove this if you're adding the demand signal in `remove_input_from_inventory`. Otherwise you'll get duplicate demand signals. Make sure this is gone after resolving conflicts.

src/Producer.cpp
```diff
@@ -35,6 +36,8 @@ void Producer::add_to_catalog(Product * product) {
     catalog.insert(product);
     for (Machine * machine : product->machines_needed) {
         machines.insert(machine);
+        input_inventory[machine] = 1.0;
+        add_demand_signal(machine, 1.0);
```
lennyyyyyyyy: I still don't understand the motivation behind this demand signal.

src/Producer.cpp
lennyyyyyyyy: Same thing about duplicate demand signals. Once you resolve conflicts, this method should be gone.

src/Firm.cpp
```diff
@@ -191,48 +193,57 @@ void Firm::start_plan(Plan * plan) {
...
     log_ended_plan(plan);
     plan->order->status = Order::ORDER_FINISHED;
+    plan->debt = -(plan->machinery_value_used + plan->raw_materials_value_used +
+            plan->labor_value_used);
+    double labor_use_adjustment = plan->labor_value_used - plan->labor_budget;
+    double plan_duration_adjustment = labor_use_adjustment / workers.size();
+    for (Machine * machine : plan->order->product->machines_needed) {
+        double machine_use_adjustment =
+            plan_duration_adjustment / machine->lifetime;
+        remove_input_from_inventory(machine, machine_use_adjustment);
```
lennyyyyyyyy: I don't know about removing excess machine lifetime after the fact... what happens when `remove_input_from_inventory` returns false? In all other cases, we're guaranteed to have enough inputs to remove, but here we're not.

On the other hand, wouldn't stopping plan execution when our machine budget runs out be safer? The producer nor the ordering firm lose anything by doing so - we can always put any excess raw material budget back into inventories, and the ordering firm may get a little less quantity than expected, but in less time. In any case, the rate of production should be the same, and the price they pay is determined by the actual quantity produced, not the quantity ordered.

lennyyyyyyyy (changes requested):

src/Firm.cpp
Amittai-Aviram: I have taken your suggestion in `Firm::remove_input_from_inventory` to call `check_and_reorder` first, so, if that function returns false, it will not do so before reordering, which puts the reorder in place before the next plan starts, which should be appropriate.  Bear in mind that the adjustment may be negative, meaning that we have used _less_ of the machine than expected, so the result of `remove_input_from_inventory` would be to bring the machine's lifespan (as expressed by the machine as a unit) back up, because we have not used it as much as we had expected. 
I think I might be a little confused by your terms. What we could theoretically run out of is not machine budget but actual machinery (lifespan). The logic of checking and reordering should prevent that from ever happening, however, just as it does for other inputs. We have already agreed that there is no limit to how much of a machine is stockpiled, just as there is none for other inputs. The rule is only that we _use_ only one instance of a given machine at a time.
All in all, I'm afraid I do not understand your concern here. I'm afraid I'd need you to explain it a bit further. All I am doing here is making the remaining lifetime of the most recently used machine match its actual use rather than the use that we had predicted at the start of the plan. It's just ensuring that our representation of the machine's use over time is accurate.

src/Firm.cpp
Amittai-Aviram: By "this," do you mean the call to `check_and_remove_input`? If so, I was about to put it first, and remove the call to it in the `if`-clause (for if there is none or not enough of the input in the inventory), but then I realized that this is a bad idea, or at least so it seems to me. If you are out of something, you do need to reorder it ASAP. Otherwise (in the normally expected case), you should remove it and then log that fact and the current level of the inventory — and _then_ check and reorder, since you have just changed the level. In other words, you need to change the level before you log the change, and you need to log the change before you check and reorder.  Am I missing something?

src/Distributor.cpp
Amittai-Aviram: I cannot find it, and, indeed, there is no code quotation in this comment, and I cannot find the comment in the GitHub diff, so I am going to guess that it's gone as a result of the merge, even though I do  not recall deleting a call to `add_demand_signal`. (But I _did_ delete at least one call to `check_and_reorder_input`, which calls `add_demand_signal`.)

src/Producer.cpp
Amittai-Aviram: Removed. I think the original idea was to make a record ahead of time of the need eventually to replace the machine, but that was before I had revised `check_and_reorder_inputs`, etc.  Anyway, it's gone!

src/Producer.cpp
Amittai-Aviram: Yeah, again, I think it's gone.

src/Firm.cpp
lennyyyyyyyy: You're right. Sorry!

src/Firm.cpp
lennyyyyyyyy: My point is that if `remove_input_from_inventory` returns false, sure, it reorders more of the machine, but `machine_use_adjustment` is never subtracted. And then we just move on with life as if we had. However unlikely it is for this to occur, I don't think we can ignore a case where we may ignore the cost of production.

> I think I might be a little confused by your terms. What we could theoretically run out of is not machine budget but actual machinery (lifespan)

Right. And that's pretty much what I want to avoid. Limiting ourselves to the budget we set for ourselves is my alternative to solve the issue.

Amittai-Aviram: If it returns false, then we know (as I have written it now) that `check_and_reorder_input` has been called. So then, in that case, in `end_plan`, we can directly deduct the adjustment, which may result in a negative value for the machine-as-input, which will then become a positive value when the order for the machine is received, unless `send_order` in `check_and_reorder_input` returns false, in which case we log it as, essentially, an error.  In other words, if we see that happen, we should investigate it as a system failure. That's one way of handling this.

The other way, as far as I can tell, is what you suggested, which is to deliver less than the order required. There are two ways to go about this:
- Keep track on every time step (as I did in a previous draft) how much of the machines we have actually used, and stop producing if any machine resource reaches 0. This inovles a bit of re-engineering to support stopping a plan early.
- Recompute everything in `end_plan` — first compute how much of the machinery was actually used, then check whether we have reached 0, and, in that case, revise everything: the quantity actually produced, the labor actually used — and even the inputs used, since we have a revised quantity produced.

So there are three options. Which do you favor? Do you have any other options in mind?

lennyyyyyyyy: > So then, in that case, in end_plan, we can directly deduct the adjustment, which may result in a negative value for the machine-as-input...

Where is this happening? As far as I can tell, the behavior of `remove_input_from_inventory` doesn't allow negative inventories. Not to mention that I think modeling a negative inventory that we "promise" will get refilled later has its own issues. 

The third option isn't very logical to me. Why figure out how much work we "overdid" when we can just stop work whenever we reach any limiting condition of our budget? For me, the second option is preferred, and as I said before, this doesn't negatively impact productivity at all. The plan produces a little less quantity, uses less inputs, and finishes a little sooner. And it's safer.

Amittai-Aviram: OK

src/Firm.cpp
Amittai-Aviram: Addressed in the latest change.

src/Firm.cpp
Amittai-Aviram: I have implemented a per-plan input inventory. The firm moves the inputs and machine capacity from its general inventory into the plan inventory at plan start, and decrements the content of the inventory on every time step (in `move_plan_forward_one_step`). It also checks whether any inventory item is out of stock, and, if so, terminates the plan. At plan end, any remaining inventory stock (and machine capacity) is returned to the general inventory. The accounting is as before, except that we do not need to check the machinery expenditure against the machinery budget anymore.

src/main.cpp
```diff
@@ -220,11 +220,23 @@ void set_params(int argc, const char ** argv, SimArgs& args) {
                 }
                 break;
             }
+            case ArgType::kInitPrices: {
```
lennyyyyyyyy, line 221: This flag shouldn't come after `long value` and `double dvalue` are read in, since it takes a string. This ends up trying to read the string as an integer, incrementing the index, and trying to read the next flag as the price mode.

src/main.cpp
```diff
@@ -71,16 +72,22 @@ void set_params(int argc, const char ** argv, SimArgs& args) {
...
+        if (arg == "-u" || arg == "--usage") {
+            print_usage();
+            exit(EXIT_SUCCESS);
+        }
         if (arg == "-j" || arg == "--json") { //"j" needs no additional value
             args.json = true;
             continue;
         }
+        if (valid_args.at(arg) == ArgType::kInitPrices) {
+        }
```
lennyyyyyyyy: Not sure what this empty if statement is.

src/PriceController.cpp
```diff
@@ -26,11 +26,21 @@ unsigned int PriceController::get_id() {
...
+                std::to_string(product->id)); 
+    }
     int now = Sim::get_current_time_step();
     int end_time = now - PRICE_AVERAGING_WINDOW;
     if (plan_history.count(product) && 
             plan_history[product].begin()->second <= end_time) {
+        Plan * old_plan = plan_history[product].begin()->first;
         plan_history[product].erase(plan_history[product].begin());
+        delete old_plan;
     }
```
lennyyyyyyyy: Not a current diff, but shouldn't this be a while loop, popping plans off the start until reaching a plan within the averaging window?

Additionally, I'm getting segmentation faults from plan_history[product] being empty. While plan_history may have a product entry, the vector of plans may be empty.

src/Firm.cpp
lennyyyyyyyy: This is now a duplicate `end_plan` call, which causes duplicate plans being reported and segmentation faults in the PriceController (after it deletes the first copy from the heap and tries the access the second copy)
Image

src/Firm.cpp
```diff
@@ -192,47 +192,74 @@ void Firm::start_plan(Plan * plan) {
...
+    Product * product = plan->order->product;
+	for (std::pair<Good * const, double>& input : product->inputs_per_unit) {
+        Product * input_product = input.first;
+        double amount_used = input.second * quantity_produced;
+        plan->input_inventory[input_product] -= amount_used;
+	}
+    for (Machine * machine : product->machines_needed) {
+        plan->input_inventory[machine] -= 1.0 / machine->lifetime;
+    }
+    for (std::pair<Product *, double> input : plan->input_inventory) {
```
lennyyyyyyyy: I'm getting some slight floating point rounding issues here - basically, `plan->quantity_remaining` is 0, but there are roundoff amounts of input left in `plan->input_inventory`. I recommend calculating (and capping) `quantity_produced` using the input_inventory instead of using `quantity_remaining`. It's the condition to end the plan anyway, so it's better to have the roundoff on `quantity_remaining` instead, and it generalizes to machines:

```
quantity_produced = std::min(quantity_produced, expected_quantity_produced * input_inventory[machine] / machine->lifetime);
```

lennyyyyyyyy (changes requested):
Debugging without machines for now - I found 3 seg fault bugs, small PriceController issue not strictly maintaining the window size, and a roundoff issue causing plans never to end.

src/main.cpp
Amittai-Aviram: I know. That is all OK. Do you have evidence that it does not work?  I tested it myself. Attempting to convert the next token into an integer or rational number does not consume the datum from the command line — it's still there in `argv`. As for `i`, that's why I read `argv[i]` as a string — `i` has already been advanced to the appropriate token. I reordered things this way to match the way the flags are ordered elsewhere inthe file and because there's not reason to treat this flag as a special case. Only `-j` is  a special case, because it is a standalone flag with no argument.

src/main.cpp
Amittai-Aviram: Deleted. I had forgotten to go back and delete this.

src/PriceController.cpp
Amittai-Aviram: Done. Even though it looks as if I am modifying the vector that I amreadig, it should be safe, because I am not iterating over it and modifying while doing so.  The iterator `begin` is newly constructed on every iteration of the `while`-loop.

src/Firm.cpp
Amittai-Aviram: Moved the check to `move_plan_forward_one_step`. I think that this should address the issue, because, this way, the same plan cannot be "ended" twice. The other way would have been to condition the original test (in `move_plans_forward_one_step`) on the status not being `ORDER_FINISHED`. But I think the logic is a bit clearer this way.  If you see a segmentation fault again with the same initial flags, please send me the flags.

src/Firm.cpp
Amittai-Aviram: About the suggestion — I'm sorry — I just do not understand what you mean.  `quantity_produced` is defined near the beginning of the function, but your suggested code refers to a specific machine, which means that it would belong in a loop iterating over machines.  But why would I set or reset `quantity_produced` inside that loop? I'm just totally confused.

About the problem — I'm not sure I understand that precisely, either. You're saying that `quantity_remaining` reaches 0, so the order is finished, but there are tiny amounts left in the inventory? You would expect that for machines — that's why we are doing this in the first place — because you can finish the order in some number of steps that may vary, so you won't necessarily use up all of your machines exactly as expected.  Usually, I take the term "roundoff error" to mean a value that is counted as 0 but is not really 0, but it doesn't sound as if that's what you mean.

lennyyyyyyyy: Your definition of "roundoff error" is what I mean - theoretically, we should've moved precisely enough raw materials into our input inventory to make exactly `order->quantity`. Instead, we're left with a minuscule amount of raw material that prevents `end_plan` from being called. You're right that it's likely for machines to be left over, but I was testing for normal behavior in a society devoid of machines first. Many plans never ended until I modified the inventory check to look like:
```
if (input.second < 0.01)
```
After this change, I got the normal behavior that we see in main.

This is just a bandaid fix that I used to demonstrate the issue, however. My above suggestion was to cap `quantity_produced` via inputs remaining, rather than by `quantity_remaining`. That way we're guaranteed to execute until the very last limiting condition of the plan inventory. The amount of machines in the inventory is a limiting condition too, which I tried to demonstrate in my first comment. In short, Firm.cpp::215 would go from:
```
double quantity_produced =
        std::min(expected_quantity_produced, plan->quantity_remaining);
```
to
```
double quantity_produced = expected_quantity_produced;
for (std::pair<Good * const, double>& input : product->inputs_per_unit) {
   quantity_produced = std::min(quantity_produced, plan->input_inventory[input.first] / input.second);
}
   quantity_produced = std::min(quantity_produced, expected_quantity_produced * plan->input_inventory[machine] / machine->lifetime);
}
```

Then at Firm.cpp::241, we can keep:
```
if (input.second <= 0.0)
```

I'd also like to point out that under the current design, the final plan inventory of a machine could reach -1.0 by the end of a plan, whereas if we consider our remaining machines in `quantity_produced`, the inventory must halt at 0.0.

Amittai-Aviram: OK, thank you. In your reference to line 241, did you mean this? —
`if (plan->quantity_remaining <= 0.0)`
I have moved that test upward in turn. I think that that should be OK: if the quantity remaining to be produced is 0, do not execute the next step, just end the plan.

Amittai-Aviram:
Closing this PR in favor of #257 .

PR #245, opened 2026-06-24
Use `std::list` for plan histories; fix warning from `main.cpp` `switch` statement

### LINKED ISSUE #244: Replace the arrays to store plan histories with queues in the Price Controller
The Price Controller computes the current price per unit of a product based on the average necessary labor time, of which the essential averaging mechanism is iterating over recent plans to get the mean living labor per unit. (The other components of the price are updated, in turn, when each of those items is produced, i.e., when a firm has completed a plan to produce it.) The plan histories are stored in a map from product to history of plans, where the history is implemented as an array (`std::vector`) of pointers to plans (`std::vector<Plan *>`). Whenever the oldest plan is too old, it is removed from the front of the array, and the new plan is appended to the end.  Such action is more efficient if the underlying representation is a linked list (`std::list`), not an array. (A queue is not exactly appropriate, because a queue does not support iteration through its contents.)

At the same time, when checking for plans that have aged out:
- Check whether the list is empty before seeking access to its front.
- Remove all expired plans, not only the first one encountered.

At the same time, address the compiler warning stemming from `src/main.cpp` regarding initial prices in the `switch` statement.

Amittai-Aviram:
Resolves #244 .

### DIFF
--- a/header/PriceController.h
+++ b/header/PriceController.h
@@ -1,7 +1,7 @@
 #pragma once
 
 #include <unordered_map>
-#include <vector>
+#include <list>
 
 #include "Agent.h"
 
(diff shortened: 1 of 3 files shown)


lennyyyyyyyy (approved):
LGTM.

Issue #210, opened 2026-05-21
Move `Plan` and `Order` definitions to separate files


Amittai-Aviram:
`Plan` and `Order` are currently defined in `Firm.h`, and their implementation code is in `Firm.cpp`, creating a bit of chaos and confusion. Let us treat each of these types as distinct.

In addition, we should try to clean these definitions up:
- Remove the obsolete `raw_materials_remaining` field. (Check the code base to make sure that it can be removed.)
- Rename `prd` as `debt` everywhere.
- Check whether we still need `total_hours`.

PR #248, opened 2026-06-25
Fixed firms calculating required number of people from non-local living labor values

### LINKED ISSUE #240: Firms are not using their local living labor per unit to predict number of workers needed
When drafting plans, firms need to use the living labor per unit numbers in two places: when predicting turnover times, and when determining the number of workers needed. Currently, they are only using their local version of this number when predicting turnover times. In the `predict_workers_needed` function, they are still using the global value for some reason. This is a big oversight.

alexbcreiner0:
Resolves #240

### DIFF
--- a/src/Firm.cpp
+++ b/src/Firm.cpp
@@ -262,7 +262,7 @@ bool Firm::is_within_work_schedule() const {
 int Firm::predict_workers_needed(Plan * plan) {
     return std::ceil(
             plan->order->quantity *
-            plan->order->product->living_labor_per_unit *
+            recorded_living_labor_per_unit[plan->order->product] *
             WEEK /
             Sim::get_work_days_weekly() / 
             plan->local_work_hours_daily /


lennyyyyyyyy:
This makes a lot of sense to me. But I think Prof. Aviram will be concerned that the price of raw materials isn't up to date at the time of budget creation. 

What I would suggest is a per-plan inventory, where the quantity produced is limited by what's in the per-plan inventory, and the plan ends once either the quantity produced or quantity remaining is 0. Then the Price Controller can take those amounts for each input and multiply by the current price. No backtracking from the budget necessary. In fact, I've already suggested this to Prof. Aviram for his machines issue. To prevent duplicate work, maybe you guys should talk.

src/PriceController.cpp
```diff
@@ -9,47 +9,43 @@
...
     int units = 0.0;
     double hours = 0.0;
     int workers = 0;
+    double materials = 0.0;
     for (std::pair<Plan *, int> entry : plan_history[product]) {
         Plan * plan = entry.first;
         units += plan->order->quantity - plan->quantity_remaining;
         hours += plan->labor_hours - plan->labor_hours_remaining;
         workers += plan->workers.size();
+        materials += plan->raw_materials - plan->raw_materials_remaining;
```
Amittai-Aviram: This gives you a sum of the _old_ prices of the raw materials. Meanwhile, any one of those raw materials may have had its price updated. That's the point of the current logic. (Note, BTW, that `raw_materials_remaining` had no meaning and has now been removed as of my latest change set [Issue #210 ].)

Amittai-Aviram (commented):
Alex, the change that you propose in the Price Controller defeats the point of the discussion we had just a few weeks ago and re-imposes a logic that keeps recycling potentially outdated prices for upstream inputs.

Amittai-Aviram:
> This makes a lot of sense to me. But I think Prof. Aviram will be concerned that the price of raw materials isn't up to date at the time of budget creation.

Yes, I noted that in my comment. I'm afraid we seem to be going in circles a bit — we reinstated my original logic of pulling the latest prices for inputs to make the new price after we had had a discussion about this, and you seemed to agree there and in subsequent conversations. 

As Devin mentioned, I might be implementing a per-plan inventory for in-kind accounting. I'm not totally sure yet, but that is a likely option.

When you get a chance, please pull the latest `main` and have a look at how the `Plan`, `Order`, and `PriceController` have changed.

alexbcreiner0:
> Alex, the change that you propose in the Price Controller defeats the point of the discussion we had just a few weeks ago and re-imposes a logic that keeps recycling potentially outdated prices for upstream inputs.

> This gives you a sum of the old prices of the raw materials. Meanwhile, any one of those raw materials may have had its price updated. That's the point of the current logic. (Note, BTW, that raw_materials_remaining had no meaning and has now been removed as of my latest change set [Issue https://github.com/BC-LTEWG/Labor-Time-Economy-Simulation/issues/210 ].)

Okay, sorry for this mess. Let me try to explain myself here. The code prior to those changes made in the week before I left for my trip looked like this 
```
#include "Constants.h"
#include "ConsumerGood.h"
#include "Distributor.h"
#include "Firm.h"
#include "Logger.h"
#include "Machine.h"
#include "PriceController.h"
#include "Product.h"
#include "Sim.h"

PriceController * PriceController::get_instance() {
    static PriceController * instance = new PriceController;
    return instance;
}

PriceController::PriceController() {}

void PriceController::update_price(Plan * plan) {
    Product * product = plan->order->product;
    int now = Sim::get_current_time_step();
    int end_time = now - PRICE_AVERAGING_WINDOW;
    if (plan_history.count(product) && 
            plan_history[product].begin()->second <= end_time) {
        plan_history[product].erase(plan_history[product].begin());
    }
    plan_history[product].push_back(std::make_pair(plan, now));
    int units = 0.0;
    double hours = 0.0;
    int workers = 0;
    for (std::pair<Plan *, int> entry : plan_history[product]) {
        Plan * plan = entry.first;
        units += plan->order->quantity - plan->quantity_remaining;
        hours += plan->labor_hours - plan->labor_hours_remaining;
        hours += plan->raw_materials - plan->raw_materials_remaining;
    }
    double machine_use_hours = hours / workers;
    double machine_hours_per_unit = machine_use_hours / units;
        double machine_cost_per_hour =
            machine->price_per_unit / machine->lifetime;
        price += machine_cost_per_hour * machine_hours_per_unit;
    }
    product->price_per_unit = price;
    Logger::log(
            Logger::SOCIETY,
            0,
            "new_price",
            LogPair("product_id", product->id),
            LogPair("price", price)
            );
}
```

At this time, I had no idea what you were talking about regarding raw materials costs changing causing problems, because that criticism was based on changes that had been made without my knowledge. I couldn't grasp what you were even telling me because, as far as I knew, at this time we were still looping over inputs per unit like we are now. I simply changed the price controller back to behaving how I originally conceived of it. I also found issues which you didn't mention, which distracted me further, such as this:

```
```

Hours here includes the hours of the raw materials, so the living labor per unit for each item was consistently being reset to the _value_ of the good. Whoever made this change seemingly did not fully understand the difference between living labor requirements and labor values. Anyway, because of my hurry to get ready for my trip, my surprise regarding how different the price controller was since I had last looked closely at it, and the issues I saw when I looked at it that you weren't bringing up, I never had a chance to process your actual issue. I just reverted it to the state I remember it being in, which happened to fix your issue, but only 'accidentally' because I actually understood it. Then I re-arrived at it myself when trying to resolve some knowledge issues which I was noticing while writing up our results. 

With that all said, I was still quite skeptical that your issues mattered, so I did some experimentation in python to test these methods out. It turns out that the historical time lag introduced by doing things this way is a real problem, and slows down convergence time significantly. I might make a discussion post showing some of my findings here, but I agree with you now that this method isn't good (update, I just finished doing this). I've also confirmed that if firms are reporting their materials used in kind, then we can use this information to replace the price controller's omniscient knowledge of the entries of $A$. So I'll wait for those updates. I'll also delete the raw materials issue since it is no longer relevant. Now, this request is only a single line of code which resolves one of the three issues, but it is still an important fix, so if you could approve it now I'll merge it in.

alexbcreiner0:
Alright, I removed the obsolete issue, and got rid of all of the changes except for the resolution o #240. It should be fine to merge now with approval. 

I've also created a discussion post demonstrating the results of my testing for different solutions to the price controller's current knowledge problem. Once you've implemented in-kind accounting for raw materials, we should be able to get around the unrealistic omnipotence problem we currently have quite easily, so I'll wait for that to be done.

lennyyyyyyyy (approved):
Looks good. Option 4 on your recent discussion definitely seems like the way to go. In-kind accounting is just necessary with changing prices. Also, per-plan inventories get rid of the "simplification" we had of consuming all inputs at the start of a plan. Now plans can just move the inputs they need to their own inventories and consume them over time freely.

Amittai-Aviram:
```
```
This statement does not appear in current `main`, but I was able to trace it to a change Devin made on 2026-05-07, in which he attempted to substitute the above statement within the iteration over the plan history for the original iteration over inputs needed for the product to get the respective current prices.

To see a full history of the file from its origins, you can enter the following command, after making sure that your local copy of `main` is up to date:
```
git log --follow -p main -- src/PriceController.cpp
```
Sal first wrote the file, but its contents are hard to recognize now. I introduced `update_price` around 2026-01-17, but it assumed that raw materials and machinery costs would remain fixed:
```
 product->price_per_unit =
+        product->price_per_unit - old_average_labor + new_average_labor;
```
I corrected this in a commit of 2026-02-13:
```
    for (std::pair<Product *, double> input : product->inputs_per_unit) {
        price += input.first->price_per_unit * input.second;
    }
        price += (machine->price_per_unit / machine->lifetime) * average_labor_per_unit;
    }
```
The machines part was incorrect, so I committed an attempted correction on 2026-02-14:
```
double machine_use_hours = hours / workers;
double machine_hours_per_unit = machine_use_hours / units;
    double machine_cost_per_hour =
        machine->price_per_unit / machine->lifetime;
    price += machine_cost_per_hour * machine_hours_per_unit;
}
```
After our discussion, i implemented an average over a per-product _plan history_ (rather than a weighted average estimate) in a commit from 2026-02-15, which you reviewed, IIRC.

So then come Devin's changes in May.

In June, I noticed the issue and brought it up at our meeting. In your 2026-06-05 commit, you restored the original treatment of raw materials. (Devin's changes had not affected machinery costs, as far as I could tell.)

PR #249, opened 2026-06-25
210 move plan and order definitions to separate files

### LINKED ISSUE #210: Move `Plan` and `Order` definitions to separate files

Amittai-Aviram:
Resolves #210 .

In addition to moving the `Order` and `Plan` type definitions to separate files, this change set revises the field names in `Plan` to reflect their meanings and also replaces the logic of decrementing `labor_time_remaining` with incrementing `labor_time_used`, which the Price Controller uses (without the need for subtraction).


Amittai-Aviram:
Screenshot 2026-06-25 at 12 14 54 PM

lennyyyyyyyy (approved):
LGTM

PR #252, opened 2026-06-27
239 determine the number of inputs for a product without reference to the overall number of products

### LINKED ISSUE #239: Determine the number of inputs for a product without reference to the overall number of products
Right now, the upper bound of the range of possible numbers of inputs for a given product depends on the overall number of products:
```
const int PRODUCT_NUM_INPUTS_MAX = STARTING_NUM_GOODS / 2;
```
This causes the supply chains to become more congested as the number of products increases, and otherwise has no strong rationale. Instead, the upper bound should be fixed, and perhaps subject to a user flag to override the default value.

Machines should be subject to the same consideration: the number of machines in the economy should not be set as a proportion of the overall number of products.

Amittai-Aviram:
Resolves #239 .

- Set the starting number of machines as a simple constant, without reference to the number of goods.
- Make the starting number of machines configurable from the command line.
- Set the maximum number of inputs per product as a simple constant, without reference to the number of goods.
- Make the maximum number of inputs per product configurable from the command line.
- Make the minimum and maximum _amounts_ of each input per unit of output product fixed constants.
- Rename the latter from `PRODUCT_INPUT_VARIANCE` (which is misleading) to `PRODUCT_INPUT_AMOUNT_MAX`).

@alexbcreiner0 —
NOTE: This change will require a (small) change in the LTE Data Aggregation Layer project, since the `-m` flag now determines the (starting) number of machines rather than the ratio of goods to machines. Thus, if Overseer uses initial parameters `-g 5 -m 5` for 5 initial goods and 5 goods for each machine, it should henceforth use `-g 5 -m 1`.


alexbcreiner0 (approved):
Both changes very welcome. LGTM. Overseer has been updated on my end. As soon as this is merged in I'll push those changes.

Amittai-Aviram:
Thank you @alexbcreiner0 !

PR #253, opened 2026-06-27
242 log random generation seed

### LINKED ISSUE #242: Log random generation seed

Amittai-Aviram:
Resolves #242 

- Save the seed, either generated by the random device or passed in from the command line, as a private field of `Sim`.
- Log the seed at program start under label `random_seed` and key `value`.
- Replace every call to `get_random_device` with a call to `get_random_generator` and remove the former from `Sim`.

On the last point — it is against best practice to get random values directly from `std::random_device` and to call it repeatedly, because it is relatively quite slow and — precisely — because it makes deterministic behavior impossible, even when desired. This change should make every pseudorandom generation event reproducible if the user enters the `-e` flag with an integer value.

### DIFF
--- a/header/Logger.h
+++ b/header/Logger.h
@@ -26,6 +26,7 @@ class Logger {
             PERSON,
             PRODUCER,
             PRODUCT,
+            SIMULATION,
             SOCIETY,
             ERROR
         };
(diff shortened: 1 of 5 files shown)


alexbcreiner0 (approved):
LGTM. I believe the plumbing is in place for Overseer to display whatever the seed is in the control panel if it gets logged. I'll probably need to fiddle with it a bit to fully get it working once this is pushed.

PR #255, opened 2026-06-28
Remove SQLite3 library

Amittai-Aviram:
Resolves #254 .


Amittai-Aviram:
@alexbcreiner0 @lennyyyyyyyy @nathanjlim  — I am taking the liberty to remove the SQLite3 library from our codebase, since it is no longer appropriate. Since this is a minimal change of simply removing the two files and the corresponding commands in the makefile, I am merging it into `main` without review.

PR #257, opened 2026-06-29
Implement the replacement cycle of machines

### LINKED ISSUE #203: Implement the replacement cycle of machines

Amittai-Aviram:
Resolves #203 .

# Essential Features
- We track input (raw materials) and machinery expenditures through a per-plan inventory, filled by transfer from the firm's general inventory in `Firm::start_plan`, and decremented appropriately in `Firm::move_plan_forward_one_step`.
  - `Plan::outlays` is a deep (i.e., independent) copy of `Plan::inventory` in `Firm::start_plan` and is not touched again until `Firm::end_plan`.
  - `Plan::needed_this_step` is a map structured the same way as `Plan::inventory`, but re-used on every time step in `Firm::move_plan_forward_one_step` to hold how much of each input and machine capacity must be used on the current step of plan execution. We use this data structure to terminate the plan "early" if we do  not have enough left of any necessary input or machine capacity to execute the next step.
-  In `Firm::end_plan`, we deduct from `Plan::outlays` any amount left of any input or machine capacity. (This could be a negative number if we used more than we had originally estimated.)
- The Price Controller uses the `Plan::outlays` field of each plan to aggregate the uses of all inputs and machines over all plans in its (moving) plan history window.  The total amount used for these indirect costs, divided by the total number of units, is then added to the price.

# Excluded from This Change Set
This change set does not attempt to account for a producer's machine capacity on hand to limit its estimated maximum quantity in a draft plan (in `Producer::draft_plan_and_return_order`). This computation would be tricky, involving a potential circular dependency: how much of fixed capital we use up in a plan depends on how long the plan runs, which depends on how many workers are working on it at the same time — but that number depends on the proposed maximum quantity to be produced.  If we decide that this limitation of the maximum is necessary, we can address it in a future issue.


src/Firm.cpp
```diff
@@ -181,26 +186,54 @@ void Firm::start_plan(Plan * plan) {
...
 
 void Firm::move_plan_forward_one_step(Plan * plan) {
-    double ideal_quantity_produced = calculate_quantity_produced_from_worker_suitability(plan);
-    double quantity_produced = std::min(ideal_quantity_produced, plan->quantity_remaining);
+    double expected_quantity_produced =
+        calculate_quantity_produced_from_worker_suitability(plan);
+    double quantity_produced =
+        std::min(expected_quantity_produced, plan->quantity_remaining);
     if (quantity_produced <= 0.0) {
+        end_plan(plan);
```
alexbcreiner0, line 223: I've always been confused as to why this wasn't here, but presumably it had to have been triggering somewhere else, right? I suppose its fine to be double checking for this, as once a plan is ended it wont be encountered again for checking, but I'm just curious.

src/Firm.cpp
```diff
@@ -181,26 +186,54 @@ void Firm::start_plan(Plan * plan) {
...
     }
-
-    double labor_hours_per_worker = quantity_produced / ideal_quantity_produced;
-	for (Person * worker : plan->workers) {
-		worker->register_hours_worked(labor_hours_per_worker);
-	}
+    Product * product = plan->order->product;
+    for (std::pair<Good *, double> input : product->inputs_per_unit) {
+        plan->needed_this_step[input.first] = input.second * quantity_produced;
+    }
```
alexbcreiner0, line 229: This is great, I was about to add something like this myself. Doing this opens up the possibility of allowing firms to accept and draft plans based on their pending inventories rather than their actual inventories, allowing for a faster paced 'just-in-time' style economy (since now we can just halt the production step here if the firm doesn't actually have the goods needed). Obviously more to do in order to allow for that but this is the crucial thing which was missing.

src/Firm.cpp
Amittai-Aviram: Previously, the routine `Firm::move_plans_forward_one_step`, in iterating over current plans, would also check whether `plan->quantity_remaining <= 0`, and would call `Firm::end_plan` on that plan. I am pretty sure that this was Sal's code, and it seemed reasonable at first to have that "driver" function do the triage over all the functions. But I think the new way is a bit more straightforward, especially since now we have two distinct conditions that can signal the end of plan execution.

src/Firm.cpp
```diff
@@ -181,26 +186,54 @@ void Firm::start_plan(Plan * plan) {
...
+    for (Machine * machine : product->machines_needed) {
+        plan->needed_this_step[machine] = 1.0 / machine->lifetime;
+    }
+    for (std::pair<Product *, double> input : product->inputs_per_unit) {
+        if (plan->inventory[input.first] <
+                plan->needed_this_step[input.first]) {
+            end_plan(plan);
+            return;
+        }
```
alexbcreiner0, line 238: On the note of my above comment, I don't think that we should be just ending the plan if the firm ever finds its inventory lacking. That seems a little overly aggressive, even without my proposed pending inventory changes. I think that we should just be returning here, so that the firm can try again to progress the plan at the next step.

alexbcreiner0: Later on, we can implement a real decision making progress for when plans are 'dropped' due to extreme delays or at the customer's decision to cancel and try somewhere else. But I think just returning should work fine for now.

Amittai-Aviram: At least for now, I don't think we're ready to go that route. The purchaser will order the remainder of the goods right away, which has the effect of balancing the load over the system.  Waiting for pending inputs to come in might be an option, but it has its down sides and would need to be discussed. I think such a design decision or option lies beyond the scope of this particular change set.

src/Producer.cpp
```diff
@@ -36,7 +37,7 @@ void Producer::on_time_step() {
 void Producer::add_to_catalog(Product * product) {
     catalog.insert(product);
     for (Machine * machine : product->machines_needed) {
-        machines.insert(machine);
+        demands[machine] = 1.0;
```
alexbcreiner0: This is not a good idea. By setting demand equal to 1, the initial inventory gets set to the `FIRM_STOCKPILE_DURATION`, initializing the machine supply arbitrarily into the thousands. These numbers are so high that none of your replacement changes ever really matter. Moreover there isn't any reason why the initial machine supply/initial projected machine demand can't mirror what is already happening with goods, whose initial numbers are calculated quite reasonably from the Leontief inverse of the requirements matrix. In order to actually test your other changes in Overseer, I made some changes to this effect, so I'll post that here:

```
void Producer::add_to_catalog(Product * product) {
    catalog.insert(product);
    double output_demand = Society::get_instance()->get_initial_production()[product];

    double demand_scale =
        output_demand *
        Sim::get_num_people() *
        Sim::get_num_goods() /
        Sim::get_num_producers();

    unsigned int starting_num_firms =
        Sim::get_num_producers() + Sim::get_num_distributors();

    unsigned int average_team_size =
        std::max<unsigned int>(
            Sim::get_num_people() / starting_num_firms,
            1
        );

    double machine_use_per_unit =
        product->living_labor_per_unit /
        average_team_size;

        double machine_demand =
            machine_use_per_unit *
            demand_scale /
            machine->lifetime;

        demands[machine] += machine_demand;
    }

    for (std::pair<Good * const, double>& input :
            product->inputs_per_unit) {
        double input_demand = input.second * demand_scale;
        demands[input.first] += input_demand;
    }
    static std::normal_distribution<double> demand_mult(1.0, DEMAND_PREDICTION_VARIANCE);
    for (std::pair<Product * const, double>& demand : demands) {
        demand.second *= demand_mult(Sim::get_random_generator());
        input_inventory[demand.first] = demand.second * FIRM_STOCKPILE_DURATION;
    }
    for (std::pair<Product * const, double>& stockpile : input_inventory) {
        log_inventory_level(stockpile.first, stockpile.second);
    }
    log_catalog_addition(product);
}
```

You can see a bit of complication in getting an expected team size, but other than that its the same as it is for regular goods. If you aren't comfortable with fractional machine inputs, you can round up the results to an integer, but judging by the rest of what is here I don't think you're prioritizing that kind of realism.

src/Firm.cpp
```diff
@@ -181,26 +186,54 @@ void Firm::start_plan(Plan * plan) {
...
         double required_input = input.second * plan->order->quantity;
         remove_input_from_inventory(input.first, required_input);
-	}
-    pooled_input_value += plan->raw_materials_budget;
+        plan->inventory[input.first] = required_input;
+    }
+    double expected_plan_duration = plan->labor_budget / plan->workers.size();
+    for (Machine * machine : product->machines_needed) {
+        double required_use = expected_plan_duration / machine->lifetime;
+        remove_input_from_inventory(machine, required_use);
```
alexbcreiner0: This is kind of a strange thing to call this, since it is not really reserving the machine for use on a particular plan but rather deducting a fractional part of its lifetime. Granted, I am more okay with it as it is now than my expected semantics (reservation of the machine) because that would surely grind production to a halt. I guess its fine for now?

alexbcreiner0 (changes requested):
Two overall overall major issues here which sort of cancel each other out to allow the system to function when it really shouldn't. Most of these issues appear to be due to existing double standards and bugs which hadn't caused issues before but only because we were lucky. 

First, all reorder attempts for machines are currently failing. The most fundamental reason for this is that the plan's order size field is still defined as an integer, which is incompatible with your fractional machine accounting. Since I still haven't implemented changed which would make machines more costly, given your fractional machine accounting here, the quantities demanded for plans are always being rounded down to zero. As a result, the plan itself gets dropped without even being logged as a draft plan because the number of workers assigned is zero. The simplest band-aid fix for this, which I used to confirm the problem and 'solve it' was to modify the `check_and_reorder` function so that machine orders had their quantities rounded up:

```
void Firm::check_and_reorder_input(Product * product) {
    double threshold = get_reorder_threshold(product);
    log_demand(product, threshold);
    float pending_inventory = get_pending_inventory_level(product);
    log_pending_inventory(product, pending_inventory);
    if (pending_inventory >= threshold || !threshold) {
        return;
    }

    double raw_order_quantity = threshold * FIRM_REORDER_MAX_PROP;

    int order_quantity;
    if (product->product_type == Product::kTypeMachine) {
        order_quantity = static_cast<int>(std::ceil(raw_order_quantity));
    } else {
        order_quantity = static_cast<int>(raw_order_quantity);
    }
    Order * order = new Order(
            product,
            order_quantity,
            this,
            FIRM_STOCKPILE_DURATION * pending_inventory *
            FIRM_REORDER_MAX_PROP / threshold
            );
    if (!send_order(order)) {
        log_reorder_failure(product, order->quantity);
    }
}
```

Second, despite all reorders for machines failing, this never causes any supply chain collapses, because other bugs ensure that machine supply is effectively irrelevant to production right now. This is because of two missing changes, one of which isn't entirely your fault:

- On the one hand, `get_max_order_quantity` isn't actually adjusting according to machine availability. 
- On the other hand, `remove_input_from_inventory` currently has a bug where it returns false if the amount to remove isn't available in the firm's inventory, which gets ignored by the `start_plan` function. The plan's inventory gets the required fraction of the machine deposited regardless of the firm's actual inventory. The reason that this hasn't been as big of a problem for regular goods is because, when drafting plans, firm's automatically adjust the requested order sizes to match what is realistic given their inventories, with a call to the `get_max_order_quantity` method. However, as per the first bullet point, this PR didn't add any adjustments to this function for machines. 
 
The bottom line with all of this is that your changes here expose some existing problems with our code. However, the fixes I've suggested should allow your PR to go through without causing problems for the time being.

src/Producer.cpp
Amittai-Aviram: I have more or less followed your code, except that I have used rational numbers consistently except where an integer value is required.

Amittai-Aviram:
> Two overall overall major issues here [...]

I'm afraid I did not see this comment until after a lot of debugging. I made a version of your first change independently. (It's not strictly necessary to perform a static cast to int — that happens by virtue of the assignment.)

On the second point, first, I have rewritten `get_max_order_quantity` to deal in rational numbers, and have `draft_plan_and_return_order` round that value up in the case of an order for a machine. In addition, `Firm::start_plan` now _returns_ if `Firm::remove_input_from_inventory` returns `false`, thus, in effect, stalling or delaying until the reorder shipment for that input (or machine) has come in.

I have verified by instrumenting the code that, in fact, machines are now produced when I run the simulation with the same flags as are passed by default from Overseer. Nevertheless, Overseer is still showing a machines supply of 0, so this makes me suspect that perhaps there might be a bug on the Overseer side.  My instrumentation in this case is to print the product ID out at the beginning of `Firm::receive_shipment`.

Amittai-Aviram:
On a point to which you referred in passing — `Producer::get_max_order_quantity` does not check the current machine capacity. At least for now, I have not changed this — the number is only determined by the input goods. That is because the expected machine usage depends on the expected duration of plan execution, which depends on the number of workers assigned to the plan — which depends on the quantity determined for the plan, which depends on the estimated maximum quantity, At least for now, my reasoning is that we simply cannot bound the offered quantity to produce in advance, because of this circular dependency of information, so we have to promise to produce at least one machine and then delay plan execution (by `return`ing from `Firm::start_plan`) until the requisite machine capacity has come in. This assumes that there was, at least at the founding of the firm, enough machine capacity for the first plan, and, therefore, that `Firm::check_and_reorder_input` had gotten to the point of reordering the machine. My instrumented code output suggests that this is working, at least in the sense that machines are being reordered and produced.  The inventories of goods in Overseer look different from how they used to look, but I'm not sure whether that is bad. Basically, they seem to level off to stable values after an initial period of turbulence.

Amittai-Aviram:
Update: after a while, consumer goods seem to run out without being restocked. Producer supplies remain stable. I shall investigate the former further.

PR #258, opened 2026-06-29
Have working hours float based on sampled busyness distributions, fix transfer bug

### LINKED ISSUE #223: Allow the length of the working week to float towards a minimal level (Outline of a plan provided)
First, let's be clear that the number we have enough information to shrink is not $T$, the number of hours in the working day, but $TW$, the number of hours in the working week. We should calculate how this should shrink, and then distribute that reduction equally across all days of the working week. We can implement a system later for deciding when to reduce $W$ as well, but this is a subjective choice of the population and not very interesting to us theoretically speaking. 

# A Reasonable Idea
Routinely we are having the `Society` instance calculating the total average busyness of a worker $b$. Busyness is defined as the number of hours worked weekly on plans divided by the number of hours in a week:

$$ **b = \frac{K}{\Gamma}** $$

where $K$ is the number of hours worked and $\Gamma = 24*7$. Multiplying both sides by $\Gamma$ gives 

$$ K = b\Gamma $$

we also know that busyness *should* tend to settle and hover around the *minimal* busyness necessary for social reproduction:

$$ b \approx \epsilon \omega $$

(see my [discussion post](https://github.com/BC-LTEWG/Labor-Time-Economy-Simulation/discussions/193) for more elaboration on this)

What we want then is to adjust the working week's length such that the probability that the required labor hours for that week exceeds this number is statistically insignificant. We can do this by constructing a **confidence interval** around the overall busyness, and allowing the working day's length to float towards the *upper bound* of this interval. 

Let $\bar{K}$ denote the true minimal number of hours needed weekly to produce the net consumption bundle, and $B$ be the random variable denote the busyness of a random worker. The society object reports the population mean busyness $\bar{b}$ whenever the method is called. By the central limit theorem, this distribution of means is normally distributed around the true mean $E(B)$, with a shrinking standard deviation, and which itself represents the minimal busyness which reproduces the net product: 

$$ \Gamma E(B) = \bar{K}$$

Let $t$ be the desired degree of confidence that the true value $\bar{K}$ falls within the interval we are constructing, and $m$ be the number of times that the average busyness has been sampled. Let $s_{\bar{b}}$ be the sample standard deviation of reported average busynesses. Using $m$ and $c$, we can look up the associated $t$-value in a [t-table](https://byjus.com/maths/t-test-table/), to construct our interval:

$$ \left[ \bar{b} - t_c \frac{s_{\bar{b}}}{\sqrt{m}}, \bar{b} + t_c \frac{s_{\bar{b}}}{\sqrt{m}} \right] $$

We have a probability of $c$ that the true value of $E(B)$ must fall inside of this interval. With this level of confidence then, we can allow the length of the working day to float towards the upper bound of this interval:

$$ TW \to \bar{b} + t_c \frac{s_{\bar{b}}}{\sqrt{m}} $$

This effectively minimizes the length of the work week while guaranteeing with a desired level of confidence that the week will always be long enough to accommodate the necessary levels of production. The adjustments can happen on a daily, weekly or monthly basis using procedures which we've employed for similar tasks (e.g. something like the [exponential smoothing](https://en.wikipedia.org/wiki/Exponential_smoothing) procedure that Devin uses to approximate busyness in the first place). 

# Details to Work Out
**However**, this is all assuming that $\Gamma\bar{b}$ actually acts as an effective approximation of $\bar{K}$. Does it? The answer is: **eventually.** However, initially, workers are unemployed, and thus not busy at all. In my testing, even when society stabilizes, it can sometimes take up to a year for the overall busyness values being reported to actually climb up to the minimum and stay there consistently:

Image

Here, the product input epsilon constant $\epsilon = 0.5$, so the minimal busyness level is $\epsilon \frac{TW}{\Gamma} \approx 0.12$. After a full year of simulation, the distribution of busyness values looks like this:

Image

$0.12$ here is still nothing short of an outlier at this point. It will take at least another two years for the distribution to correct itself and actually centralize around the appropriate value. Thus our updating procedure requires some guards, or some damping of past values (e.g. more exponential smoothing), or both. For what it is worth, the distribution here of $b$ is easy to fit to a curve. It is clearly a [beta distributed](https://en.wikipedia.org/wiki/Beta_distribution) random variable. It is standard to go with a beta distribution in scenarios where the variable appears normal, but has a bounded length support, which is the case here: $b$ is bounded in the interval $[0,\omega]$. I don't know how helpful this information will be, but it is worth noting.

I fear that we will need to implement some guards to make sure that the working day doesn't change if things aren't stable, which is a can of worms because 'sufficiently stable' is a difficult thing to pin down rigorously. Additionally, the confidence interval I described is probably not good enough. What we will want in actuality is something something more like: "what is our confidence that the busyness never leaves this interval for an entire year or more?" or something like that.

This will require some thought from everyone, some discussion, and some trial and error. But we are _very_ close to being able to do this, and I wanted people to see how close we were to it.

lennyyyyyyyy:
Resolve #256 
Resolve #223 
Resolve #261 
Notes:
 - Working hours are updated every `WORK_HOURS_UPDATE_PERIOD`. There is no new constant for when busyness distributions are sampled. They are simply sampled every `BUSYNESS_AVERAGING_WINDOW`, with the logic being that the busyness of the individual will have taken on a nearly independent random value by then.
 - The busyness data isn't really accurately described by a normal distributed variable, which is then maxed at some value, which is what the [Tobit model](https://en.wikipedia.org/wiki/Tobit_model) describes. Essentially, there isn't a discontinuous spike at the maximum busyness. Instead, you would just see bunching at the maximum that deviates from normality. In the Tobit model, you consider all datapoints exactly equal to the maximum to be affected by censoring; in our model, `EFFECTIVELY_CENSORED_BUSYNESS_BUFFER` is used to create a range that we consider to be significantly affected by censoring.
 - The gradient descent algorithm used to predict the uncensored distribution uses learning rate and tolerance constants that I decided weren't important enough to put in Constants.h
 - Added the new logs `log_busyness_data`, `log_censored_busyness_distribution`, `log_predicted_uncensored_busyness_distribution`, and `log_work_hours_weekly`


alexbcreiner0:
This is exciting, but will take me some time to look through carefully. I also might try to push some of my my own stability changes before bringing this in (think there are still some bugs with my consumer goods demand that I've been working on). Just a warning that there might be some delay on my review, but I very much do want to be the one to review it carefully.

alexbcreiner0 (changes requested):
Devin, this is incredible work. It goes above and beyond anything I would expect of an undergrad, and is extremely impressive. Your method of estimating the mean and variance of the counterfactual distribution which _would_ prevail if there were no upper bound on the length of the working week is an extremely elegant way to approach the problems we are dealing with. However, some testing in Overseer reveals that it needs some work still. 

In particular, while I think the mathematical details of your estimation approach are valid and very elegant, it all hinges on one critical assumption: that the overall busyness of society will actually 'bunch' around the minimum length that we are 'censoring' with our maximum working day's length. Unfortunately, it seems more complicated than that: 

Image

This was initially an extremely healthy economy, where busyness values very quickly snapped to the expected level. I've set the update parameter to 1 month instead of 3 to more easily witness the issue. In particular, pay close attention to what happens after the second update. The max working day length seems to go to where we would want it, but then the actual busyness levels of society _seem to repulse from it_. This indicates to me that the busyness levels do not actually tend to bunch around the maximum working day length that we set, but actually keep some minimal distance from it. I'm not sure why this is immediately. Transfer inefficiencies come to mind, but it could be a very complex concoction of a variety of necessary inefficiencies, some of which might be due to issues in our code base but others of which might be inevitable. Regardless, because of this distance, the updating algorithm continues to drop below the minimum, and our society eventually collapses.

For your algorithm to work here, we need to figure out where our busyness levels _actually_ tend to bunch. It might be the case that we need some secondary estimation procedure for doing this. Let's definitely talk about this at the next meeting. If you can't make it, I am perfectly fine with presenting the mathematical details of your estimation procedure to the group myself, and already have my own write up for it that I could post as a discussion, with your consent.  

Parameters which reproduce:
```yaml
    params:
      N_c: 5
      N_S: 10000
      N_h: 100
      N_inputs_max: 5
      N_m: 1
      N_p: 5
      N_d: 5
      W: 5
      T: 8
      S: 0.1
      v_ability: 0.1
      N_a: 5
      fixed_seed: false
      seed: 2728540000
      is_logging: false
      productivity: 0.5
      consump_epsilon: 0.6
      init_prices: labor_values
```

I'll push the changes to the aggregation layer which will allow you to see the maximal busyness falling right after submitting this review. 

Two more lesser issues with your approach:
1. Since you are using the busyness data for individuals in the population rather than average busyness levels as your datapoints, you can no longer assume that the busyness distribution is normal. That was a luxury of the central limit theorem, which we get from working with averages. You said yourself that busyness levels would _not_ be normal in our situation, and I agree with you, especially with our 'censoring' via a cap to the length of the working week. However, the confidence interval we are building still assumes that these values are normally distributed. This isn't really an issue necessarily, but it's worth noting that we are now asserting something that we weren't before: that the _uncensored_ busyness levels are normally distributed.
2. You are taking the busyness levels of every single person as data. This is fine, but in the real world would not be feasible. Instead, some kind of census would be needed, and only a sample of the busyness data would be obtainable. You would at that point need to adjust your initial estimates of $\mu$ and $\sigma$ to be the _sample_ mean and the _sample_ standard deviation. The only thing this would change is that you would divide by `data.size()-1` to get the variance instead of `data.size()` as you are now. If we want to implement more of a 'proof of concept' algorithm, we might consider only taking a sample of the busyness data and doing this instead, to make it more applicable in the real world. I'm fine with how it is right now though.

Additionally, please remove your changes to the labor transfer system from this PR. I've already created a separate PR which includes those as well as some other related fixes. If you want to simply review and approve those, I can merge them into main, and then you can merge them back into your branch.

lennyyyyyyyy:
The repelling of the societal average busyness from the maximum is pretty expected - for it to bunch a the maximum for a given timestep, every individual would have to be maximally busy for society to be maximally busy. I think that's just probability at work. What I actually said would have clear bunching / censoring were the data points from individuals. That's the assumption that the prediction makes, not that the average across all individuals would have significant bunching, which I think you slightly misunderstood. There's actually a log that would let you check this: `log_busyness_data`, which gives the series index, and each datapoint of the series. Looking at this graph (which I don't think you have yet) would be pretty useful, to see if that assumption actually holds up, and to also answer the question of individual busynesses being normally distributed or not. 

Also, a month for the work hour update might be a little fast. That's around 3-4 samples, if `BUSYNESS_AVERAGING_WINDOW` is at a week. The distribution that the prediction algorithm analyzes might differ from normality then, but I don't know for sure. Again, looking at the log would help.

alexbcreiner0:
Those are fair points. It wasn't a misunderstanding, but rather an oversight in my assessment. Regardless of whether my assessment of the why is correct or not though, it is the case that the working day length seems to be falling below the theoretical minimum (even with updates every 3 months), and I've never seen it pick back up again. This could be due to other stability issues however. 

I'll get that graph going and go back to the drawing board in assessing how well this is working.

PR #259, opened 2026-06-30
Implement Doctest Framework & Coverage

### LINKED ISSUE #241: Add unit tests to cover functions throughout the code base
We currently have a `test` subdirectory containing two outdated Doctest unit testing files and one shell script to run all tests in the directory.
- Update or replace the current tests.
- Add thorough unit tests for every class and item in the code base.

nathanjlim:
Resolves #241, resolves #269

**Doctest Implementation:**
- test_Person.cpp
- test_Producer.cpp
- test_Product.cpp
- test_Society.cpp
- test_ConsumerGood.cpp
- test_Distributor.cpp
- test_Good.cpp
- test_Machine.cpp
(Run via: "make tests", "cd test", "./run_tests.sh")

**Coverage:** Currently managed via the Makefile by integrating lcov ("make score").
- Temporary setup at the moment

**Automation Check:** Includes commits related to the automation check (which is technically a separate issue).
- Currently broken, but does not affect the simulation or doctests' functionality.


.github/pr-tests.yml
nathanjlim, line 1: Includes commits related to the automation check (can be ignored)

test/test_ConsumerGood.cpp
nathanjlim, line 1: Product type is assigned: Confirms constructor assignment is correct by checking if product_type explicitly equals TYPE_CONSUMER_GOOD.

Pointers are mapped: Proves raw good and consumer good are correctly linked in memory by verifying the base good's pointer loops exactly back to test_cg.

Consumption frequency works: Ensures randomized demand metric initialized by stating mean_consumption_frequency is strictly greater than 0.

test/test_Distributor.cpp
nathanjlim, line 1: get_client_type() identifies distributor: Confirms object accurately reports its type as Logger::DISTRIBUTOR.

add_to_catalog() calculates inventory: Verifies initial stocking by stating the inventory level is strictly greater than 0 for both the consumer item and underlying base good.

try_sell_goods() deducts inventory on purchase: Simulates live transaction by temporarily forcing the item into the public sector (bypassing monetary edge cases). Sale succeeds and internal inventory decrements by exactly 1.

test/test_Good.cpp
nathanjlim, line 1: Product type is assigned: Confirms constructor sets classification correctly as product_type explicitly equals TYPE_GOOD.

test/test_Machine.cpp
nathanjlim, line 1: Product type is assigned: Confirms constructor sets classification correctly as product_type explicitly equals TYPE_MACHINE.

Machine lifetime within bounds: Verifies randomized durability metric by stating machine's lifetime falls strictly between predefined MACHINE_LIFETIME_MIN and MACHINE_LIFETIME_MAX constants.

test/test_Person.cpp
nathanjlim, line 1: charge() deducts money: Validates account transactions by successful deduction for valid funds and rejection for overdraft attempts.

productivity() scales: Base productivity metric equals 1.0 when healthy and correctly scales down to the UNHEALTHY_PRODUCTIVITY constant when sick.

set_firm() updates pointer: Confirms employment assignment by checking initial unemployed (null) state updates to new Firm.

train() updates ability scores: Simulates ability training target and internal ability dictionary correctly updates.

update_busyness() applies average: Historical busyness metric recalculates and current time step counter resets to 0.0.

suitability() averages: Correct mathematical average of given skill set is properly scaled by the agent's health status (vice versa).

consume() reduces inventory: Simulates agent using product and internal inventory count for specific ConsumerGood decrements by 1.

shop() returns budget: Simulates with an empty bank account (0.0) and inventory remains valid and unaffected (preventing unauthorized or negative purchases).

test/test_Producer.cpp
nathanjlim, line 1: get_client_type() identifies producer: Object reports type as Logger::PRODUCER.

add_to_catalog() calculates demand and inventory: Producer successfully stocks physical machinery and raw input goods (>0) required for products.

can_produce() checks producer: Firm correctly returns true when evaluating capability to manufacture a product already established.

get_max_order_quantity() calculates limits correctly: Dividing current raw inventory by required inputs, and firm's internal method calculates this maximum limit.

test/test_Product.cpp
nathanjlim, line 1: Constructor has labor: living_labor_per_unit and size of required_abilities fall within predefined min/max constant bounds.

set_inputs() values equal recipes: Product possesses valid manufacturing recipe as inputs_per_unit meets minimum size threshold.

set_machines() assigns tools: Product requires necessary physical tools for creation as machines_needed array meets minimum size threshold.

test/test_Society.cpp
nathanjlim, line 1: set_abilities distributes and randomly sizes: Ability pools match configured simulation size limits.

set_initial_account calculates correct baseline money: Agent starting funds match expected baseline.

get_instance has singleton pattern: Object has singleton identity via identical memory addresses.

get_id() returns 0: Singleton ID is strictly 0 and throws invalid_argument if altered.

on_time_step() gets average: Updates society-wide average busyness and account balances.

get_total_employment returns percentage: Employment rate between 0.0 and 1.0.

public_fund adds/subtracts: Financial transactions correctly modify public fund balance.

set_initial_products generates properly: Product list size equals the sum of goods, consumer goods, and machines.

birth_person increments population: Global population and unemployed tracking vectors increment by 1.

populate_io_matrix_and_labor_vector structures: Populated Input-Output matrix maps positive labor vectors.

get_max_eigenvalue gets largest value: Identifies highest eigenvalue in test matrix.

get_leontief_inverse computes (I - A)^-1: Utility correctly computes Leontief Inverse matrix.

makefile
nathanjlim, line 1: clean: Expanded to remove LCOV/GCOV.

coverage: Configures flags and triggers clean test environment.

score: Automates test execution and includes LCOV to report for "src" directory.

test/test_ConsumerGood.cpp
Amittai-Aviram: For consumption frequencies — I think there's another invariant that we shall want to test, which is the theoretical basis for the assignment of consumption frequencies in the first place. This should be a separate issue. Please see Issue #144 for the mathematical basis of setting the consumption frequencies in such a way that people will produce at least as much as they consume. Then please write up an issue to write a unit test that ensures that we always maintain this invariant.

test/test_Society.cpp
Amittai-Aviram: Here, too, I think there's an invariant to be tested regarding the input-output matrix A after normalization (`adjust_io_matrix`), which is that the matrix is [productive](https://en.wikipedia.org/wiki/Productive_matrix). A better definition of a productive matrix that is also more relevant to our context is that it is a matrix **A** such that there exists a vector **x** with all strictly positive (i.e., > 0) entries such that $\mathbf{0} < \mathbf{A}\mathbf{x} < \mathbf{x}$ — that is, the vector you get after transformation is still all strictly positive, but all the entries are less than the original entries in **x**. You need that because the input-output matrix **A** tells you what you need to produce the required output products, but those products, in turn, are inputs to produce products, also specified in the matrix. So the total work (value) you need, $\mathbf{x}$, to produce your output products whose respective needed quantities are the entries of a vector $\mathbf{d}$ is the sum 
$\mathbf{x} = \mathbf{d} + \mathbf{A}\mathbf{d} + \mathbf{A}^2\mathbf{d} + \dots = \sum_{k = 0}^{\infty} \mathbf{A}^{k}\mathbf{d}$
That infinite sum needs to converge! So each step back in time has to require less work than the step that comes after it.

Amittai-Aviram: I am a little unsure about this, but I think you can take the vector of initial production (`Society::get_initial_production`), multiply that by **A**, and see whether the entries in the resulting vector are smaller than those in the initial production.

Amittai-Aviram (approved):
LGTM
Sorry it took me so long to approve!
This is a great start, on which we can now build further.

PR #265, opened 2026-07-04
Fix various worker transfer issues

alexbcreiner0:
Resolves #256 (and my additional problems with the system in the comments of that issue)

Shouldn't be hard to review and approve so I requested everyone, just the first person who has a sec because this is a borderline emergency hotfix.

### DIFF
--- a/src/Firm.cpp
+++ b/src/Firm.cpp
@@ -92,8 +92,13 @@ std::vector<Person *> Firm::propose_transfer(int workers_wanted) {
         log_busyness(firm_busyness, societal_busyness, 0);
         return {};
     }
-    int max_workers_to_transfer = (int) (workers.size() * (1.0 - firm_busyness / adjusted_societal_busyness)); 
-    max_workers_to_transfer = std::max(max_workers_to_transfer, workers_wanted);
+    int available_workers_for_transfer = std::max(
+        static_cast<int>(
+            workers.size() * (1.0 - firm_busyness / adjusted_societal_busyness)
+        ),
+        0
+    ); 
+    int max_workers_to_transfer = std::min(available_workers_for_transfer, workers_wanted);
     log_busyness(firm_busyness, societal_busyness, max_workers_to_transfer);
     std::vector<Person *> transfers;
     for (Person * worker : standby_workers) {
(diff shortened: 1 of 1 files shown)


src/Firm.cpp
```diff
@@ -92,8 +92,13 @@ std::vector<Person *> Firm::propose_transfer(int workers_wanted) {
         log_busyness(firm_busyness, societal_busyness, 0);
         return {};
     }
-    int max_workers_to_transfer = (int) (workers.size() * (1.0 - firm_busyness / adjusted_societal_busyness)); 
-    max_workers_to_transfer = std::max(max_workers_to_transfer, workers_wanted);
+    int available_workers_for_transfer = std::max(
+        static_cast<int>(
+            workers.size() * (1.0 - firm_busyness / adjusted_societal_busyness)
```
Amittai-Aviram, line 97: Something I noticed

src/Firm.cpp
```diff
@@ -280,17 +285,17 @@ void Firm::assign_workers(Plan * draft_plan) {
 
     int workers_left = predict_workers_needed(draft_plan);
     for (Person * worker : sorted_standby_workers) {
-        if (workers_left == 0) return;
+        if (workers_left <= 0) return;
```
Amittai-Aviram: Could `workers_left` ever be < 0 here? It comes in as the return value of `predict_workers_needed`, which is a product of a bunch of factors, none of which looks to me as if it could be negative. Aside from that initial value, the only way that this variable could hold a negative value, then, is if it had started out as 0 and then been decremented, but that's also impossible.

Style: please revise this to use an explicit block:
```
if (workers_left == 0) {
    return;
}
```
(— or `<= 0`, as the case may be).  Please do that for all such `if`-clauses.

src/Firm.cpp
```diff
@@ -280,17 +285,17 @@ void Firm::assign_workers(Plan * draft_plan) {
...
         draft_plan->workers.push_back(worker);
         workers_left--;
     }
     for (Person * unemployed_person : Society::get_instance()->get_unemployed_people()) {
-        if (workers_left == 0) return;
+        if (workers_left <= 0) return;
```
Amittai-Aviram: Again, I'm afraid I do not see how `workers_left` could be less than 0 here, unless it had started out below 0 as the return value of `predict_workers_needed`.  (Of course, it does not hurt — I just do not understand why these three changes are necessary.)

src/Firm.cpp
```diff
@@ -280,17 +285,17 @@ void Firm::assign_workers(Plan * draft_plan) {
...
         draft_plan->workers.push_back(unemployed_person);
         workers_left--;
     }
     for (Firm * firm : Society::get_instance()->get_firms()) {
-        if (workers_left == 0) return;
+        if (workers_left <= 0) return;
```
Amittai-Aviram: As above.

Amittai-Aviram (approved):
The changes to `Firm::propose_transfer` are good. I'm afraid I do not understand why the checks for `workers_needed <= 0` (as opposed to `== 0`) are needed. The value cannot emerge from the first loop < 0, because that implies that it was 0 coming into the loop, but the function would have returned at that point — etc. I'm sorry if I am missing something obvious. Anyway, please make the style changes as shown. Thanks!

src/Firm.cpp
alexbcreiner0: These changes were Devin's. I think he was concerned about the case where a requesting firm received more workers back than what was requested. I believe my other changes block this from happening, but I kept them anyway because it can't hurt.

PR #272, opened 2026-07-13
Various bugfixes and stability/safety tweaks

alexbcreiner0:
I know I said that I would make this multiple PRs but I think this should be manageable as a whole. 
Resolves #271 
Resolves #270

Note: With my changes to `check_and_reorder`, `get_reorder_threhold` is now unused within the simulation. I hesitated to remove it though because Nathan's has unit tests which use it, and I don't understand that code well enough to revise his test to not use it. So @nathanjlim if you could let me know what to do there I would appreciate it.


src/Distributor.cpp
```diff
@@ -32,6 +33,14 @@ Logger::Client Distributor::get_client_type() {
 
 void Distributor::on_time_step() {
     Firm::on_time_step();
+
+    for (Product * product : catalog) {
```
lennyyyyyyyy: Can't you use the same strategy as in Producer, where if the demand is touched, you call check_and_reorder? Then you can put the loop in Firm.

src/Firm.cpp
```diff
@@ -163,35 +182,73 @@ double Firm::get_pending_inventory_level(Product * product) {
...
+        }
 	}
+
+    for (Person * worker : plan->workers) {
+        move_worker_off_standby(worker);
+    }
+
+    for (std::pair<Good * const, double>& input : product->inputs_per_unit) {
+        check_and_reorder_input(input.first);
+    }
```
lennyyyyyyyy: I don't think this is necessary. Including this keeps the issue of multiple successful reorders for a product going out in the same time step, which causes overallocation of workers in that sector.

src/Firm.cpp
```diff
@@ -60,13 +60,30 @@ void Firm::receive_payment(Plan * plan, double transaction_amount) {
...
+    input_inventory[product] -= quantity;
+    deducted_inputs.push_back(
+        std::make_pair(product, quantity)
+    );
+    add_demand_signal(product, quantity);
+    log_inventory_reduction(product, quantity);
+    log_inventory_level(product, input_inventory[product]);
+    return true;
+}
+
```
lennyyyyyyyy, line 79: Can't you just do 
```
if (remove_input_from_inventory(product, quantity)) {
   deducted_inputs.push_back(
        std::make_pair(product, quantity)
    );
}
```

src/Firm.cpp
```diff
@@ -163,35 +182,73 @@ double Firm::get_pending_inventory_level(Product * product) {
...
+    if (product->product_type == Product::ProductType::TYPE_MACHINE) {
+        order_quantity = std::ceil(order_quantity);
+    }
     Order * order = new Order(
             product,
-            threshold * FIRM_REORDER_MAX_PROP,
+            order_quantity,
             this,
-            FIRM_STOCKPILE_DURATION * pending_inventory * FIRM_REORDER_MAX_PROP / threshold
+            order_quantity * product->price_per_unit / demand
```
lennyyyyyyyy: `order_quantity / demand` is the time it takes for the order quantity to be depleted, but I don't really get why you're multiplying by the price. If 10 units that cost 2 hours each are being depleted per hour, you want a reorder of 10 units to come in every hour, not 2 hours, because then you wouldn't replenish your inventory in equilibrium, right?

src/Producer.cpp
```diff
@@ -62,35 +65,88 @@ bool Producer::can_produce(Product * product) {
...
-            return_order_quantity,
+            0,
             order->customer,
-            std::max(1.0, order->requested_turnaround_time
-            * return_order_quantity / order->quantity)
-            );
-	Plan * draft_plan = draft_plan_for_order(return_order);
-    if (!draft_plan) {
+            1.0
+        );
```
lennyyyyyyyy: Not too sure why creating a new order is necessary when you're rejecting.

lennyyyyyyyy (changes requested):

src/Firm.cpp
alexbcreiner0: Price of the order / demand is the _time_ that it would take a single worker to complete the order, since price is approximately labor value. So the baseline is one plan = one worker, was my idea.

src/Firm.cpp
alexbcreiner0: If that is actually a problem, then we should add some more specific gating than just removing this, because a recursive cascading of checking and reordering on a single time step is absolutely necessary. Just looping over each thing once and reordering at most once for each product isn't good enough to process the whole supply chain. (Try it yourself! Disable this and try a higher difficulty of production with this commented out vs not, and tell me if this seems necessary!)

src/Firm.cpp
alexbcreiner0: Yeah good call, I'll do this.

src/Distributor.cpp
alexbcreiner0: I suppose. Will try.

alexbcreiner0: Actually I'm not sure if I understand what you mean here. The difference is that here we need to also loop over the underlying goods as well as the consumer goods. I think it's fine to have it like this tbh unless there's something I'm not understanding.

src/Firm.cpp
alexbcreiner0: Yeah, I think your argument is sound. I'll change.

src/Producer.cpp
alexbcreiner0: Well, the function is supposed to return an order, right?

src/Firm.cpp
alexbcreiner0: Okay you might actually be right about this. I'm confused by it though. Let's talk about it at the meeting tomorrow. I will remove for now.

src/Distributor.cpp
lennyyyyyyyy: I think they should both have non-zero demand. So the loop should catch all of them.

src/Producer.cpp
lennyyyyyyyy: Why isn't it the same logic as 94-95? You can just use the same Order object you already have. Same goes for 123-133.

lennyyyyyyyy:
Ok, so I did something that actually seemed to completely fix the over-allocation at the beginning of time (which is still happening in your current PR) and over-transfers of workers after the beginning. Just as I suspected, it has to do with the ability of a Firm to send out multiple successful reorders of the same product - causing multiple plans to run at the same time upstream. This doesn't seem wrong, but remember that each one predicts its production rate and its worker allocation separately to match demand, *as if the order is the only one supplying the firm*. If any firm has multiple reorders out, then the actual production rate and worker allocation is always going to be wildly over what it needs. 

I thought that the change to reorder on every time step would be enough, since this at least prevents multiple reorders going in a single time step, but this is actually still not enough and is quite arbitrary. It just places a limit that a successful reorder can be put out once every hour. Instead, I made a small change that doesn't allow another reorder for a product unless there are no outbound orders for that product. The result is an immediate jump to equilibrium employment values at time 0 and a really low variance from then on:
image
image

I think this proves without a doubt what the issue behind over-allocation is, but is it a full solution? A new issue that might arise is the effect of a changing demand. Multiple orders let firms respond to a sharp increase in demand, whereas restricting each firm to one causes a delay in the response. Basically, if we are to allow multiple concurrent reorders, we need to require that the production rate of each successive reorder takes into account the production rates of all other currently outbound orders. That's probably a change for a different PR though.

src/Distributor.cpp
```diff
@@ -80,23 +77,25 @@ void Distributor::check_and_reorder_input(Product * product) {
...
-            threshold * FIRM_REORDER_MAX_PROP
+            reorder_threshold * FIRM_REORDER_MAX_PROP
             );
     if (!distribution_quantity) {
         return;
     }
     double distribution_time = 
-        static_cast<double>(distribution_quantity)
-        / demands[consumer_good];
+        static_cast<double>(distribution_quantity) * consumer_good->price_per_unit
```
lennyyyyyyyy: Remove price_per_unit here.

lennyyyyyyyy (approved):
I guess we can talk about the over-allocation at the meeting. Just one small mistake left here.

PR #278, opened 2026-07-14
Optimize test dependencies and add runtests

nathanjlim:
Resolves #268

Replaced `*.cpp` with `*.o` dependencies in `TEST_DEPS` to properly link object files. 
* Added a dedicated `runtests` rule that automatically builds the tests if they don't exist, changes directories via `pushd`/`popd` and then returns to the root.


makefile
```diff
@@ -71,15 +71,22 @@ ${BUILD_DIR}/Logger.o : ${SRC_DIR}/Logger.cpp ${HDR_DIR}/Logger.h ${HDR_DIR}/Con
...
 
 tests: ${TEST_EXECS}
 
 test/%.test: test/%.cpp ${TEST_DEPS}
 	g++ ${FLAGS} $^ -o $@
 
-.PHONY: trace plot-tool graphs
+.PHONY: trace plot-tool graphs runtests clean
+
+SHELL := /bin/bash
```
Amittai-Aviram: Why do we need this? If we do not, please remove it. Different systems have different default shells, and, unless there's a good reason to do so, we should not try to override those.  (E.g., Macs use ZSH by default, which derives from BASH but has slightly different syntax for some commands.)

Amittai-Aviram (approved):
Great work, Nathan! Please check about the SHELL variable definition. Also, I'm afraid you'll need to update the main branch upon which this branch is based. Please always do this before merging into main in any event:
```
git switch main
git pull
git switch <your-branch>
git merge main
```
Otherwise, LGTM!

PR #282, opened 2026-07-17
Fix over-allocation of workers due to simultaneous outbound reorders not considering production rate

lennyyyyyyyy:
Resolve #280 
This set of changes does the following:
- For each new reorder, it considers all other reorders so that the summed production rate of those reorders is at least the rate at which the input is consumed
- Abolishes the pending inventory - since production rates of individual orders now vary, the pending inventory is no longer reliable. Now, reorders are considered based on the actual inventory, and are limited by the total production rate of all outbound orders.

Theoretically, an unexpected increase in demand will cause adjustment reorders that will match with an increase in production rate. At constant demand, the simulation shows similar behavior and results to [devins-branch-3](https://github.com/BC-LTEWG/Labor-Time-Economy-Simulation/tree/devins-branch-3), where I've implemented the constant production rate fix that doesn't allow more than one simultaneous reorder. The results are the same in both - much more reasonable employment and transfer variance.


src/Firm.cpp
```diff
@@ -170,24 +170,24 @@ double Firm::get_reorder_threshold(Product * product) {
...
+        needed_production_rate -= production_rate;
     }
-    return pending_inventory;
+    return needed_production_rate <= 0 ? 0 : needed_production_rate;
 }
 
 void Firm::check_and_reorder_input(Product * product) {
-    double demand = demands[product];
-    double reorder_threshold = demand * FIRM_STOCKPILE_DURATION;
+    double production_rate = get_needed_production_rate(product);
```
alexbcreiner0: I think this  should be renamed to something like `resupply_rate` since the firm isn't producing this themselves (also, this isn't the rate of production for society generally since there will be ongoing production for firms other than themselves)

header/Constants.h
```diff
@@ -50,7 +50,7 @@ const int FIRM_STOCKPILE_DURATION = 1.5 * WEEK;
 const double FIRM_REORDER_MAX_PROP = 0.25;
 const int DEMAND_AVERAGING_WINDOW = MONTH;
 const int BUSYNESS_AVERAGING_WINDOW = WEEK;
-const double TRANSFER_BUSYNESS_THRESHOLD = 0.00;
+const double TRANSFER_BUSYNESS_THRESHOLD = 0.05;
```
alexbcreiner0: I don't feel that you tested this number as much as I emphasized the need for. 0.05 effectively makes employment transfers impossible it seems. Here is how employment looks with the typical settings (except $\epsilon_A = 0.8$ and $N_p = 6$, seed 666495000) when the threshold is 0.05. This is in your branch but main looks the same:

Image

We need to discuss this and test more closely. If there is no reasonable number to set this to besides 0, then that is a larger problem with our transfer system than anything you're addressing here. 

I recommend setting it back to 0.0 or at most 0.005 until we have added it in as a parameter of the Overseer layer and tested it more closely.

src/Distributor.cpp
```diff
@@ -93,13 +96,11 @@ void Distributor::check_and_reorder_input(Product * product) {
...
         return;
     }
-    double distribution_time = 
-        static_cast<double>(distribution_quantity) / demands[consumer_good];
     Order * order = new Order(
             consumer_good,
             distribution_quantity,
             this,
-            distribution_time
+            distribution_quantity / production_rate
```
alexbcreiner0: This makes a lot of sense but I do worry what new problems will be created by replacing demand with production rate here. Maybe none.

alexbcreiner0 (approved):
Your changes here do a lot to sync reordering more closely to demand, and because of that add a good amount of stability to the system. Because of this, I'm going to approve the changes, but you should wait for Amittai to look at them as well before merging. Also see my comments for a couple of other small requested changes. I want to look at a particular test run I did with you though and alert you to some problematic things that I saw:

Settings here are the default to Overseer except that $N_p = 6$, $\epsilon_A = 0.8$, and the fixed seed is 666495000. I also changed the `EMPLOYMENT_TRANSFER_THRESHOLD` to 0 (see the comment). I'll email you the data so you can import it and run it without running the whole simulation again (just drop the two files in a folder of the Aggregation-Layer directory called `saved_results`, then click the Results menu. 

Image

Despite a massive overproduction stint at cycle 4000 and a crash at cycle 7000, your changes here are a big improvement from main, which looks like this on the same seed:
 
Image

So your improvements are real, but clearly we still have work to do. Nonetheless, this is good work.

header/Firm.h
```diff
@@ -60,7 +60,7 @@ class Firm : public Agent {
         std::vector<std::pair<Product *, double>>& deducted_inputs
     );
     double get_reorder_threshold(Product * product);
-    double get_pending_inventory_level(Product * product);
+    double get_needed_production_rate(Product * product);
```
Amittai-Aviram: I think I'm inclined to agree with Alex in calling this the resupply rate rather than the production rate, because it's the production rate of the upstream supplier.  (Anything even clearer might be too long — `get_needed_upstream_production_rate`, etc.) Please use your judgment, but, of course, make all the names consistent.

src/Distributor.cpp
```diff
@@ -75,15 +75,18 @@ void Distributor::check_and_reorder_input(Product * product) {
...
     Good * good = consumer_good->corresponding_good;
 
-    double reorder_threshold = demands[consumer_good] * FIRM_STOCKPILE_DURATION;
+    double reorder_threshold = get_reorder_threshold(consumer_good);
     log_demand(consumer_good, reorder_threshold);
 
-    double pending_inventory = get_pending_inventory_level(consumer_good);
-    log_pending_inventory(consumer_good, pending_inventory);
-    if (pending_inventory >= reorder_threshold || !reorder_threshold) {
+    int inventory = get_inventory_level(consumer_good);
```
Amittai-Aviram: I'm a bit worried about annoying side effects of type conversion here. We know that consumer goods are stored in the inventory in integral units, so `get_inventory_level` should return an integer-valued double. But I've already seen some cases — and I think you have as well — where, for some reason, the value you can get is 0.99999999 or something, and that gets converted to `int` as ... 0!  The safe thing, I think, would be to take the ceiling of the right-hand side. Please apply the same principle elsewhere — I might not always catch it.

src/Firm.cpp
```diff
@@ -170,24 +170,24 @@ double Firm::get_reorder_threshold(Product * product) {
...
+    if (production_rate <= 0) {
+        return;
+    }
+    double reorder_threshold = get_reorder_threshold(product);
     log_demand(product, reorder_threshold);
-    int pending_inventory = get_pending_inventory_level(product);
-    log_pending_inventory(product, pending_inventory);
-    if (pending_inventory >= reorder_threshold || !reorder_threshold) {
+    int inventory = get_inventory_level(product);
```
Amittai-Aviram: I guess you wanted to underestimate the inventory level, but I think it's safer just to get the actual level by storing it in a `double`.

src/Firm.cpp
```diff
@@ -170,24 +170,24 @@ double Firm::get_reorder_threshold(Product * product) {
...
+    if (inventory >= reorder_threshold || !reorder_threshold) {
         return;
     }
     double order_quantity = reorder_threshold * FIRM_REORDER_MAX_PROP;
```
Amittai-Aviram: I think you might want to raise this to the ceiling. One at least needs to do that for machines, but I don't think it hurts to do it in general, since that quantity will eventually become the "quantity" field in an order, of type `int`, and measured as an integer.

Amittai-Aviram (changes requested):
Yeah, this is beautiful!
Devin, I noted a couple of wee instances of possible numeric issues. Could you please make those changes, and maybe do your own extra walk-through to make sure that we've covered all such potential issues?

src/Distributor.cpp
lennyyyyyyyy: I think it's okay actually. We don't know if floating point precision errors will always be down. Also, in this usage of the inventory, it's better to err on the side of the lower value (to think that you need to reorder when it's not necessary is safer than not reordering when it is necessary)

src/Firm.cpp
lennyyyyyyyy: Using double is a good idea. I'll do that.

src/Firm.cpp
lennyyyyyyyy: Yeah, you're right. Currently it's obscurely rounding down when passed into a new order.

header/Constants.h
lennyyyyyyyy: Ok.

src/Firm.cpp
lennyyyyyyyy: Got it.

header/Firm.h
lennyyyyyyyy: Done.

Amittai-Aviram (approved):
LGTM

PR #283, opened 2026-07-20
Restrict reorders on time rather than quantity, add a slight correction to the resupply_rate

lennyyyyyyyy:
In main, the system looks as if it keeps inventories stable while keeping transfers low. However, given enough time, society collapses due to inputs running out. Theoretically, our `resupply_rate` should match demand, but it cannot do so perfectly forever. There ends up being a drift in the inventory that is just sort of ignored until it gets so low that it causes plan rejection, which causes the sudden collapse. At $\epsilon_c$ = 0.6, this sometimes happens after 10000 timesteps, as the system isn't under a lot of pressure.  At $\epsilon_c$ = 0.8 or 0.85, this happens much quicker. During the collapse, there are never a shortage of workers - it's just that the inputs have slowly dwindled. Putting a very slight negative feedback factor called `FIRM_RESUPPLY_URGENCY` works very well while maintaining very low transfers.

I've also redesigned `FIRM_REORDER_MAX_PROP` into `FIRM_REORDER_DURATION_PROP`. This is less important for stability, but I think is more logical. Essentially, reorders just aim to take `FIRM_REORDER_DURATION_PROP * FIRM_STOCKPILE_DURATION` amount of time. This has the same effect as `FIRM_REORDER_MAX_PROP` did before (as it aimed to have quantities in that proportion of the reorder threshold), but it's just more consistent now that we play with the number of workers. Before PR #280, when the resupply rate was about the same for every plan, this change actually wouldn't have made a difference.

Other things:
 - Did some refactoring in `draft_plan_from_order` and `draft_plan_and_return_order`.
 - Fixed a bug in `is_within_work_schedule`
 - Put a `requested_resupply_rate` field to the Order class to be more explicit. The log for `log_accepted_order` has changed, so Overseer will bug if that isn't commented as of now

In main, you can see the drift at $\epsilon_c$ = 0.6. Producer inventories look like random walks:
image

Fast collapse at $\epsilon_c$ = 0.8, which was a feasible consumption level before:
image

`devins-branch-2`, at $\epsilon_c$ = 0.6: 
image

Corresponding employment:
image

As expected, employment variance generally decreases with increasing people (300 in this case). Here's a run with no employment variance at all:
image
image

Employment variance increases again when $\epsilon_c$ = 0.8 (still 300 people):
image
image

Even $\epsilon_c$ = 0.8 can be smoothed out by higher numbers of people (500):
image
image

To avoid making the same mistake as PR #280, I'm probably gonna go through and at least get statistics for stability for differing amount of people and consumption difficulties.

Stability stats, 10 seeds (20000 timesteps):
500 people, $\epsilon_c$ = 0.6: 
 - 0 - stable, l.s.e.r (largest sectoral employment range) = 0
 - 1 - stable, l.s.e.r = 8
 - 2 - stable, l.s.e.r = 0
 - 3 - stable, l.s.e.r = 1
 - 4 - stable, l.s.e.r = 0
 - 5 - stable, l.s.e.r = 24
 - 6 - stable, l.s.e.r = 0
 - 7 - stable, l.s.e.r = 0
 - 8 - stable, l.s.e.r = 0
 - 9 - stable, l.s.e.r = 0

500 people $\epsilon_c$ = 0.8:
- 0 - stable, l.s.e.r = 14
- 1 - stable, l.s.e.r = 13
- 2 - stable, l.s.e.r = 20
- 3 - stable, l.s.e.r = 18
- 4 - stable, l.s.e.r = 19
- 5 - not stable
- 6 - stable, l.s.e.r = 13
- 7 - stable, l.s.e.r = 36
- 8 - stable, l.s.e.r = 25
- 9 - stable, l.s.e.r = 15


header/Constants.h
```diff
@@ -47,7 +47,8 @@ const int MACHINE_LIFETIME_MIN = 1000;
 const int MACHINE_LIFETIME_MAX = MACHINE_LIFETIME_MIN * 10;
 
 const int FIRM_STOCKPILE_DURATION = 1.5 * WEEK;
-const double FIRM_REORDER_MAX_PROP = 0.25;
+const double FIRM_REORDER_DURATION_PROP = 0.25;
```
Amittai-Aviram, line 50: I guess the name change is helpful, but the concept is still a little arbitrary, isn't it? We are saying that we only need to stock at least a quarter of what we use over time, where "what we use over time" is subject to decay toward a horizon.  Why a quarter? I wonder whether this should eventually be a function of something else, such as the population (which determines the overall consumption rate). Do you have any thoughts about that? (I remember that you first defined this constant as a kind of temporary hack to try to attain general stability — or that is how I remember it.)

src/Distributor.cpp
Amittai-Aviram, line 79: It seems to me more natural to move this down to below the declaration/initialization of `distribution_quantity`: (a) How much do we need? (b) Do we have enough on hand? (c) If we need more, how much do we need, in view of the orders already in process?  That would also move the declaration of `resupply_rate` closer to its use in the outgoing order. (This is obviously a matter of opinion.)

src/Firm.cpp
```diff
@@ -171,7 +168,11 @@ double Firm::get_reorder_threshold(Product * product) {
 }
 
 double Firm::get_needed_resupply_rate(Product * product) {
-    double needed_resupply_rate = demands[product];
+    double inventory = get_inventory_level(product);
+    double reorder_threshold = get_reorder_threshold(product);
+    double needed_resupply_rate = 
+        demands[product] 
+        * std::pow(reorder_threshold / inventory, FIRM_RESUPPLY_URGENCY) ;
```
Amittai-Aviram, line 175: Would it be possible to provide a mathematical rationale for this formula, in particular the exponentiation? I like the general idea of an exponential increase in the needed resupply rate as that ratio increases.

src/Firm.cpp
```diff
@@ -305,59 +306,84 @@ double Firm::calculate_quantity_produced_from_worker_suitability(Plan * plan) {
         Society::get_instance()->get_underlying_living_labor_per_unit(plan->order->product);
 }
 
-bool Firm::is_within_work_schedule() const {
+bool Firm::is_within_work_schedule(Plan * plan) const {
     return Sim::get_current_time_step() % DAY <
```
Amittai-Aviram: Store the current time step locally so that you do not need to fetch it again from `Sim`?

src/Firm.cpp
```diff
@@ -305,59 +306,84 @@ double Firm::calculate_quantity_produced_from_worker_suitability(Plan * plan) {
...
-            recorded_living_labor_per_unit[plan->order->product] *
-            WEEK /
-            Sim::get_work_days_weekly() / 
-            plan->local_work_hours_daily /
-            plan->order->requested_turnaround_time
-            );
+double Firm::get_max_order_quantity(Product * product) {
+    double max_order_quantity = std::numeric_limits<double>::infinity();
+    for (std::pair<Good * const, double>& input : product->inputs_per_unit) {
+        if (input.second <= 0.0) {
```
Amittai-Aviram: I know that this test does not hurt, but could the value ever be <= 0.0?

src/Firm.cpp
```diff
@@ -305,59 +306,84 @@ double Firm::calculate_quantity_produced_from_worker_suitability(Plan * plan) {
...
 
-    int workers_left = predict_workers_needed(draft_plan);
-    for (Person * worker : sorted_standby_workers) {
-        if (workers_left <= 0) { 
-            return;
-        }
-        draft_plan->workers.push_back(worker);
-        workers_left--;
-    }
     for (Person * unemployed_person : Society::get_instance()->get_unemployed_people()) {
```
Amittai-Aviram, line 337: I guess you are now prioritizing unemployed workers over workers on standby. Can you explain that decision? _Prima facie_, I would expect the opposite, because I would want to leave the unemployed workers for other firms unless I really needed them, whereas my standby workers are already with the firm and waiting for something to do.

src/Firm.cpp
```diff
@@ -305,59 +306,84 @@ double Firm::calculate_quantity_produced_from_worker_suitability(Plan * plan) {
...
+        available_workers.push_back(worker);
+        max_workers--;
     }
     for (Firm * firm : Society::get_instance()->get_firms()) {
-        if (workers_left <= 0) {
-            return;
+        if (max_workers <= 0) {
+            return available_workers;
         } 
         log_transfer_request();
```
Amittai-Aviram: I did not catch this before, but should we be logging a transfer request if `firm == this`? No request actually goes out in that case.

src/Firm.cpp
```diff
@@ -305,59 +306,84 @@ double Firm::calculate_quantity_produced_from_worker_suitability(Plan * plan) {
...
+    return available_workers;
+}
+
+int Firm::predict_workers_needed(Plan * plan) {
+    return std::ceil(
+            recorded_living_labor_per_unit[plan->order->product] *
+            WEEK /
+            Sim::get_work_days_weekly() / 
+            plan->local_work_hours_daily /
+            plan->order->requested_resupply_rate
```
Amittai-Aviram: I wish there were a way to make this arithmetic clearer. Could you please at least group `WEEK / Sim::get_work_days_weekly() / plan->local_work_hours_daily` with parentheses?  

Tell me if this is wrong. The requested supply rate is now many units per time step we need, and we need to know how many workers it will take to provide that rate.  Suppose it takes 2 hours to make one unit of a good, and we want 20 units every hour. Then it should take 10 workers, working continually, to make the 20 units in an hour. But then we need that production rate to be the average production rate even including periods during which people are not working, so we need to adjust the number upwards with a kind of scaling factor that is the ratio of hours needed to hours actually worked.  `WEEK / Sim::get_work_days_weekly` gives us a number of hours per day as if the hours of a full week were collapsed into a work week.  The ratio of this to the actual number of hours in a work day gives you the amount by which to scale the raw number of workers you would have needed if everybody worked all the time. Is all that correct?

If so, maybe even store that value first under a meaningfully-named variable and use that? Perhaps something like this? —
```
double work_week_adjustment = (WEEK / Sim::get_work_days_weekly()) / plan->local_work_hours_daily;
return std::ceil(
    (recorded_living_labor_per_unit[plan->order->product] / plan->order->requested_resupply_rate) *
    work_week_adjustment;
```
If this looks OK to you, maybe you could use a similar strategy in other functions that do this kind of scaling? Maybe you could have a little utility function to return the "work week adjustment"?

src/Firm.cpp
```diff
@@ -422,16 +448,33 @@ void Firm::assign_plan_dependent_fields(Plan * draft_plan) {
...
+    if (max_order_quantity <= 0) {
+        return nullptr;
+    }
+
+    Order * new_order = new Order(
+            order->product,
+            max_order_quantity,
+            order->customer,
+            order->requested_resupply_rate
+            );
```
Amittai-Aviram, line 456: We create a new order here without deleting the original, incoming order. This makes sense, because, before the customer chooses a producer, it may have to send that same order to several producers. But the order it eventually uses after accepting a draft plan is the one created here for the draft plan.  In my PRs, I have been deleting that order, with its plan, when it really is no longer needed, in the Price Controller. My changes still have not gone in because of all these other issues related to machines. But we also need to delete the original order, which is _not_ the order that the customer uses, once it chooses a producer.

src/Firm.cpp
```diff
@@ -422,16 +448,33 @@ void Firm::assign_plan_dependent_fields(Plan * draft_plan) {
...
+    draft_plan->order = new_order;
     draft_plan->firm = this;
     draft_plan->local_work_hours_daily = 
         Society::get_instance()->get_current_work_hours_daily();
-    assign_workers(draft_plan);
-    if (draft_plan->workers.empty()) {
+
+    int requested_workers = predict_workers_needed(draft_plan);
+    std::vector<Person *> available_workers = get_available_workers(requested_workers);
+    draft_plan->workers = available_workers;
```
Amittai-Aviram: Do you need the two statements? I think it results in two copies of the vector returned by `get_available_workers`.

src/Firm.cpp
```diff
@@ -305,59 +306,84 @@ double Firm::calculate_quantity_produced_from_worker_suitability(Plan * plan) {
...
+}
+
+double Firm::predict_timely_quantity(Plan * plan) {
+    return FIRM_STOCKPILE_DURATION * 
+            FIRM_REORDER_DURATION_PROP *
+            plan->workers.size() * 
+            plan->local_work_hours_daily *
+            Sim::get_work_days_weekly() /
+            WEEK /
+            recorded_living_labor_per_unit[plan->order->product];
```
Amittai-Aviram, line 385: Again — would it be possible to use some local variables to help make the reasoning behind this arithmetic clearer?

Amittai-Aviram (changes requested):
I like these changes! Could you please consider the further changes I requested? Also, maybe in your headnote, could you possibly present a more precise rationale for your exponential scaling of urgency? As I said in my comment, I like the idea, and it makes sense, but it does involve another arbitrary constant, and so could use a bit of explanation. Thanks!

src/Firm.cpp
```diff
@@ -305,59 +306,84 @@ double Firm::calculate_quantity_produced_from_worker_suitability(Plan * plan) {
         Society::get_instance()->get_underlying_living_labor_per_unit(plan->order->product);
 }
 
-bool Firm::is_within_work_schedule() const {
+bool Firm::is_within_work_schedule(Plan * plan) const {
     return Sim::get_current_time_step() % DAY <
-        Society::get_instance()->get_current_work_hours_daily() &&
+        plan->local_work_hours_daily &&
```
alexbcreiner0: I've never understood why we even have this quantity as a property of the plan. Seems very odd to me. Why are we doing this? I feel like this should not be a property of plan at all and we should just be grabbing the real thing from Society.

src/Firm.cpp
alexbcreiner0: No, I don't think so. I've always brushed this off as unimportant but it seems pointless to me as well.

src/Firm.cpp
```diff
@@ -305,59 +306,84 @@ double Firm::calculate_quantity_produced_from_worker_suitability(Plan * plan) {
...
+        if (input.second <= 0.0) {
+            continue;
+        }
+        double input_max_order_quantity = 
+                get_inventory_level(input.first) / input.second;
+                
+        max_order_quantity = 
+            std::min(max_order_quantity, input_max_order_quantity);
+    }
+    return max_order_quantity;
```
alexbcreiner0, line 325: Github makes it pretty annoying to just move things around huh?

src/Firm.cpp
alexbcreiner0: Unemployed hires first are how I thought it was all along. I was imagining an initial public campaign towards 'full employment' - a reasonable first act of any socialist regime!

src/Firm.cpp
alexbcreiner0: This is a good catch, it should be going after that line. That is probably why the numbers seem so high in Overseer.

src/Firm.cpp
alexbcreiner0: You're correct @Amittai-Aviram, the 'work week adjustment' term is actually the $\omega$ term I frequently use in my mathematical explanations. (Actually, it's $\frac{1}{\omega}$.) You're identifying what makes this term important  in general. Effectively here it's performing the unit conversion:

$$ \frac{x \textrm{ hours of labor}}{\omega} = y \textrm{ actual hours of time passing} $$

Previously, we were taking the local unit living labor value of the product $l$, and the order quantity $q$, to get the number of hours of labor ($x$ above), converting it to the actual amount of time that would translate to, and then dividing by the desired turnover time in order to get the total number of workers which would be needed to satisfy the plan in the desired timeframe. 

Devin's resupply rate here is (currently) just the number $R = \frac{q}{\tau}$, where $\tau$ is the turnover time. So the old equation changes: 

$$ \frac{lq}{\omega \tau} = \frac{l}{\omega} \frac{q}{\tau} = \frac{l}{\omega}R $$

which is a problem because what I see here is 

$$ \frac{l}{\omega R} $$

That _seems_ like a mistake. However, my best guess is that you (@lennyyyyyyyy) didn't intend for the `order_quantity` to belong here at all, but rather forgot to replace the fourth field of the Order created in `check_and_reorder` with this rate in place of the old turnover time calculation. If you had done that, then $R$ would be a number derived from $D$ which approaches zero as the rate of resupply approaches D, and approaches D as the rate of resupply goes to zero. So if there is no outstanding orders, then 

$$ \frac{l}{\omega}D $$

is the number of plan execution hours per unit times the number of units consumed per hour, i.e. the number of plan execution hours per hour, i.e. the number of workers needed to satisfy the rate of consumption. If we re-substitute $R$ in place of $D$, which is a portion of the resupply rate which is still needed, we get something which makes a lot of sense as the desired number of workers.

However, even with that charitable interpretation, your $R$ is still in the wrong place. It is in the denominator, when it should be in the numerator. This is probably another error.

However, even with these errors fixed, I am very nervous about replacing our old calculation with something like this. We should understand as a group what you are proposing: eliminate what was previously a simple stand in for natural behavior with a very specific instruction about what signals to send out when you require more of something. I'll return to say more about this when I'm writing the overall review of this PR.

src/Firm.cpp
alexbcreiner0: @Amittai-Aviram I think that the term (`local_work_hours_daily`*`work_days_weekly` / WEEK / `living_labor_per_unit`) amounts to the term $\frac{\omega}{l}$. I think that this represents the number of units which would be produced by a single worker per real hour of time elapsing. Then that is multiplied by the number of workers attached to the plan to give the number of units which would be produced per hour of real time if the plan were pursued. So @lennyyyyyyyy  either of those would be decent local variable names.

src/Firm.cpp
lennyyyyyyyy: When work hours update (which they don't in this branch), we don't want ongoing plans to update their work schedule. It'll throw off the predicted turnaround time and resupply rates.

src/Firm.cpp
```diff
@@ -422,16 +448,33 @@ void Firm::assign_plan_dependent_fields(Plan * draft_plan) {
...
+
+    int requested_workers = predict_workers_needed(draft_plan);
+    std::vector<Person *> available_workers = get_available_workers(requested_workers);
+    draft_plan->workers = available_workers;
+    int timely_quantity = std::min(draft_plan->order->quantity, static_cast<int>(predict_timely_quantity(draft_plan)));
+    if (timely_quantity <= 0) {
+        delete draft_plan->order;
         delete draft_plan;
         return nullptr;
     }
```
alexbcreiner0, line 470: So if I'm understanding things correctly, this can be zero for one of two reasons:
1. The maximum producible quantity of the good is 0, given the firm's inventory
2. The resupply rate was so low that no workers were assigned to the plan. 

Those are sensible reasons to stop the process. Am I understanding things properly?

src/Firm.cpp
lennyyyyyyyy: Yeah, because it doesn't even throw anything, I'll just get rid of it.

src/Firm.cpp
```diff
@@ -422,16 +448,33 @@ void Firm::assign_plan_dependent_fields(Plan * draft_plan) {
...
+    draft_plan->order->quantity = timely_quantity;
```
alexbcreiner0, line 471: And then here, you are actually setting the order quantity (for the first time, meaningfully), as the closest to `FIRM_REORDER_DURATION_PROP*FIRM_STOCKPILE_DURATION` times the units per hour of output expected from the number of workers assigned. Then tautologically this plan will end after exactly `FIRM_REORDER_DURATION_PROP*FIRM_STOCKPILE_DURATION` many hours. I see.

src/Firm.cpp
lennyyyyyyyy: What ends up happening in very stable and low pressure simulations is that firms end up not even needing the whole populace to work. Total employment ends up at 0.9, and firms just keep reworking the workers they already have. However, this creates a large gap in busyness. Considering the unemployed population first ensures that everyone is put into the system quickly.
 
Also, I think it's fine to assume that the disruptiveness of a transfer is less harmful than just not having a job.

src/Firm.cpp
lennyyyyyyyy: Fixed.

src/Firm.cpp
lennyyyyyyyy: Sure.

src/Firm.cpp
lennyyyyyyyy: Ok, I'll delete the order on the Firm::check_and_reorder side of things.

alexbcreiner0 (changes requested):
Devin, this was a very, very difficult PR to review. Amittai gave some substantive feedback here but just generally in addition to the lack of local variables and intermediate steps, there are also some very questionable choices for variable names, and an overall feeling that the changes you are introducing required for readability sake more significant edits to the code base than what you did (e.g. not merely repurposed functions but name changes to them, etcetera). However, after spending several hours on this, I think I understand the vision here. What you are doing marks a major departure for this project in two ways that need to be taken extremely seriously.

Orders, if this PR were approved, would no longer be something we can envision firm managers as consciously thinking about or using their best judgement on, or really as anything which is typically thought of in real life as an order anymore. An order according to this PR is really a pair of cybernetic signals, to be processed automatically into a plan. We're going from firms requesting concrete quantities delivered in a concrete amount of time, which made things easy to think about, and allowed for easy and intuitive applicability of our model to the real world - giving the real firms implementing our system room to use their own best judgement. 

The same can be said of the firm on the other side drafting the plan - they were making **locally rational decisions** in order to satisfy the order to the best of their abilities. With your replacements here, you are replacing that with an instruction: just follow this algorithm, don't question it, don't complain at what the turnover times turn out to be, and everything will be fine. Keep track of your signals and obey the plan. We have gutted our actors of their decision making agency in a major way.

I'm not saying that this departure isn't warranted, by the way. I'm just saying that it is something we should discuss and take seriously as a group.

On the other hand, this PR marks a very nearly triumphant departure away from thinking in terms of stocks and durations and towards thinking in terms of flows and rates. Substituting blocky concrete amounts and shifting more towards adjustments according to continuous rates is where the stability is being derived from. This much is a good thing, but I am starting to realize that there is a major contradiction in DEP between the principle of local rationality and the principle of operating according to cybernetic flows.

I am very uneasy about your net goal of having the firm's ultimate goal in the plan drafting process being just to complete it in a _constant, exogenous, and arbitrary_ amount of time at all costs. That really seems to be dubious to me, and is another separate way that this PR violates the principle of local rationality.

However, I think that there are a lot of other, _much_ simpler ways that we could potentially have solved this problem of inputs dwindling over time. I don't feel like the substitutions you make here are completely in the service of that. This is really the biggest issue I see with this PR. **I see no real causal explanation here as to why inventories were falling over time.** What I see here is a sidestepping of that - a redesign which makes the problem disappear, but without ever understanding where it came from. I think that if you really tried to find the exact thing causing that deficiency, we would find ourselves a much simpler fix.

Second, and in agreement with Amittai's point, the variable names are a mess, and more broadly the overall intention here is very hard to parse. The system here really feels like its outgrowing the functions that it takes place within. A lot of work needs to be done to make what is going on more clear.

What you initially call the order_quantity in the plan is not an order quantity, and I think you forgot to enter what you actually wanted to for the resupply_rate. Order probably needs a separate temporary field for this data you are giving it, and order_quantity should either be made null until actually calculated or made a property of the plan and removed from Order entirely. 

Overall, I think that the changes here are ambitious and very interesting, but not the changes we need at the immediate moment.

src/Firm.cpp
lennyyyyyyyy: Shortened.

src/Firm.cpp
lennyyyyyyyy: Pulled out the work week proportion into a separate function, and made the math clearer.

src/Firm.cpp
lennyyyyyyyy: Done.

src/Firm.cpp
alexbcreiner0: Has this been confirmed to be a significant destabilizing factor? I figured that was the reason but I'm really not sure that is desirable or necessary? Seems overly cautious. (Not suggesting it change in this PR, just commenting.)

PR #286, opened 2026-07-22
Fix non-deterministic execution, optimize builds, and adjust coverage

nathanjlim:
Resolve #276 
Resolve #284 
Resolve #285

Notes:
* Removed the `static` keyword from random distributions in `Product.cpp` (like `num_inputs_dist` and `living_labor_dist`) to fix the non-deterministic execution.
  * As Aviram noticed, passing a `min` and `max` of 1 into the `static` distribution resulted in an output of 3. This happened because a `static` local variable is initialized exactly once. On subsequent function calls, the distribution simply ignored the new parameters (like `min=1, max=1`) and relied on its internal state from the very first call. When bounds fluctuated, this caused the distribution to generate out-of-bounds indices. C++ reads the memory at that address, which changes at every run, randomized values into the simulation, and destroys determinism.
    * Tested with: `bin/sim -e 1 -n 10000 -p 100 -h 8 -w 5 -g 5 -m 0 -r 5 -d 5 -s 0.1 -a 5 -v 0.1 --production_difficulty 0.5 --consumption_demand 0.6 --init_prices labor_values` and `make runtests`.
 
* Added `-O2` as the global default in the `Makefile` to speed up simulation execution as well as implemented an override that swaps to `-O0` when running `make coverage` for accuracy.
  * Defined as `OPT_FLAG=-O0` & `=-O2`.

* Removed `lcov` and its HTML outputs, switching to `gcov` by implementing a script that parses its raw data, filters out empty libraries, and prints a clean total percentage directly to the terminal.
  * `make score` also confirms it forces `-O0` for the tests.

* Replaced datatypes and added approximations in `test_Producer.cpp` to fix precision issues.


header/Firm.h
```diff
@@ -50,6 +50,7 @@ class Firm : public Agent {
     std::unordered_map<Product *, double> demands;
     std::unordered_map<Product *, std::unordered_set<Order *>> product_to_outbound_orders;
     std::unordered_map<Product *, double> recorded_living_labor_per_unit;
+
```
Amittai-Aviram: Could you please remove this blank line? I prefer to use blank lines very sparingly.  (Having the lines clumped together makes it easier to see the blocks at a glance, and especially to distinguish one function from another.)

Amittai-Aviram (changes requested):
These are great changes so far! It turns out that the -O2 flag makes a _huge_ difference!

Alas, as things stand right now, I'm still not getting reproducible results on two subsequent runs, I think because this PR does _not_ include substituting ordered for unordered (hash-based) maps. Were you thinking of that as a separate issue? If so, could you please write it up?

I think what we need is a `debug` target, so that you can run `make debug`, which will set optimization to `O1`, add the `-g` flag (which should be removed otherwise), and define a symbol in the command (such as `DEBUG`) such that the preprocessor will replace every instance of `std::unordered_map` with `std::ordered_map`. This all is distinct from the `-e` flag in the command to _run_ the simulation. (Maybe the conditional compilation should also override that option and just always set the seed to be fixed and have a value of 1 or whatever.  Please bring that up at the next team meeting.)

My PR #288 has (finally) gone in, so you have some merge conflicts that you'll have to resolve. In view of that fact, I tend to think that it would be better if you could go ahead and implement the `debug` build rule as described above as part of _this_ PR. But let me know if you disagree with that.

header/Firm.h
nathanjlim: Just to give a quick recap on the non-determinism hunt over the last couple of days, I originally thought that the random distributions in `Product.cpp` that I mentioned in the PR notes fixed the issue. (I forgot to add that into the commits here in this PR - noted by your comment).

However, the simulation was still drifting, and it turned out raw memory was still causing an issue that changed the iteration order. In order to fix this, I introduced static ID counters to give the objects a deterministic order. I added variables like `static unsigned int unique_plan_id = 0;` (and did the same for Product, Order, and Person) so every object gets a permanent ID when created. This makes sure that the new `ordered std::map` and `std::set` sorts by IDs instead of memory addresses.

I also added `make debug` to the Makefile (`-O1 -g -DDEBUG`). In the code, I set up `#ifdef DEBUG` blocks to swap the `unordered` for `ordered std::map` and `std::set`. 

- And added a check in `main.cpp` that intercepts the `-e` flag and forces the seed to `1` in debug mode.

I ran some basic testing and believe everything should be good to go, but it wouldn't hurt if you tried it on your end as well. 

Let me know if you run into any issues when you test it on your end, or if you have any questions about the changes!

header/Firm.h
```diff
@@ -14,11 +16,27 @@
...
 struct Order;
+struct Plan;
 struct Product;
 class Firm;
 class Producer;
 class Society;
 
+#ifdef DEBUG
+struct ProductID { 
+    bool operator()(const Product* a, const Product* b) const;
```
Amittai-Aviram, line 27: Style; please write `Product * a`, etc. — space before and after the asterisk.

header/Firm.h
```diff
@@ -14,11 +16,27 @@
...
 class Firm;
 class Producer;
 class Society;
 
+#ifdef DEBUG
+struct ProductID { 
+    bool operator()(const Product* a, const Product* b) const; 
+};
+struct OrderID { 
+    bool operator()(Order* a, Order* b) const;
```
Amittai-Aviram, line 30: Style: see above.

Amittai-Aviram (commented):
Is it impossible to do this? —
```
#ifdef DEBUG
#define unordered_map map
#define unordered_set set
#endif
```
Then you could leave the rest of the code basically untouched. This is not exactly standard, because it's risky if the token `unordered_map` or `unordered_set` ever occurs in the code as an _identifier_. However, we do not, in fact, name identifiers this way, and this, itself, would have been a pretty bad practice, I think.

In addition, you could then put the above preprocessor directives in a separate header file — say, `container_types.h`, and simply `#include` that header file in each header file whose code uses those types. That would avoid clutter as much as possible.

Amittai-Aviram:
The `runtests` target works great! But I don't know how to test deterministic execution. There seems to be no `debug` target.  Also, is it possible that you forgot to include the new header with `git add`? I do not have it on  my end.

PR #287, opened 2026-07-22
Minimal fix for main - resupply rate correction, consistent deadlines without limiting local rationality

lennyyyyyyyy:
This PR is meant to have the exact same effect as PR #283, without changing the Order at all, and without making major changes in `draft_plan_for_order` and `draft_plan_and_return_order`. As far I can tell, the stability and employment variance is just as good here as it is there.

It seems that plans having consistent turnaround times, now that workforce varies between plans, does much more for stability than having consistent quantities. In PR #283, this was implemented by the drafting firm always aiming for a "timely quantity". This seemed unnecessarily controlling, as this deadline was an arbitrary rule of the system imposed onto the drafting firms. Here, I use a different strategy to get the exact same result:
- The ordering firm gives its deadline through the order that it sends, which is consistent between reorders of the same product. Its requested quantity will always be that deadline multiplied by the resupply rate.
- The drafting firm acts rationally and in goodwill to fulfill the information given in the order to the best of its ability. It does the following:
    - It adjusts the order quantity if it doesn't have enough inputs to fulfill the original quantity. *Now, it does not touch the deadline while doing so.*
    - It tries to hire the amount of workers required to make the adjusted quantity by the deadline.
    - It reevaluates if it can meet the deadline (it may have not been able to attain all the workers it wanted), and *prioritizes the deadline over the quantity*. Thus, it adjusts the quantity again to be able to make the deadline.

Other things:
- Kept the correction to the resupply rate.
- Now that plan lengths are actually consistent, pending inventory makes sense to add back, preventing overordering.
- Made the math in `predict_workers_needed` and `predict_turnaround_time` a lot clearer. The functionality of these methods are the exact same.
- Workers are taken from the unemployed pool first, instead of after standby workers.
- Small fixes in `is_within_work_schedule` and `predict_labor_hours`


src/Firm.cpp
```diff
@@ -186,17 +189,17 @@ void Firm::check_and_reorder_input(Product * product) {
     }
     double reorder_threshold = get_reorder_threshold(product);
     log_demand(product, reorder_threshold);
-    double inventory = get_inventory_level(product);
+    double inventory = get_pending_inventory(product);
     if (inventory >= reorder_threshold || !reorder_threshold) {
         return;
     }
-    int order_quantity = std::ceil(reorder_threshold * FIRM_REORDER_MAX_PROP);
+    int deadline = FIRM_STOCKPILE_DURATION * FIRM_REORDER_DEADLINE_PROP;
```
Amittai-Aviram: You might as well store this as a `double`, since that is how you will use it.

src/Firm.cpp
```diff
@@ -305,21 +308,34 @@ double Firm::calculate_quantity_produced_from_worker_suitability(Plan * plan) {
...
 
-bool Firm::is_within_work_schedule() const {
-    return Sim::get_current_time_step() % DAY <
-        Society::get_instance()->get_current_work_hours_daily() &&
-        Sim::get_current_time_step() / DAY % 7 <
-        Society::get_instance()->get_current_work_days_weekly();
+bool Firm::is_within_work_schedule(Plan * plan) const {
+    int time = Sim::get_current_time_step();
+    return time % DAY < Society::get_instance()->get_current_work_hours_daily() &&
+        time / DAY % 7 < Society::get_instance()->get_current_work_days_weekly();
```
Amittai-Aviram, line 398: Please parenthesize to group: `(time / DAY) % 7`.

src/Firm.cpp
```diff
@@ -305,21 +308,34 @@ double Firm::calculate_quantity_produced_from_worker_suitability(Plan * plan) {
...
+    double pending_inventory = get_inventory_level(product);
+    for (Order * order : product_to_outbound_orders[product]) {
+        pending_inventory += order->quantity;
+    }
+    return pending_inventory;
+}
+
+double Firm::get_plan_work_week_proportion(Plan * plan) {
+    return static_cast<double>(plan->local_work_hours_daily) 
+        * Society::get_instance()->get_current_work_days_weekly()
```
Amittai-Aviram: Likewise — please group the first two factors with parentheses so that it's clear that this is the ratio (hours worked in a week ) / (all hours in a week).

src/Firm.cpp
```diff
@@ -305,21 +308,34 @@ double Firm::calculate_quantity_produced_from_worker_suitability(Plan * plan) {
...
+        * get_plan_work_week_proportion(plan);
     return std::ceil(
             plan->order->quantity *
-            recorded_living_labor_per_unit[plan->order->product] *
-            WEEK /
-            Sim::get_work_days_weekly() / 
-            plan->local_work_hours_daily /
-            plan->order->requested_turnaround_time
+            recorded_living_labor_per_unit[plan->order->product] /
+            work_time
```
Amittai-Aviram, line 422: Please group the  numerator factors together in parentheses.

src/Firm.cpp
```diff
@@ -350,35 +366,44 @@ void Firm::assign_workers(Plan * draft_plan) {
...
         }
         workers_left -= transfers.size();
     }
 }
 
+void Firm::adjust_quantity_for_deadline(Plan * plan) {
+    double living_labor_per_timestep = plan->workers.size() * get_plan_work_week_proportion(plan);
+    int maximum_quantity_for_deadline = 
+        plan->order->requested_turnaround_time *
+        living_labor_per_timestep /
```
Amittai-Aviram, line 469: Please enclose these two factors in parentheses.

src/Firm.cpp
```diff
@@ -350,35 +366,44 @@ void Firm::assign_workers(Plan * draft_plan) {
...
         for (Person * transfer : transfers) {
             draft_plan->workers.push_back(transfer);
         }
         workers_left -= transfers.size();
     }
 }
 
+void Firm::adjust_quantity_for_deadline(Plan * plan) {
+    double living_labor_per_timestep = plan->workers.size() * get_plan_work_week_proportion(plan);
+    int maximum_quantity_for_deadline =
```
Amittai-Aviram, line 467: Do we really want this to be the floor of the computation? Why?

src/Firm.cpp
```diff
@@ -350,35 +366,44 @@ void Firm::assign_workers(Plan * draft_plan) {
...
-           Sim::get_work_days_weekly() / 
-           plan->local_work_hours_daily /
-           plan->workers.size();
+           recorded_living_labor_per_unit[plan->order->product] /
+           labor_hours_per_timestep;
 }
 
 double Firm::predict_labor_hours(Order * order, std::vector<Person *>& workers) {
     if (workers.empty()) {
         throw std::runtime_error("Cannot predict labor hours with 0 workers: " + order->product->product_name);
```
Amittai-Aviram: This should use `product->id` rather than `product->product_name`. I already fixed it, I believe, in PR #288, so please just be sure to merge from `main`.

src/Firm.cpp
```diff
@@ -350,35 +366,44 @@ void Firm::assign_workers(Plan * draft_plan) {
...
     }
     return order->quantity *
-           recorded_living_labor_per_unit[order->product] / 
-           workers.size();
+           recorded_living_labor_per_unit[order->product];
```
Amittai-Aviram: Likewise, this is fixed (IIRC) in PR #288 .

src/Firm.cpp
```diff
@@ -428,6 +453,11 @@ Plan * Firm::draft_plan_for_order(Order * order) {
     draft_plan->local_work_hours_daily = 
         Society::get_instance()->get_current_work_hours_daily();
     assign_workers(draft_plan);
+    adjust_quantity_for_deadline(draft_plan);
+    if (order->quantity <= 0) {
+        delete draft_plan;
```
Amittai-Aviram, line 547: Should the order also get deleted? Deleting the plan does not delete the order.

Amittai-Aviram (changes requested):
I think this is a good change set, but please note the small additional changes I requested. And, of course, I would defer to @alexbcreiner0 's view.

PR #288, opened 2026-07-23
Implement machine life cycle

### LINKED ISSUE #203: Implement the replacement cycle of machines

Amittai-Aviram:
Resolves #203 .
Resolves #247 .
Resolves #281 .

# Essential Features
- We track input (raw materials) and machinery expenditures through a per-plan inventory, filled by transfer from the firm's general inventory in `Firm::start_plan`, and decremented appropriately in `Firm::move_plan_forward_one_step`.
  - `Plan::outlays` is a deep (i.e., independent) copy of `Plan::inventory` in `Firm::start_plan` and is not touched again until `Firm::end_plan`, where it is updated to reflect actual expenditures.
  - `Plan::needed_this_step` is a map structured the same way as `Plan::inventory`, but re-used on every time step in `Firm::move_plan_forward_one_step` to hold how much of each input and machine capacity must be used on the current step of plan execution. We use this data structure to terminate the plan "early" if we do  not have enough left of any necessary input or machine capacity to execute the next step.
- Once a firm starts a plan, it commits to completing it (partially or fully).
  - In `Firm::start_plan`, the firm first _marks the plan as in progress_ and places it into the `plans_in_progress` queue (vector), so that `Firm::move_plans_forwarsd_one_step` will always try to execute it. The firm then tries to pull input goods and reserved machinery capacity from the general inventory to the plan-local inventory. If any input (including machine capacity) is insufficient, it rolls back any transfers of inventory made so far and _returns_, so that, on the next time step, it will try the operation again, with the expectation that the firm's inventory will get replenished sooner or later enough to carry the plan forward.
  - In `Firm::move_plan_forward_one_step`, the firm follows a similar logic to that of `start_plan` to check whether the plan's inventory is adequate to execute the next step.  If the plan's inventory is inadequate, it checks whether there is enough of the given input (or machinery capacity) to transfer into the plan's inventory and does so if there is. If not, it returns to the firm's inventory the quantities of those inputs that it had attempted to transfer into the plan's inventory.
-  In `Firm::end_plan`, we deduct from `Plan::outlays` any amount left of any input or machine capacity. 
- The Price Controller uses the `Plan::outlays` field of each plan to aggregate the uses of all inputs and machines over all plans in its (moving) plan history window.  The total amount used for these indirect costs, divided by the total number of units, is then added to the price.

Screenshot 2026-07-23 at 11 28 24 AM


src/Producer.cpp
```diff
@@ -67,10 +85,16 @@ double Producer::get_max_order_quantity(Product * product) {
...
         double input_max_order_quantity = 
                 get_inventory_level(input.first) / input.second;
-                
         max_order_quantity = 
             std::min(max_order_quantity, input_max_order_quantity);
     }
+    for (Machine * machine : product->machines_needed) {
+        double machine_max_order_quantity =
+            (input_inventory[machine] * machine->lifetime) /
+            recorded_living_labor_per_unit[machine];
```
alexbcreiner0: This should be `recorded_living_labor_per_unit[product]`, not `recorded_living_labor_per_unit[machines]`. I think you might have confused yourself with the name `machine_max_order_quantity`, but the thing we are still trying to calculate here is the maximum amount of the _product_ of the order that can be produced, if a particular machine were the only consideration. So we take the number of hours of machine use we have left (`input_inventory[machine] * machine[lifetime]`), and divide it by the living labor required for the product.

src/Firm.cpp
alexbcreiner0: This is something that I missed myself, but `machines` here is never initialized to anything anywhere in the simulation. It's empty. So right now the machinery budget is always set to zero. A cheap fix would be to simply replace this line with `for (Machine * machine : draft_plan->order->product->machines_needed) {

And then eventually we can delete or maintain this list in a future patch, because this is the only place where it is used. 

At present, this only effects the debt and pooled account numbers, so has no bearing on stability.

src/Firm.cpp
```diff
@@ -202,70 +202,119 @@ void Firm::check_and_reorder_input(Product * product) {
...
+                        deducted_inputs
+                        )) {
+                return_inputs_to_inventory(deducted_inputs);
+                return;
+            }
+        }
+    }
+    for (std::pair<Product *, double> input : deducted_inputs) {
+        plan->inventory[input.first] += input.second;
+    }
```
alexbcreiner0, line 297: You are forgetting to add the deducted inputs to the `plan->outlays`. This will cause the PriceController to underestimate machine input costs if not done.

src/Producer.cpp
```diff
@@ -81,27 +105,24 @@ Order * Producer::draft_plan_and_return_order(const Order * order) {
...
     if (!draft_plan) {
         return_order->status = Order::kOrderRejected;
         return return_order;
     }
-
     double max_order_quantity = get_max_order_quantity(order->product);
-
+    if (order->product->product_type == Product::ProductType::kTypeMachine) {
+        max_order_quantity = std::ceil(max_order_quantity);
+    }
```
alexbcreiner0: This is something that I noticed and didn't bring up earlier because I was able to get a stable run without doing it, but it still really needs addressing. Taking the ceiling of the MAXIMUM order quantity like this undermines that entire calculation, and introduces unnecessary risk of `start_plan` failing. 

You can and should safely delete this block entirely. If the fear was that we wanted to ensure that machines must be produced as a full unit or not at all, that will be handled directly below with the `feasible_quantity` calculation, which casts the number as an int and floors it. If it floors to zero, it will be directed below that, and the plan drafting will stop.

alexbcreiner0 (changes requested):

src/Producer.cpp
Amittai-Aviram: Yeah — sorry! That was just an oversight. Corrected.

src/Firm.cpp
Amittai-Aviram: Corrected. Ugh! That iteration goes way back to before we started modeling machine capacity as part of the inventory. I have finally also removed the obsolete `machines` field from the `Firm` class altogether.

src/Firm.cpp
Amittai-Aviram: Yeah! Another oversight. Thanks. Corrected.

src/Producer.cpp
Amittai-Aviram: Corrected. Yeah, that code predated your revision of `Producer::draft_plan_and_return_order`.

alexbcreiner0 (approved):

src/Firm.cpp
```diff
@@ -202,70 +202,120 @@ void Firm::check_and_reorder_input(Product * product) {
...
 
 void Firm::start_plan(Plan * plan) {
-    std::vector<std::pair<Product *, double>> deducted_inputs;
-
-	for (std::pair<Good * const, double>& input :
-            plan->order->product->inputs_per_unit) {
-        double required_input = input.second * plan->order->quantity;
-        if (!remove_input_from_inventory(input.first, required_input, deducted_inputs)) {
-            rollback_plan_inputs(plan, deducted_inputs);
+    plans_in_progress.insert(plan);
```
lennyyyyyyyy, line 226: Is it intended to put the plan in progress even if `rollback_plan_inputs` is called?

src/Firm.cpp
```diff
@@ -202,70 +202,120 @@ void Firm::check_and_reorder_input(Product * product) {
...
-    double ideal_quantity_produced = calculate_quantity_produced_from_worker_suitability(plan);
-    double quantity_produced = std::min(ideal_quantity_produced, plan->quantity_remaining);
+    double expected_quantity_produced =
+        calculate_quantity_produced_from_worker_suitability(plan);
+    double quantity_produced =
+        std::min(expected_quantity_produced, plan->quantity_remaining);
     if (quantity_produced <= 0.0) {
+        end_plan(plan);
         return;
     }
```
lennyyyyyyyy, line 268: I think that you might find that you can get a "free" quantity that is produced that the plan may not have had the inputs for, since you check that stuff after. Isn't the better solution to simply use the inventory as bounds (machines as time bounds, and goods as quantity bounds) to figure out `quantity_produced`? Then, if that's less than `expected_quantity_produced`, you know that the plan must end.

lennyyyyyyyy (commented):
Most of this looks good. I like the changes to the PriceController.

src/Firm.cpp
Amittai-Aviram: Yes, that is intended. Otherwise, `move_plans_forward_one_step` will not cycle back to this plan again — it will essentially disappear (at least I think so). That is why it is safe to `return` after this — the plan is still active. On the next time step. `move_plans_forward_one_step` will try to start this plan again.

src/Firm.cpp
Amittai-Aviram: `quantity_produced` here means the quantity _about to be produced_ if we get past the checks. (Perhaps a name change would help, but that complicates the rest of the code a bit.) That quantity is not actually produced until the inventory quantities are successfully withdrawn, the labor hours are accounted for, the workers are paid, and the quantity remaining is decreased. That is, the actual work of this hour starts at line 298. Everything before that is just checking whether the hour can go forward. Note that `expected_quantity_produced` and `quantity_produced` are local variables — they do not reflect the state of the plan yet.

PR #291, opened 2026-07-30
Big stability patch

alexbcreiner0:
A big PR that mostly resolves the stability problems that we have been grappling with this summer. This PR is to be paired with a branch of the [LTE-Data-Aggregation-Layer](https://github.com/BC-LTEWG/LTE-Data-Aggregation-Layer) called `late-july-patch`. That will be merged into that repo alongside this one. 
- Machines are now initialized with more realistic numbers. Machine costs are generally orders of magnitude higher than normal goods, controlled via the `MACHINE_SCALE_MULTIPLIER` constant.
- Turnover times are now an invariant, set as a fixed proportion of the `FIRM_STOCKPILE_DURATION`. Devin, if you want to write up an actual issue for this, I can add the number here as resolved. That would be a big help. 
- Demand split into a producer and consumer component, so that at high difficulties of production, where producer demand dwarfs consumer demand, a brief oversupply of producer inventories can't cause a nosedive of the demand signals towards the consumer goods demand, such that pending inventory can never keep up, leading to economic catastrophe. (See Devin's post [here](https://github.com/BC-LTEWG/Labor-Time-Economy-Simulation/discussions/290) for more details)
- Various small fixes to the machine logic to avoid crashes. In particular, `get_reorder_threshold` now takes the maximum of the usual threshold and the inventory that needs to be held in order to produce a minimal number of units of goods in a firm's catalog, controlled by the `FIRM_STOCKPILE_MINIMUM` constant. 
- Improved logging for when plans stall, either within `start_plan` or `move_plan_forward_one_step`. Plans now have ids, allowing Overseer to keep track of which specific plans are stalled and when stalled plans are resolved. Additional text logging shows the specific missing inventory which caused the stallage.
- When a plan stalls mid-execution, a firm will now try to place an emergency order for the missing good. This should not happen in normal circumstances.
- Plan failures are logged more accurately and plan failure reasons should be more reliable when viewed in Overseer
- Added logging for average weekly lead times, order quantities, and team sizes. 
- Refactored and renamed various things to make the code more readable. In particular, I broke down `add_to_catalog` into several functions, not because anything in it was wrong, but rather because `add_to_catalog` is not supposed to just be an initialization helper - it is something which we might potentially want to use during a simulation, if in the future we decide it is worthwhile to allow firms to conditionally shift their production priorities from one good to another, or introduce new goods into the economy post-initialization. If we tried to do this with `add_to_catalog` in its current state, all of the firm's inventory would be completely overridden when we did that.

There are probably some other things I forgot to list here. Overall, this marks a major milestone in the project, as we can now simulate virtually any difficulty of production reliably.


header/Firm.h
```diff
@@ -26,60 +26,84 @@ class Firm : public Agent {
...
     virtual void add_to_catalog(Product * product) = 0;
+    virtual void initialize_inventory() = 0;
+    void inject_randomness_into_demand();
     double get_avg_productivity();
     virtual double get_inventory_level(Product * product);
     void receive_shipment(Order * order);
     void receive_shipment(Plan * plan);
     void receive_payment(Plan * plan, double transaction_amount);
     double get_busyness();
+    std::unordered_set<Product *> get_catalog();
```
alexbcreiner0: this is no longer necessary. will remove

header/Constants.h
```diff
@@ -42,13 +44,19 @@ const double DEFAULT_CONSUMPTION_DEMAND = 0.6;
...
 const int FIRM_STOCKPILE_DURATION = 1.5 * WEEK;
+const double FIRM_STOCKPILE_MINIMUM = 1.0;
+
 const double FIRM_REORDER_MAX_PROP = 0.25;
-const int DEMAND_AVERAGING_WINDOW = MONTH;
+const double FIRM_REORDER_URGENCY = 0.05;
+const int DISTRIBUTOR_DEMAND_AVERAGING_WINDOW = 2*WEEK;
+const int PRODUCER_DEMAND_AVERAGING_WINDOW = 4*MONTH;
 const int BUSYNESS_AVERAGING_WINDOW = WEEK;
+const int TEAM_SIZE_AVERAGING_WINDOW = 4;
```
alexbcreiner0: this is obsolete. will remove.

header/Producer.h
```diff
@@ -16,11 +16,13 @@ class Person;
 
 class Producer : public Firm {
   public:
+    inline static bool wiretripped = false;
```
lennyyyyyyyy: Are you planning to keep this wire? If you are, you should prob name it something else.

header/Producer.h
```diff
@@ -29,7 +31,9 @@ class Producer : public Firm {
   private:
     std::unordered_map<Firm *, Plan *> customer_to_draft_plan;
 
-    double get_max_order_quantity(Product * product);
+    double get_max_order_quantity(const Order * order);
+    void adjust_workers_for_quantity(Plan * draft_plan);
```
lennyyyyyyyy: Unneeded.

src/Distributor.cpp
```diff
@@ -35,23 +38,33 @@ void Distributor::add_to_catalog(Product * product) {
...
     ConsumerGood * consumer_good = static_cast<ConsumerGood *>(product);
     Good * good = consumer_good->corresponding_good;
-    demands[good] = 
-        demands[consumer_good] = 
+    consumer_demands[good] = 
+        consumer_demands[consumer_good] = 
         consumer_good->mean_consumption_frequency 
         * Sim::get_num_people() 
         / Sim::get_num_distributors();
     static std::normal_distribution<double>
```
lennyyyyyyyy: Should be deleted.

header/Constants.h
```diff
@@ -42,13 +44,19 @@ const double DEFAULT_CONSUMPTION_DEMAND = 0.6;
...
 const int PRICE_AVERAGING_WINDOW = MONTH;
 
-const int MACHINE_LIFETIME_MIN = 1000;
-const int MACHINE_LIFETIME_MAX = MACHINE_LIFETIME_MIN * 10;
+const int MACHINE_LIFETIME_MIN = 100;
+const int MACHINE_LIFETIME_MAX = MACHINE_LIFETIME_MIN * 4;
+const int MACHINE_SCALE_MULTIPLIER = 100;
 
 const int FIRM_STOCKPILE_DURATION = 1.5 * WEEK;
+const double FIRM_STOCKPILE_MINIMUM = 1.0;
```
lennyyyyyyyy: Not sure if you need this as a constant. Being able to produce the minimum non-zero order size is already non-arbitrary.

src/Distributor.cpp
```diff
@@ -90,29 +104,33 @@ void Distributor::check_and_reorder_input(Product * product) {
...
+    int lead_time = 
+        get_inventory_level(good) / resupply_deficit;
+
+    lead_time = std::min(
+        lead_time,
+        static_cast<int>(
+            FIRM_STOCKPILE_DURATION * 
+            FIRM_REORDER_MAX_PROP
+        )
+    );
```
lennyyyyyyyy: I think I was wrong implementing this actually. This should be 
`double lead_time = static_cast<int>(get_inventory_level(good)) / resupply_deficit;`
and
`lead_time = std::min(lead_time, FIRM_STOCKPILE_DURATION * FIRM_REORDER_MAX_PROP);`

src/Firm.cpp
```diff
@@ -167,56 +203,80 @@ Producer * Firm::send_order(Order * order) {
...
     }
     double reorder_threshold = get_reorder_threshold(product);
     log_demand(product, reorder_threshold);
-    double inventory = get_inventory_level(product);
+    double inventory = get_pending_inventory(product);
     if (inventory >= reorder_threshold || !reorder_threshold) {
         return;
     }
-    int order_quantity = std::ceil(reorder_threshold * FIRM_REORDER_MAX_PROP);
+    int lead_time = FIRM_STOCKPILE_DURATION * FIRM_REORDER_MAX_PROP;
```
lennyyyyyyyy: I think same thing here, it should be `double lead_time = FIRM_STOCKPILE_DURATION * FIRM_REORDER_MAX_PROP;`

No idea why I thought the requested deadlines were ints.

src/Distributor.cpp
```diff
@@ -90,29 +104,33 @@ void Distributor::check_and_reorder_input(Product * product) {
...
     Order * order = new Order(
             consumer_good,
-            distribution_quantity,
+            lead_time * resupply_deficit,
```
lennyyyyyyyy: Same as in Firm.cpp, this should probably be `int order_quantity = std::ceil(lead_time * resupply_deficit);`

src/Firm.cpp
```diff
@@ -472,31 +664,62 @@ void Firm::assign_plan_dependent_fields(Plan * draft_plan) {
...
     if (draft_plan->workers.empty()) {
+        insufficient_workers = true;
+    }
+    adjust_quantity_for_deadline(draft_plan);
+    if (order->quantity <= 0) {
+        if (insufficient_workers) {
+            log_reorder_failure(order->product, "not_enough_workers_available");
+        } else {
+            log_reorder_failure(order->product, "insufficient_resources");
+        }
```
lennyyyyyyyy: I think this can be simplified to
```
if (draft_plan->workers.empty()) {
   log_reorder_failure(order->product, "not_enough_workers_available");
} else if (order->quantity <= 0) {
   log_reorder_failure(order->product, "insufficient_resources");
}
```

src/Firm.cpp
```diff
@@ -472,31 +664,62 @@ void Firm::assign_plan_dependent_fields(Plan * draft_plan) {
...
+
+void Firm::update_average_team_size(Plan * plan) {
+    double observed_team_size = static_cast<double>(plan->workers.size());
+    Product * product = plan->order->product;
+
+    average_team_sizes[product] = 
+        1.0 / TEAM_SIZE_AVERAGING_WINDOW * observed_team_size +
+        (1.0 -  1.0 / TEAM_SIZE_AVERAGING_WINDOW) * average_team_sizes[product];
+}
+
```
lennyyyyyyyy: To delete.

src/Producer.cpp
```diff
@@ -32,54 +33,66 @@ Logger::Client Producer::get_client_type() {
...
-                );
-    double machine_use_per_unit =
-        product->living_labor_per_unit / average_team_size;
+        Sim::get_num_producers() +
+        Sim::get_num_distributors();
+
+    average_team_sizes[product] = std::max<double>(
+            Sim::get_num_people() / starting_num_firms,
+            1.0
+        );
```
lennyyyyyyyy: Change back to `double average_team_size =`

src/Product.cpp
```diff
@@ -48,29 +48,42 @@ void Product::set_inputs() {
...
     for (int index : indices) {
         inputs_per_unit[goods[index]] =
             input_per_unit_dist(Sim::get_random_generator());
     }
+    // if (product_type == kTypeMachine) {
+    //     living_labor_per_unit *= MACHINE_SCALE_MULTIPLIER;
+    //     Society::get_instance()->
+    //         set_underlying_living_labor_per_unit(this, living_labor_per_unit);
+    // }
 }
```
lennyyyyyyyy, line 57: Clean up?

src/Society.cpp
```diff
@@ -166,7 +169,7 @@ void Society::set_initial_products() {
         products.push_back(new_machine);
     }
     for (Product * product : products) {
-        product->set_inputs();
+        product->set_inputs(); // These should be in the contructor
```
lennyyyyyyyy: Actually, they can't be in the constructor, because you haven't constructed all the goods to pick the inputs from.

src/Society.cpp
```diff
@@ -264,46 +292,52 @@ void Society::log_total_employment() {
...
+        machine->living_labor_per_unit *=
+            MACHINE_SCALE_MULTIPLIER;
 
-    double divisor = get_max_eigenvalue(productive_matrix) / Sim::get_difficulty_of_production();
+        set_underlying_living_labor_per_unit(
+            machine,
+            machine->living_labor_per_unit
+        );
+    }
+}
```
lennyyyyyyyy, line 332: I feel like you should just override the Machine constructor to apply the scaling, instead of doing this. Something like:
```
Machine() {
   Product();
   living_labor_per_unit *= MACHINE_SCALE_MULTIPLIER;
   Society::get_instance()->set_underlying_living_labor_per_unit(this, living_labor_per_unit);
}
```

lennyyyyyyyy (changes requested):

header/Constants.h
alexbcreiner0: It's not arbitrary in concept, but the specific minimum is. I mostly included it for the sake of experimentation. I think if we're going to remove it, we should do it later, when we are more confident.

header/Producer.h
alexbcreiner0: No, I meant to remove it.

src/Society.cpp
alexbcreiner0: Yeah that was an old comment I wrote for myself and forgot to remove. This is fine as it is though, right?

src/Society.cpp
alexbcreiner0: Yes, I would have done it that way but the way I was originally attempting to scale the machine inputs was a lot more ambitious than what the final product turned out to be. Will do.

src/Firm.cpp
alexbcreiner0: I thought you had some secret reason lol.

src/Society.cpp
alexbcreiner0: Actually, I can't move it to the constructor. If I do, then it will end up double multiplying the living labor vector `l` and the system ends up with an inflated estimate of the initial labor values. Keeping these here.

lennyyyyyyyy: Can you explain why it double multiplies? I thought you would remove the logic here, so that the multiplication would only be taken care of by the Machine constructor.

lennyyyyyyyy (approved):
Approved. If you could respond to the thing about redundant scaling of the machine living labor that would be cool.

PR #293, opened 2026-08-04
Improved logging, and full Overseer integration into the project. 

alexbcreiner0:
# What's included
## Major improvements to the logging system in general as well as Overseer:
- Logging of observed living labor values over time, resupply rate, and resupply deficit added.
- Demand is now logged as actual demand (prior, the reorder_threshold was logged and Overseer called it demand). Now, the reorder_threshold is no longer a simple multiplier of demand, and our logic for reordering has moved away from firm's trying to make their pending inventory match the reorder_threshold and towards trying to make their reorder_rate match their rate of demand. Thus it was time to log these separately.
- Overseer now has a Demand category, where actual demand can be tracked along with these other related rates, as well as a Supply Chain category, which takes the role that the Demand category of plots did previously, allowing you to see how the reorder threshold compares to pending inventory. 
- What was being logged as a reorder failure previously was actually a drafting failure. These notions have now been soundly split. A drafting failure happens when a rejected plan is returned to the requester, and a reorder fully fails when there is no chosen producer. Overseer now allows for you to view these failures separately from each other. 
- Additionally, logging of drafting failures includes more reliable and specific information on why it failed. Whether it is due to a lack of workers or resources is reported, and if it was resources, the specific resource that is missing is reported.
- Similarly, start_plan and move_plan_forward_one_step failures also now log more reliably and give more specific information when they fail, including what is lacking as well as the deficit. 
- Reorder attempts are logged explicitly, whenever a firm enters the body of their check_and_reorder methods past the check for whether or not they will do so. 
- Overseer now has two different 'Problems' categories. Problems (Casualties) allows you to see which industries are experiencing problems, while Problems (Causes) allow you to see which industries are the source of the problem. By going back and forth between these, Overseer should be a fair bit more capable as a diagnostic tool. 

## Overseer integration
Overseer is now included inside of this repo as a **submodule**. This means the following:
- A folder appears in the repository for Overseer, but clicking it just takes you to the Overseer github page.
- If you clone the repo normally, Overseer will not be included. In order for submodules to be included, you have to clone it with the `--recurse-submodules` flag.
- I can push my own changes to Overseer however I see fit, without affecting the project, because the submodule is bound to a specific commit of Overseer. In order to update Overseer _within_ this project, you would need to enter the folder in your local branch, `git pull` to pull down the latest version, and then submit a pull request, with the change being 'we are deciding to use the repo with a newer commit of Overseer'. 
- Along with Overseer as a submodule, the LTE-Data-Aggregation-Layer repo is being retired, and brought into this repo permanently. This has its trade-offs - in particular it will add friction between me and my ability to quickly make changes to those files, but it will also force others to consider what that code is doing more closely, and to make changes to that code in parallel with the changes they are making to the C++ codebase. 
- A `setup.sh` script has been created which will allow new users to immediately jump into our simulation in Overseer without much fuss. They just have to clone the repo (with the `--recurse_submodules` flag so as to actually clone Overseer as well), then run the script. That script will do the following:
1. Create a virtual environment, install Overseer into it
2. make the binary from the makefile
3. Pre-empts Overseer's autocreation of a config file and creates one which sets the repo itself as the location for storing log files as well as where models are found. (the logs directory is added to the gitignore).
5. Create a demos.yml file with whatever initial view information we decide to have as 'official'. (This is kept in a file called `bc_lte_demo.yml`) `overseer_model/demos.yml` is part of the gitignore. This allows all of us to save whatever views are useful for us during development without dealing with constant merge conflicts, while also maintaining an 'authoritative' initial demo view that we can curate for new users.  
6. Create a `params.yml` file which properly binds the demo to the makefile which was created. (Both steps 5 and 6 here are executed via the `prepare_demo_and_params.py` python script. 
7. Create a `run.sh` script which the user can run to launch Overseer and the simulation without any knowledge of what a virtual environment even is. 

- There is also a `setup.ps1` PowerShell script for Windows users which does the same thing (untested, I had an LLM convert my `setup.sh` file, I have asked a friend using Windows to test it out on their machine and am still awaiting a confirmation that it works.)
- Overall, there are now several new files that are sort of awkwardly sitting in the root directory of our repo. I wasn't really sure where to put them. On the one hand, the python script and yaml file definitely don't belong where they are, but at the same time I thought it important to make the setup scripts as overt and hard to miss as possible. It didn't seem right to make a folder just for the single yaml and python script, so I left everything out in the front, but if someone wants to suggest otherwise I can change that. 
- Updated the README file with instructions relating to all of this.


header/Logger.h
```diff
@@ -67,7 +67,7 @@ void Logger::log(
         << "\"client\":\"" << clients[client] << "\","
         << "\"id\":" << id << ","
         << "\"label\":\"" << label << "\"";
-    (std::cout << ... << pairs) << "}" << std::endl;
+    (std::cout << ... << pairs) << "}\n";
```
alexbcreiner0, line 70: It can't be overstated how drastically this improved performance.

header/Firm.h
```diff
@@ -110,26 +111,40 @@ class Firm : public Agent {
...
-    void log_reorder(const Product * product, int quantity);
     void log_initial_employment(const unsigned int worker_id, const unsigned int workplace_id);
     void log_busyness(double firm_busyness, int max_workers_for_transfer);
     void log_employment_transfer(
             const unsigned int worker_id,
             const unsigned int old_workplace_id,
             const unsigned int new_workplace_id
             );
-    void log_reorder_failure(const Product * product, std::string reason);
+    void log_drafting_failure_goods(
```
Amittai-Aviram: Does "goods" in this name mean "inputs"? If so, `inputs` would be clearer, I think — and a clearer contrast with `workers` in the next function.

header/Firm.h
```diff
@@ -110,26 +111,40 @@ class Firm : public Agent {
...
             const double quantity
             );
-    void log_accepted_order(const Order * original_order, const Order * chosen_return_Order);
-    void log_plan_stallage(Plan * plan, std::string situation);
-    void log_start_plan_stalled(Plan * plan, Product * product);
+    void log_accepted_order(
+        const Order * original_order,
+        const Order * chosen_return_Order
+    );
+    void log_plan_stallage(Plan *, Product *, double);
```
Amittai-Aviram: Could we please change this name to `log_stalled_plan` or `log_plan_stall`?  (`Stallage` is an unnecessary, and perhaps somewhat gauche, neologism, if I may say so.)

header/Firm.h
```diff
@@ -110,26 +111,40 @@ class Firm : public Agent {
...
+    void log_plan_stallage_resolved(Plan *);
```
Amittai-Aviram: `log_plan_stall_resolved`

header/Plan.h
```diff
@@ -24,5 +24,6 @@ struct Plan {
     double quantity_remaining = 0.0;
     int outgoing_units_consumed = 0;
     bool is_stalled = false;
+    Product * stalled_resource = nullptr;
```
Amittai-Aviram: The resource is not stalled. I presume that you mean the resource (input good or machine?) whose shortage had caused the plan to stall, right? If so, I would go with `missing_resource` (which goes with the name of your new parameter in `get_max_order_quantity`), or `stalling_resource` or `short_resource`. But I wish there were a different way of doing this. If I understand the scenario correctly, this field is _usually_ `nullptr` — it only has meaning if the plan has stalled.

header/Plan.h
```diff
@@ -24,5 +24,6 @@ struct Plan {
     double quantity_remaining = 0.0;
     int outgoing_units_consumed = 0;
     bool is_stalled = false;
```
Amittai-Aviram, line 26: Shouldn't the status of being stalled be represented as just another one of the enumerators for the plan's status?

src/Firm.cpp
```diff
@@ -260,7 +261,9 @@ void Firm::check_and_reorder_input(Product * product) {
             this,
             lead_time
             );
-    send_order(order);
+    if (!send_order(order)) {
```
Amittai-Aviram, line 264: What happened here? It used to be the condition `if (!send_order(order))` — I wonder how it got changed so that you had to change it back.

src/Firm.cpp
```diff
@@ -353,15 +324,33 @@ void Firm::start_plan(Plan * plan) {
         plan->raw_materials_budget + plan->machinery_budget;
     if (plan->is_stalled) {
```
Amittai-Aviram, line 325: As I mentioned earlier, you might consider handling this as a distinct _status_ of the plan (i.e., as one of the enumerators) rather than as a boolean flag. But I will not put up a fight if you wish to keep it as it is. The logic struck me as a bit murky for a couple of re-readings. Maybe it's that I had taken the word _stalled_ to mean _stalled out_ rather than just _temporarily stalled_ or _waiting_ — which may be on me.

src/Firm.cpp
Amittai-Aviram: Please fix this indentation.

src/Firm.cpp
```diff
@@ -423,22 +415,19 @@ void Firm::move_plan_forward_one_step(Plan * plan) {
...
                 return_inputs_to_inventory(deducted_inputs, plan->order->customer);
 
-                if (!plan->is_stalled) {
+                bool new_stall = !plan->is_stalled ||
+                    plan->stalled_resource != requirement.first;
+
+                if (new_stall) {
+                    if (plan->is_stalled) {
+                        log_plan_stallage_resolved(plan);
+                    }
```
Amittai-Aviram, line 424: It looks as if there's some code duplication between here and `handle_start_plan_failure`. Can it be  refactored out?

src/Person.cpp
```diff
@@ -27,8 +27,9 @@ Person::Person():
     static unsigned int unique_id = 0;
     id = unique_id++;
 
+    double sigma = Sim::get_ability_stddev();
```
Amittai-Aviram: I know that $\sigma$ is highly conventional, but, as a rule, I'm for avoiding mere arbitrary mathematical symbols. Could this just be renamed `ability_std_dev`?

src/Producer.cpp
```diff
@@ -90,23 +91,34 @@ bool Producer::can_produce(Product * product) {
     return catalog.count(product);
 }
 
-double Producer::get_max_order_quantity(const Order * order) {
+double Producer::get_max_order_quantity(
+    const Order * order,
+    std::vector<Product *>& missing_resources,
+    bool& no_workers_available
```
Amittai-Aviram: Can `no_workers_available` be renamed to make it clear that it's a boolean out-parameter? Maybe `workers_are_unavailable`?

src/Producer.cpp
```diff
@@ -90,23 +91,34 @@ bool Producer::can_produce(Product * product) {
...
         double input_max_order_quantity = 
                 get_inventory_level(input.first) / input.second;
+        if (input_max_order_quantity < 1.0) {
+            missing_resources.push_back(input.first);
+        }
         max_order_quantity = 
             std::min(max_order_quantity, input_max_order_quantity);
     }
     std::vector<Person *> available_workers = get_available_workers(order);
+    no_workers_available = available_workers.empty();
```
Amittai-Aviram: If no workers are available, the maximum quantity to be returns is 0.0, correct?  (I can see that you are also checking for missing inputs as a side effect, so you would not want to return right away in this case.)

BTW, if the maximum quantity is less than the original order quantity, we could wind up allocating more workers than we need, right? Could we partially correct for this — while avoiding the circular dependency — by replacing `order->quantity` with the maximum order quantity as determined after looking at the inputs but before looking at machinery?

src/Firm.cpp
```diff
@@ -296,26 +299,8 @@ void Firm::start_plan(Plan * plan) {
...
-                    LogPairS("Status", "stalled when starting"),
-                    LogPair("Missing good",  input.first->id),
-                    LogPair("Required", amount_needed),
-                    LogPair("Have on hand", get_inventory_level(input.first))
-                );
-                plan->is_stalled = true;
-            }
-            return;
+            handle_start_plan_failure(plan, input.first, amount_needed);
+            return false;
```
Amittai-Aviram, line 303: OK, so now I am really confused! As soon as we detect anything missing, we just quit, which is the approach that you had wanted to begin with, rather than the approach of waiting until inputs come in. So then the customer goes through the whole plan drafting, supplier selection, etc., sequence. So why do we even need `is_stalled`? I'm sorry — I'm having trouble following the logic here.

Amittai-Aviram (changes requested):
A lot of good ideas here. 
This PR combines three fundamentally different things:
- Restructuring of the repo with Overseer and the Overseer model
- Various changes to the Overseer model
- Various changes to the simulation logic, and, especially, changes to the handling of stalled plans.
At least ideally, shouldn't these be three distinct PRs?
Meanwhile, I am pretty confused about your changes to the handling of stalled plans. Some of your logic seems to make sense on the assumption that we will follow my earlier approach, which is to keep trying to execute the plan, but your code also looks as if you are following your preferred approach, which is to have the plan quit as soon as it encounters anything missing. It's OK — I do  not mean to be dogmatic or stand in the way of progress, but I just do not fully understand your implementation in that case, because I don't understand how the `is_stalled` flag functions, especially in relation to `start_plan`. (I think I see how it works in `move_plan_forward_one_step`.) Could you please clarify?

src/Firm.cpp
alexbcreiner0: We don't just quit. I'm confused too. I just consolidated what we were already doing into the helper function `handle_start_plan_failure`. The returned boolean is only used for logging purposes. There is no major change to the logic that I am aware of outside of the logging.

src/Producer.cpp
alexbcreiner0: > If no workers are available, the maximum quantity to be returns is 0.0, correct? (I can see that you are also checking for missing inputs as a side effect, so you would not want to return right away in this case.)

Yes, that is why this boolean is necessary. I need to know whether the quantity was zero because there were no workers or because there were no resources. Without it, you can mistake lack of workers to lack of resources. 

> BTW, if the maximum quantity is less than the original order quantity, we could wind up allocating more workers than we need, right? Could we partially correct for this — while avoiding the circular dependency — by replacing order->quantity with the maximum order quantity as determined after looking at the inputs but before looking at machinery?

If the maximum quantity is less than the original order quantity, then the quantity of the return_order will be adjusted. Then, in `draft_plan_for_order`, workers will be assigned to the plan according to that downward adjustment. So no, I don't think so.

src/Firm.cpp
alexbcreiner0: Not as far as I can tell, unfortunately. Where `start_plan` needs to call `rollback_plan_inputs`, `move_plan_forward_one_step` needs to call `return_inputs_to_inventory`. Where `start_plan` needs to call `log_start_plan_stallage_resolved`, `move_plan_forward_one_step` needs to call `log_plan_stallage_resolved`. Because of those differences, the best I can do easily is move these steps into a separate helper function from `handle_start_plan_failure`. But I only created that helper because unlike here, those actions need to be taken potentially twice. So that helper avoids some code duplication in `start_plan`, but it can't easily be reused here.

src/Firm.cpp
alexbcreiner0: I might be misremembering, but: I think that I had it like this in my bugfix branch in early July, but that was towards a different purpose - of performing upstream cleanup whenever `start_plan` failed. Since we went with your 'trying again' strategy, it never made it into main. Here, it's just being used to log the failure, which is a different purpose.

src/Producer.cpp
alexbcreiner0: `no_workers_available` sounds like a boolean to me (it is certainly a logical proposition) but I am impartial so sure.

src/Person.cpp
alexbcreiner0: okay

src/Firm.cpp
alexbcreiner0: This isn't even a change I made? But okay.

src/Firm.cpp
alexbcreiner0: I don't have a problem with this, and agree it would be more elegant, but I am a little nervous about how that might disrupting the overall state machine dependent on the order status. I've also never really worked with enumerations before. Maybe this could be done in a future PR?

header/Firm.h
alexbcreiner0: Done.

header/Firm.h
alexbcreiner0: I suppose you may. Done.

header/Plan.h
alexbcreiner0: Yes, it is usually a `nullptr`. It's only there so the plan can remember what was missing. I don't have much feedback on the elegance of this, but I don't really know of a better way either. I changed the name as you wished.

src/Producer.cpp
Amittai-Aviram: The problem is not whether it's a boolean but that it's an out-parameter rather than an in-parameter — you are using it to _find out_ whether there are no workers.  The name of the boolean field `is_stalled` works this way — it implies a question — "Is the plan stalled?" — and it holds the answer, which is sort of how a boolean out-parameter would work. Obviously, it's not a big deal, but I paused over it while reading your code, because you pass `false` in, but only to initialize it to something — where its real meaning, so to speak, is its value after the function returns. Do you see what I mean?  But it's OK — hopefully, the reader will see that it's a reference in order to serve as an out-parameter.

src/Producer.cpp
Amittai-Aviram: I understand that that is the purpose of the `workers_not_available` out-parameter, but I do not see where `maximum_order_quantity` is _set_ to 0 as a direct consequence of no workers being available. It does get set to 0 in the following loop (for machines), because `machine_max_order_quantity` will always be 0 as a result of the multiplication in the numerator. But — if there are no workers, then, actually, there's no point in going through all the machines anyway — you really could just return 0 after setting `workers_not_available` to `true`, no?

src/Firm.cpp
Amittai-Aviram: OK, sorry — I see! Thanks!

Amittai-Aviram (approved):
Thank you for the explanations! They are very helpful.

Could you please consider my tiny suggestion about `get_max_order_quantity`? If you agree with it, perhaps you or I could just write a separate issue to change that logic to return 0 "early" in case there are no workers, and, therefore, any machine deficits would not be counted anyway (`if ... !available_workers.empty()`)?

Otherwise, LGTM! Thanks again for your explanations.

PR #295, opened 2026-08-10
Added surplus labor transfer system for busyness equalization between sectors

### LINKED ISSUE #294: Have workers who aren't busy enough sent to other firms, even if no firm is requesting it.
We've been aware for a while that there are some workers whose accounts drop continually over time. With the recent improvements to the logging system, we now have more reliable logging of sectoral busyness, as well as separate tracking of average accounts for workers in different sectors. With these, the problem becomes pretty clear:

Image

Workers in the distribution sector are significantly less busy than workers in other sectors. With that said, shouldn't this lead to employment transfers to other sectors? Not unless other sectors _request_ the transfer. With Devin's improvements to the ordering and drafting process, employment is now much more stable over time. In general, and especially at higher numbers of people, the need for transfers just isn't high enough to pull workers away from the non-busy sector.

I think that an easy fix would be to have firms send workers away who aren't busy enough even if no firms are explicitly requesting it.  Have firms who are consistently above the busyness average propose transfers even if the workers aren't explicitly needed for a plan. Something like that should do the job of forcing non-busy workers to bounce between sectors until the distribution of surplus workers is appropriate to each sector.

alexbcreiner0:
# Changes
- Busyness now only decays during working hours. As discussed during the meeting, this makes the signal significantly smoother. However, the signal being calculated isn't the same number anymore. Rather than being then number of hours worked on average divided by the number of total hours in a week, it now measures the number of hours worked on average divided by the total hours in the _working week_. A simply normalization is performed in Overseer (simply multiplying by $\omega$) has been added on the logger side to convert it into the old number for graphing purposes.
- Some functions relating to the transfer system are refactored for better readability.
- Resolves #294, but some additional discussion is necessary here. 

# Discussion of the solution for 294
Our current employment transfer system implicitly puts the agency in the hands of the firm with the labor deficit, since transfers only happen when a firm *needs* the extra labor. This was a logical, conservative, utilitarian choice that was sufficient for ensuring that every firm gets enough workers to meet social reproduction. However, because of our system's commitment to full employment, more workers will be hired to each sector than what these minimums require. While performing labor transfers when firms _need_ the additional labor is enough to make sure enough workers are allocated to each firm, this act alone is not enough to ensure that each firm ends up approximating a perfect ratio of 'surplus' to necessary employees necessary such that when plans are rationed, every worker in the society has roughly the same level of busyness. 

This is also problematic with respect to the conceptual backdrop for this system. Our labor transfer system is supposed to be a stand-in for the consequences of assuming that workers in these sectors that aren't busy enough would leave. But if employment transfers can only happen when another firm needs the labor, and no firms happen to need the labor, then workers in less-busy sectors are effectively trapped. 

The solution to both of these problems is to allow firms to transfer surplus workers to other firms even when another firm doesn't need the labor. The implementation in this PR performs this task once a week. The same busyness threshold calculations are done (the exact calculation of the number of workers has been slightly streamlined), and the workers are transferred to the firm which is currently the most busy. This happens in addition to the old transfer system which happened during drafting, though I'm not sure how necessary it is still. 

Doing this created a knock-on problem. Firm busyness was previously being computed as the average busyness of workers at the firm. The problem with this is that when a non-busy firm moves its workers to a busy firm, it creates an instantaneous change to the busyness of both firms, even though nothing has happened yet. In many cases, this was enough to completely flip the busyness situation in a complete 180, creating long-term stable oscillations in which one sector would empty into another and then back again. To deal with this, I had to add the tracking of firm busyness as an independent signal from person busyness. Societal busyness in lieu of this is also altered to take a weighted average of the firm busyness values rather than looking directly at the workers.

Here is an example run of the improvements:
image

Parameter settings for reproducing:
```
    params:
      exe_path: /home/alex/Nextcloud/Personal-Programming/Labor-Time-Economy-Simulation/bin/sim
      N_g: 5
      N_S: 10000
      N_h: 500
      N_inputs_max: 5
      N_m: 1
      N_p: 5
      N_d: 5
      W: 5
      T: 8
      S: 0.1
      v_ability: 0.1
      N_a: 5
      fixed_seed: true
      seed: 2981340000
      is_logging: false
      productivity: 0.9
      consump_epsilon: 0.6
      init_prices: labor_values
      free_goods: false
      new_free_good_interval: 2160
```

We can see that accounts are decently stable now. Not perfect, but I think that perfectly stable accounts is impossible with our current approach, and also not really necessary (even the GIC discusses the necessity of making credits have an expiration date). We see busyness equalizing without any significant chaos injected into the employment numbers over time. Finally, in the bottom right plot, you can see the ratio of the minimum number of employees needed vs the actual number of employees. You can see that these are all converging towards 0.6, which is the `consumption_demand_level` setting. In general, these will all tend towards whatever that parameter is, which makes sense given how we use it in the normalization process. 

# Limitations
This isn't a perfect fix, and in some ways exposes problems with the simulation that were harder to see before. Even with these changes, I had to increase the `DISTRIBUTION_COST_MULT` constant to something high enough to ensure that a decent number of workers were needed in the distribution sector. If this number is very small, then a single surplus worker can dramatically change the busyness of a firm, to the point where the system can't find a good ratio of surplus to minimal workers and ends up unable to stabilize. 

Machines expose a similar problem. If these changes are merged into main without any further alterations, it will no longer be a good idea to have a single firm producing only one single machine. The reason for this is that despite the resource costs for machines being very high, machine making only happens sporadically. The machine sector will end up looking not busy in its 'off season', and send off all of its workers, sometimes creating problems when it becomes time for machines to be made again. The solution to this problem is to improve the initialization algorithm to make machine making a more consistent task, not just a more difficult one.


src/Firm.cpp
```diff
@@ -49,8 +49,72 @@ void Firm::on_time_step() {
...
+
+void Firm::update_busyness() {
+    double current_busyness = 0.0;
+    if (!workers.empty()) {
+        current_busyness =
+            labor_hours_this_time_step / workers.size();
+    }
+
+    recent_busyness +=
+        (current_busyness - recent_busyness)
```
Amittai-Aviram, line 79: Suppose the workers have done some work just before this time step, but have not done any in this time step. Then wouldn't `(current_busyness - recent_busyness) / workers.size()` be `(0 - recent_busyness) / workers.size()` < 0? Does that make sense?

src/Firm.cpp
```diff
@@ -49,8 +49,72 @@ void Firm::on_time_step() {
...
+void Firm::offer_workers_for_transfer() {
+    int worker_surplus = get_number_available_workers_to_transfer();
+    if (worker_surplus <= 0) {
+        return;
+    }
+
+    Firm * best_receiver = nullptr;
+    double highest_busyness = -1.0;
+    for (Firm * firm : Society::get_instance()->get_firms()) {
+        if (firm == this) {
```
Amittai-Aviram, line 94: What if this firm is the one with the greatest busyness? Then it should not offer any workers, correct?

src/Firm.cpp
```diff
@@ -49,8 +49,72 @@ void Firm::on_time_step() {
...
+    int total_workers_transferred = 0;
+    std::vector<Person *> sorted_standby_workers(
+            standby_workers.begin(),
+            standby_workers.end()
+    );
+    std::sort(
+        sorted_standby_workers.begin(),
+        sorted_standby_workers.end(),
+        [&](Person * a, Person * b) {
+            return a->get_account() < b->get_account();
```
Amittai-Aviram, line 113: Interesting choice to sort on accounts rather than individual busyness.  Do you have an argument against using the latter?

src/Firm.cpp
```diff
@@ -49,8 +49,72 @@ void Firm::on_time_step() {
...
+        }
+    );
+    for (Person * worker : sorted_standby_workers) {
+        if (total_workers_transferred >= worker_surplus) {
+            break;
+        }
+        best_receiver->hire_worker(worker);
```
Amittai-Aviram, line 120: Anybody looking at this could well claim that individuals have no autonomy here — the firm just hands over the worker by making the other firm hire him/her. I think this is an argument in favor of the further revision we discussed, which would move the logic to the Person.

src/Firm.cpp
```diff
@@ -124,18 +188,27 @@ bool Firm::remove_input_from_inventory(
...
+    for (Person * worker : standby_workers) {
+        if (static_cast<int>(transfers.size()) == max_workers_to_transfer) break;
+        transfers.push_back(worker);
+    }
+    return transfers;
+}
+
+int Firm::get_number_available_workers_to_transfer() {
     if (workers.empty()) {
         return {};
```
Amittai-Aviram, line 213: Wouldn't you return 0 in this case?  (Integer numeric literals have type `int` by C++ rules.)

src/Firm.cpp
```diff
@@ -147,26 +220,17 @@ std::vector<Person *> Firm::propose_transfer(int workers_wanted) {
     if (relative_difference < TRANSFER_BUSYNESS_THRESHOLD) {
         return {};
```
Amittai-Aviram, line 221: Again — this function returns a value of type `int`, not a `std::vector` or `struct` type.

src/Firm.cpp
```diff
@@ -147,26 +220,17 @@ std::vector<Person *> Firm::propose_transfer(int workers_wanted) {
...
-        if (static_cast<int>(transfers.size()) == max_workers_to_transfer) break;
-        transfers.push_back(worker);
-    }
-    return transfers;
-}
-
-void Firm::finalize_transfer(Person * worker) {
-    standby_workers.erase(worker);
-    workers.erase(worker);
+    double target_workers = workers.size() * (firm_busyness) / societal_busyness;
```
Amittai-Aviram, line 223: Why is `firm_busyness` in parentheses?

src/Firm.cpp
```diff
@@ -147,26 +220,17 @@ std::vector<Person *> Firm::propose_transfer(int workers_wanted) {
...
+    int available_workers_for_transfer = std::max(
+        0,
+        static_cast<int>(
+            std::floor(workers.size() - target_workers)
+        )
+    );
+    return std::min(
+        available_workers_for_transfer,
+        static_cast<int>(standby_workers.size())
+    );
```
Amittai-Aviram, line 233: Could you please clarify your rationale here? In particular, why are the standby workers brought into consideration at all?  As I understand it, `available_workers_for_transfer` is the number of workers you would want to keep on so that they would all be reasonably busy.  The standby workers are a subset of the workers taken into account to compute `available_workers_for_transfer` — they are the workers not busy in a plan at the moment when this computation is done.

Amittai-Aviram (changes requested):
Good changes for now.
As I mention in a comment, this implementation definitely looks as if the decision to transfer workers lies in the firm's rather than the workers' hands. The "system" element here is the the reporting of persons', firms', and society's busyness, as well as the mechanisms making transfer possible (e.g., `hire_worker`).  Those should remain in place, but I think we should (soon-ish) change the implementation (the combination of "reasonable behavior" and "unpredictable decisions") to put the worker's choice at the center. I will write up an issue accordingly.
There are a couple of issues involving returning the wrong type. It strikes me as odd that the compiler has no trouble silently converting `{}` to `0`, but I don't think we should rely on this in our code, if I may say so.

src/Firm.cpp
alexbcreiner0: I don't think that's possible. If they are indeed the busiest firm, then their busyness won't fall underneath the threshold checked in `get_number_available_workers_to_transfer`, which compares the firm busyness to the social average. Then the worker surplus would be zero, and the function would have already returned by this point. Still, it couldn't hurt to add a check for this.

src/Firm.cpp
alexbcreiner0: Well, the accounts are the true symptom we're trying to correct for. Even though busyness is the cause, I thought it appropriate to address the symptoms first, so to speak.

Another explanation: suppose a worker whose account was dwindling got moved from a non-busy firm to a busy firm, which suddenly became non-busy as soon as they started at the new firm. Then their account wouldn't have recovered at all, and leaving them at the new firm would make matters worse for them. 

This just seemed like the more aggressive approach given that the goal is really to equalize accounts rather than busyness.

src/Firm.cpp
alexbcreiner0: Well, the argument is the same as it always has been: that the alignment of interests makes it not important who is deciding these things. That said, I've been saying the same thing since Devin first implemented our system back in March. So while I'm happy you're starting to share this concern with me, fixing that problem wasn't really the goal of this PR.

src/Firm.cpp
alexbcreiner0: Yeah it should. I don;t know why the compiler let me get away with returning this instead lol

src/Firm.cpp
alexbcreiner0: Yeah I'll fix.

src/Firm.cpp
alexbcreiner0: Because I was previously doing something else here and didn't fully clean up. Will fix.

src/Firm.cpp
alexbcreiner0: I'm confused. The rule being enforced here is: 'you can't transfer workers who aren't on standby`. You seem to be saying this yourself in the last sentence here. If the number of workers who the calculation revealed you can transfer is 7, but you only have 5 workers on standby, then return 5 instead of 7. That's what this is doing. Could you please clarify what you think is wrong with this?

lennyyyyyyyy (changes requested):
Still a little iffy on the strategy to correct accounts far from the average by periodically inducing transfers. It's good if this solution fixes the system settling into an employment split that isn't optimally clustered in busyness, but I don't think we should target accounts, especially since if we find that optimal split in employment, that divergence goes down with population size.

src/Firm.cpp
```diff
@@ -49,8 +49,77 @@ void Firm::on_time_step() {
     if (plans_in_progress.size()) {
         log_plans();
     }
+    if (Society::get_instance()->is_within_work_schedule()) {
+        update_busyness();
+    }
     double firm_busyness = get_busyness();
     log_busyness(firm_busyness);
+    if (Sim::get_current_time_step() % WEEK == 0) {
```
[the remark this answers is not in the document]
lennyyyyyyyy: What does the WEEK refer to then? If it's just matching the busyness averaging window, you should use `BUSYNESS_AVERAGING_WINDOW` instead.

src/Firm.cpp
```diff
@@ -49,11 +49,98 @@ void Firm::on_time_step() {
...
+}
+
+void Firm::update_busyness() {
+    double current_busyness = 0.0;
+    if (!workers.empty()) {
+        current_busyness =
+            labor_hours_this_time_step / workers.size();
+    }
+
+    double working_week_length =
```
Amittai-Aviram: Can you move this code out to a tiny helper function to avoid code duplication?

src/Firm.cpp
```diff
@@ -49,11 +49,98 @@ void Firm::on_time_step() {
...
+    double current_busyness = 0.0;
+    if (!workers.empty()) {
+        current_busyness =
+            labor_hours_this_time_step / workers.size();
+    }
+
+    double working_week_length = 
+        Sim::get_work_hours_daily() * Sim::get_work_days_weekly();
+
+    current_busyness *= working_week_length / WEEK;
```
Amittai-Aviram: Actually, maybe you should have that helper function return `working_week_length / WEEK`, i.e., `omega`. Translated into English, that might be something like `work_portion_of_week`.

src/Firm.cpp
```diff
@@ -49,11 +49,98 @@ void Firm::on_time_step() {
...
+
+    recent_busyness +=
+        (current_busyness - recent_busyness)
+        / BUSYNESS_AVERAGING_WINDOW;
+
+    labor_hours_this_time_step = 0.0;
+}
+
+void Firm::offer_workers_for_transfer() {
```
Amittai-Aviram: This function does not offer workers, it transfers them. The distinction is important — there is no moment of decision-making, either on the part of the worker or on that of the receiving firm. The name should reflect what the function does.

Amittai-Aviram: Nothing currently calls this function, AFAI can tell.

src/Firm.cpp
```diff
@@ -121,50 +208,46 @@ bool Firm::remove_input_from_inventory(
...
 }
 
 std::vector<Person *> Firm::propose_transfer(int workers_wanted) {
+    int max_workers_to_transfer = std::min(
+        workers_wanted,
+        get_number_available_workers_to_transfer()
+    );
+    std::vector<Person *> transfers;
+    for (Person * worker : standby_workers) {
+        if (static_cast<int>(transfers.size()) == max_workers_to_transfer) break;
```
Amittai-Aviram: I do not understand this condition. Did you perhaps mean `<` rather than `==`? I'd expect that you would push a worker from `standby_workers` onto `transfers` if and only if the number `transfers` is less than the maximum number to be transferred.

src/Firm.cpp
```diff
@@ -121,50 +208,46 @@ bool Firm::remove_input_from_inventory(
...
-        transfers.push_back(worker);
+        return 0;
     }
-    return transfers;
-}
-
-void Firm::finalize_transfer(Person * worker) {
-    standby_workers.erase(worker);
-    workers.erase(worker);
+    double target_workers = workers.size() * firm_busyness / societal_busyness;
```
Amittai-Aviram: Can you please group `(firm_busyness / societal_busyness)` in parentheses?

src/Firm.cpp
```diff
@@ -596,15 +689,16 @@ void Firm::adjust_quantity_for_deadline(Plan * plan) {
...
-    if (plan->workers.empty()) {
-        return std::numeric_limits<double>::infinity();
-    }
-    double labor_hours_per_timestep = 
-        plan->workers.size()
-        * get_work_week_proportion();
-    return plan->order->quantity *
-           recorded_living_labor_per_unit[plan->order->product] /
-           labor_hours_per_timestep;
+    return plan->order->requested_turnaround_time;
```
Amittai-Aviram: So can you please just replace the (sole) call to this function with its body and eliminate this function? The call is in `Firm::assign_plan_dependent_fields`:
```
    draft_plan->order->predicted_turnaround_time =
        predict_turnaround_time(draft_plan);
```
This would be changed to the following:
```
    draft_plan->order->predicted_turnaround_time =
        draft_plan->order->requested_turnaround_time;
```
This, of course, makes me wonder whether we can eliminate one of the two fields in `Plan` and simply call it `turnaround_time`.

src/Firm.cpp
```diff
@@ -596,15 +689,16 @@ void Firm::adjust_quantity_for_deadline(Plan * plan) {
...
+    // if (plan->workers.empty()) {
+    //     return std::numeric_limits<double>::infinity();
+    // }
+    // double labor_hours_per_timestep = 
+    //     plan->workers.size()
+    //     * get_work_week_proportion();
+    // return plan->order->quantity *
+    //        recorded_living_labor_per_unit[plan->order->product] /
+    //        labor_hours_per_timestep;
```
Amittai-Aviram: Please delete — preferably (from my POV) by eliminating the whole funciton.

src/Person.cpp
```diff
@@ -212,9 +212,15 @@ void Person::update_health_status() {
...
-    double growth = busyness_this_time_step / BUSYNESS_AVERAGING_WINDOW;
-    double decay = busyness / BUSYNESS_AVERAGING_WINDOW;
-    busyness += growth - decay;
+    double growth = busyness_this_time_step;
+    double decay = busyness;
+
+    double working_week_length =
+        Sim::get_work_hours_daily() * Sim::get_work_days_weekly();
+
+    growth *= working_week_length / WEEK;
```
Amittai-Aviram: As noted earlier, you might consider defining a helper function, perhaps a `static` function in `Sim`, that computes `omega` — `work_portion_of_week` or whatever.

README.md
Amittai-Aviram: ... or, equivalently, run ...
(That is, if you agree that it has exactly the same effect. If it does, it's helpful to the user to know that it makes no difference which procedure he/she follows — cloning Overseer with `git clone` or pulling in the submodule recursively — since Overseer, as you mentioned, itself, has no submodules.)

README.md
```diff
@@ -19,7 +19,7 @@ If you have already cloned the repo and did not use this flag, you can either cl
 git submodule update --init --recursive
 ```
 2. Navigate inside of the repo and run one of the following two scripts depending on your operating system: 
-- If you are on Mac OS or Linux, run `./setup.sh`. 
+- If you are on Mac OS or Linux, run `./setup.sh`. If you are a Mac OS user, you may first need to make the script executable before this command will work: `chmod +x setup.sh`
 - If you are on Windows, run `.\setup.ps1` (or simply right click the file and choose Run with PowerShell). (Note that you must use PowerShell to run this script. Do not try to use `cmd`.)
 3. In either case, this script will create a file called `run.sh` or `run.ps1`, again depending on your operating system. Running that file in the same way should launch Overseer and immediately begin the simulation.
```
Amittai-Aviram: You might want to mention that `run.sh` is not actually part of the repo, it's an artifact of the setup script — and also let the user know what he/she should do if he/she accidentally deletes `run.sh` and would like to have it again.  (This was an issue for me because, when I attempted to switch to any branch, Git prevented me, saying that it would overwrite my unsaved changes in `run.sh`.  It also seems that `run.sh` is not currently in `.gitignore`. Did you mention that you planned to add it in?)

run.sh
```diff
@@ -0,0 +1,9 @@
+#!/bin/bash
+
```
Amittai-Aviram: SInce this file is an artifact of `setup.sh` or `setup.p1`, shouldn't it be removed from the repo and also added to `.gitignore`? And please let uw know what to do if something messes the file up or deletes it. The setup script is idempotent, correct? So it's safe simply to re-run it to generate `run.sh` — correct?

Amittai-Aviram (changes requested):
Sorry — a few more small changes requested, plus some handling of `run.sh` before we go any further. Thanks!

src/Firm.cpp
alexbcreiner0: I already had one, I forgot to use it. Done.

src/Firm.cpp
alexbcreiner0: We already have a function for this that I just forgot to use `get_work_week_proportion`

src/Firm.cpp
alexbcreiner0: Huh? It just means stop transferring when you have transferred the maximum number.

src/Firm.cpp
alexbcreiner0: Actually, I don;t think I should delete this. This was from some testing I was doing to see if what I was doing would be the same, and it wasn't. So thanks for the reminder, it's reverted to what it was before now.

run.sh
alexbcreiner0: Yes, sorry. That was an oversight.

src/Firm.cpp
Amittai-Aviram: Sorry. Please put `break` on a line by itself as a block:
```
if (...) {
    break;
}
```
Thanks.

src/Firm.cpp
Amittai-Aviram: OK

README.md
Amittai-Aviram: Good!

Amittai-Aviram (approved):
I am asking for one tiny style-related change — in the code that accounted for my having missed the word `break` at the end of a longish line of code. Otherwise, LGTM.  Nevertheless, perhaps we should discuss this at our meeting before it goes in? I am a bit concerned about this approach to transfering workers --- or, put another way, I would prefer an implementation that allows an individual to move to another firm if his/her account is low relative to consumption and his/her firm is not busy (or he/she is not busy there), with some probability in [0, 1] that models an individual choice. My question is how hard it would be to undo the current approach and replace it with such an implementation. And perhaps we all need to be on board — including Devin, the other reviewer.

lennyyyyyyyy (approved):

lennyyyyyyyy (approved):

PR #296, opened 2026-08-14
Implement a firm's general account

### LINKED ISSUE #243: Implement a firm's general account
Generalize the current property of a `pooled_input_account` in `Firm` to represent the firm's general account of labor-time credits.
- The account begins at 0.
- When a plan begins (`Firm::start_plan`), its raw materials and machinery budgets are transferred into its general account. This creates a debt in the plan.
- When a plan ends and the firm receives payment, the payment is counted against the debt.
- The debt (or credit) is transferred into the account at some point to be determined as part of this issue.

Amittai-Aviram:
Resolves #243 

# Sequence of Events
- When a firm buys machinery and input goods, it pays for them out of its general account (its `account`).
- When a firm drafts a plan, it drafts a budget with three line items — `machinery_budget`, `raw_materials_budget`, and `labor_budget`. A fourth item, `debt`, is simply the sum of these three. (Note that I have switched this to a _positive_ sum. It is the quantity of debt created by fiat when the budget is put into effect.)
- When a firm starts executing a plan, in `start_plan`, it transfers the value of the three lines (or, equivalently, the debt), as a positive value, to its account, for use in paying for machinery, inputs, and labor.
- When the firm moves the plan forward on each time step, it pays its workers from its account.
- When the firm completes the plan, in `end_plan`, 
  - It ships the order to the customer and receives payment, which is _deducted_ from the plan's debt.
    - If the total value of the shipped product equals the original budget, the debt will come down to 0. 
    - If the price has come down since this firm began executing its plan, because the average necessary direct or indirect labor has decreased, a portion of the debt will remain.
    - If the price has increased since the firm started executing the plan, the debt will be less than 0.
  - It _subtracts_ the debt from its general account. This ensures that the account will reflect the degree to which the firm has gotten reimbursed for its expenditures on the plan by selling its output products to the ordering customer.
    - If the debt is still positive, this will decrease the general account.
    - If the debt is negative, it will increase the account, showing that the firm has sold its goods for more than what it had originally budgeted.
  - it adds (refunds) the value of any unused inputs (including machine capacity) to the account. This ensures that the amount put into the account, by the end of the plan, reflects the plan's actual expenditures.

The new _Firm Accounts_ graph shows how the firms' accounts oscillate within a small range, hovering around 0:
Screenshot 2026-08-13 at 10 30 26 PM

### DIFF
--- a/header/Firm.h
+++ b/header/Firm.h
@@ -72,10 +72,9 @@ class Firm : public Agent {
     bool start_plan(Plan * plan);
     void handle_start_plan_failure(Plan * plan, Product * product, double missing);
     void return_inputs_to_inventory(
-        std::unordered_map<Product *, double> deducted_inputs,
-        Firm * firm
+        std::unordered_map<Product *, double> deducted_inputs
     );
-    void rollback_plan_inputs(Plan * plan, Firm * firm);
+    void rollback_plan_inputs(Plan * plan);
     void move_plan_forward_one_step(Plan * plan);
     void end_plan(Plan * plan);
     void move_plans_forward_one_step();
--- a/header/Firm.h
+++ b/header/Firm.h
@@ -73,7 +73,7 @@ class Firm : public Agent {
         std::unordered_map<Product *, double> deducted_inputs
         );
     void refund_for_unused_inputs(
-            const std::unordered_map<Product *, double> deducted_inputs);
+            const std::unordered_map<Product *, double> returned_inputs);
     void rollback_plan_inputs(Plan * plan);
     void move_plan_forward_one_step(Plan * plan);
     void end_plan(Plan * plan);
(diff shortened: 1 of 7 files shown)


alexbcreiner0 (changes requested):
This is great work. After a bit of thought, it makes perfect sense how the firm accounts should equalize to zero roughly over time. I don't see any difference between what the code says and what your description explains. Only thing that needs changing is a compiler error due to a badly named loop variable, a few minor cosmetic alterations to the plotting, and possibly a small refactoring for the sake of code readability. 

Two more small things that I think you might consider adding in this PR, relating to plans. They don't really fit this PR, but they are small changes that would help clean up the code ahead of the presentation.
1. I think that the plan's `outgoing_units_consumed` field is completely unused anywhere in the codebase. We can delete it.
2. I think that `needed_this_step` shouldn't be a plan field either. The only place it is used is during `move_plan_forward_one_step`, where it is completely recalculated every time it executes. It could and probably should just be a local variable within that function rather than a plan property.

If you agree with these points, then we could smuggle that housekeeping in here and that would be nice. I could also include this in my own PR instead of having you include them in this one.

alexbcreiner0 (approved):

PR #299, opened 2026-08-17
Float working week with a simpler heuristic

### LINKED ISSUE #223: Allow the length of the working week to float towards a minimal level (Outline of a plan provided)

lennyyyyyyyy:
Resolve #223 
- Makes the smoothed busyness metric work with local plan working weeks (which may differ from the societal working week for a short period of time)

For the working week, I decided to try a less sophisticated method using the new smoothed busyness. On every `WORK_WEEK_ADJUSTMENT_PERIOD`, it calculates the proportion of each individual busyness to the current upper bound of busyness: $\frac{b}{\omega}$.
The update to the working week then goes in the direction that gets the proportion within the range `WORKING_WEEK_STABLE_START - WORKING_WEEK_STABLE_END`, which is set to 0.65 - 0.75.

If that was it, then there is no consideration of the spread of the busyness distribution. The algorithm is equivalent to a single evaluation on the mean busyness $\frac{\bar{b}}{\omega}$. So, a `WORKING_WEEK_BUSYNESS_POWER` is introduced, which both the proportion $\frac{b}{\omega}$ and the stable range are exponentiated on. This results in pushing the effective target range higher when there is more spread to the distribution.

Results:
image
image
image


overseer_model/simulation/parameters.py
```diff
@@ -25,3 +25,5 @@ class Params:
     init_prices: str = 'labor_values'
     free_goods: bool = False
     new_free_good_interval: int = 2160
+    work_week_adjustments: bool = True
+    work_week_adjustment_interval: int = 336
```
Amittai-Aviram: How did you arrive at this value?

src/Person.cpp
```diff
@@ -47,6 +47,7 @@ Person::Person():
     log_inventory();
     account = Society::get_instance()->get_initial_account();
     log_account();
+    busyness = Sim::get_product_consumption_mult() * Society::get_instance()->get_work_week_proportion();
```
alexbcreiner0: I really don't like this. We can't be just keep assuming approximations of every signal prior to the simulation starting. This also doesn't even make sense given the other starting conditions - everyone is unemployed to start. Please instead introduce a delay in the updating procedure to allow for the busyness signal to become accurate. This can easily be done intelligently using an integer multiple of the `BUSYNESS_AVERAGING_WINDOW`. Even better, have the simulation wait this long _after_ full employment has already been reached.

header/Constants.h
```diff
@@ -13,6 +13,11 @@ const unsigned int INITIAL_WORK_HOURS_DAILY = 8;
 const unsigned int INITIAL_WORK_DAYS_WEEKLY = 5;
 const double INITIAL_ACCOUNT_DURATION = YEAR;
 
+const double WORKING_WEEK_BUSYNESS_POWER = 2;
```
alexbcreiner0: I think this should be removed. The power isn't arbitrary - the choice being made is to take the root mean square busyness. So I think other local variable names in the function using it can be changed to simply reflect that this is what you are doing, instead of having this constant here.

alexbcreiner0 (approved):
As a toggleable and experimental feature of the sim, I think that this is acceptable as a proof of concept. Approved.

